PR #8, opened 2019-10-29
Generate a cluster id if one is not provided

skitt:
This allows the user to skip that part of the CR.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/8)

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -3,6 +3,7 @@ package submariner
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"strconv"
 
 	"github.com/go-logr/logr"
@@ -370,6 +371,10 @@ func newRouteAgentDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet
 	return routeAgentDaemonSet
 }
 
+const (
+	dns1123 = "abcdefghijklmnopqrstuvwxyz0123456789"
+)
+
 //TODO: move to a method on the API definitions, as the example shown by the etcd operator here :
 //      https://github.com/coreos/etcd-operator/blob/8347d27afa18b6c76d4a8bb85ad56a2e60927018/pkg/apis/etcd/v1beta2/cluster.go#L185
 func setSubmarinerDefaults(submariner *submarinerv1alpha1.Submariner) {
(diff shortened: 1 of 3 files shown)


pkg/controller/submariner/submariner_controller.go
```diff
@@ -367,19 +368,32 @@ func newRouteAgentDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet
...
 		spec.Repository = "quay.io/submariner"
 	}
 
-	if spec.Version == "" {
+	if len(spec.Version) == 0 {
 		spec.Version = "0.0.2"
 	}
+
+	if len(spec.ClusterID) == 0 {
+		// Generate a random, 16-character cluster id
```
mangelajo: Hmm, big question, if we do this, shouldn't we "write-back" the CR in the API?, otherwise, on next run we would generate a different clusterID because it's again empty.

skitt: Good point, we need that anyway to be able to debug...

pkg/controller/submariner/submariner_controller.go
```diff
@@ -367,19 +368,32 @@ func newRouteAgentDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet
...
+	letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+)
+
 //TODO: move to a method on the API definitions, as the example shown by the etcd operator here :
 //      https://github.com/coreos/etcd-operator/blob/8347d27afa18b6c76d4a8bb85ad56a2e60927018/pkg/apis/etcd/v1beta2/cluster.go#L185
 func setSubmarinerDefaults(submariner *submarinerv1alpha1.Submariner) {
 
 	spec := submariner.Spec
-	if spec.Repository == "" {
+	if len(spec.Repository) == 0 {
```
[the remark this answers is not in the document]
skitt: Also, it matches https://github.com/coreos/etcd-operator/blob/8347d27afa18b6c76d4a8bb85ad56a2e60927018/pkg/apis/etcd/v1beta2/cluster.go#L185

pkg/controller/submariner/submariner_controller.go
skitt: This is now done in the reconciliation function.

skitt:
I’m not sure why this is failing. The ClusterID is generated correctly, and the CR is updated, as can be seen from the logs. However the IPsec keys are never set up and the engine spends its time looping over updates from the broker.

mangelajo:
@skitt what do you think about basing this on the idea of using the "default" k8s namespace uuid? (or a part of it?)

manosnoam:
Currently we expect users to modify their kubeconfig context, before running subctl join with their chosen clusterId.

@mangelajo, @skitt 
Can you set clusterId with a constant identifier of current cluster context, with some API call (instead of generating a random UUID that beeing hard to trace later) ?

skitt:
Dropped in favour of #616.

PR #12, opened 2019-10-30
Add subctl skeleton

skitt:


mangelajo:
It already passed linter/validation, and the rest of deployment details won't change because of this patch, so I'm merging! :)

PR #13, opened 2019-10-30
WIP: Add fns to create Endpoints/Clusters CRDs

skitt:
Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

PR #14, opened 2019-10-30
Set the broker up using subctl

skitt:
It took me way too long to figure out that we need a different clientset for CRDs...

Depends: #13

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -52,27 +52,6 @@ function kind_clusters() {
     fi
 }
 
-function install_helm() {
-    helm init --client-only
-    helm repo add submariner-latest https://submariner-io.github.io/submariner-charts/charts
-    if kubectl --context=cluster1 -n kube-system rollout status deploy/tiller-deploy > /dev/null 2>&1; then
-        echo Helm already installed on cluster1, skipping helm installation...
-    else
-        logs=$(mktemp)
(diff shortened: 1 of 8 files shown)


skitt:
This is still a mess, and should be split into several commits, but I thought it would be interesting to get reviews of the final result.

TODO:
[ ] Move the CRD creation to a separate package
[ ] Split out the labeling change to a separate commit
[ ] Split out the Helm removal to a separate commit

scripts/kind-e2e/e2e.sh
```diff
@@ -52,27 +52,6 @@ function kind_clusters() {
     fi
 }
 
-function install_helm() {
```
dfarrell07: Huzzah!

scripts/kind-e2e/e2e.sh
```diff
@@ -94,86 +73,14 @@ function setup_broker() {
         echo Submariner CRDs already exist, skipping broker creation...
     else
         echo Installing broker on cluster1.
-        helm --kube-context cluster1 install submariner-latest/submariner-k8s-broker --name ${SUBMARINER_BROKER_NS} --namespace ${SUBMARINER_BROKER_NS}
+        ../bin/subctl --kubeconfig ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster1 create broker
```
mangelajo: wohooo! :-D

PR #15, opened 2019-10-31
Network discovery and info command

mangelajo:
This patch implements a network discovery that works at least for:

* weavenet
* flannel
* gke (only pod IPs)

In addition to the already existing OpenShift4 detection.

Also adds a basic info command which can be used to inspect a cluster
and get basic information.


mangelajo:
Posting as WIP for now. Today/tomorrow I'm not around.

Central fun is here:
https://github.com/submariner-io/submariner-operator/pull/15/files#diff-92a528ef86209edbc036b07f552ea8a3

And the patch is bigger than I expected sorry... :]

PR #17, opened 2019-11-04
Ensure we update the CR after setting defaults

skitt:
Setting defaults should update the CR; this patch changes the provided
instance and updates the CR.

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -101,6 +101,9 @@ func (r *ReconcileSubmariner) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	setSubmarinerDefaults(instance)
+	if err = r.client.Update(context.TODO(), instance); err != nil {
+		return reconcile.Result{}, err
+	}
 
 	// Create submariner-engine SA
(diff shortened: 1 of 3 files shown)


pkg/controller/submariner/submariner_controller.go
```diff
@@ -101,6 +101,9 @@ func (r *ReconcileSubmariner) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	setSubmarinerDefaults(instance)
+	if err = r.client.Update(context.TODO(), instance); err != nil {
```
mangelajo: Should we always update? or just if instance changed?, may be we should return a flag from setSubmarinerDefaults that indicates changes, or make a copy and compare afterwards?

skitt: I noticed that if we update all the time, the stored CR is updated to match the validated CR; so for example, the invalid `size` field is removed. I think this is useful, because users who retrieve the CR from the cluster will see what was actually used, not what they fed in. Other operators follow this pattern too.

mangelajo: Ack, thanks for the explanation.

mangelajo:
@dfarrell07 ^ we need a second review :)

PR #18, opened 2019-11-04
Configure replicas

skitt:
Depends: https://github.com/submariner-io/submariner-operator/pull/17

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/18)

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -184,7 +184,10 @@ func newDeploymentForCR(cr *submarinerv1alpha1.Submariner) *appsv1.Deployment {
 		"component": "engine",
 	}
 
-	replicas := int32(1)
+	replicas := int32(cr.Spec.Count)
+	if replicas < 1 {
+		replicas = 1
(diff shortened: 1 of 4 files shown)


mangelajo:
@skitt It seems like you need to fix the verify step on the replica count:
https://travis-ci.com/submariner-io/submariner-operator/builds/134898133#L2598

And that'd be it.

skitt:
> @skitt It seems like you need to fix the verify step on the replica count:
> https://travis-ci.com/submariner-io/submariner-operator/builds/134898133#L2598

Yes (and thanks for the review), but does that make sense? The current test checks that we’ve got as many replicas running as we requested, which is also what Kubernetes uses to consider that a service is available. So I wonder whether changing the test would only be hiding a bad design decision: I’m not sure we should request more replicas than the number of labeled nodes.

dfarrell07:
I see the same timeouts locally as CI shows FYI.

mangelajo:
I guess we can drop this one if we are moving to daemonset.

PR #24, opened 2019-11-06
Install submariner-operator logic for subctl

mangelajo:
This PR automatically:

*  generates a crdyamls.go file with the
operator CRDs embedded. It's not included in the repository
and it's automatically re/generated by the Makefile logic
when it's missing or the original yaml files have changed.

* ensures the operator is installed.


mangelajo:
Oh, dang it!, I need to refactor this to grab other types of yaml, like the role, and role bindings, etc...

mangelajo:
This one is now ok, hopefully should pass CI , it did locally. Another chunk of the E2E bash script gone! :)

mangelajo:
had a missing file... new push..

scripts/kind-e2e/e2e.sh
```diff
@@ -122,7 +122,7 @@ function create_subm_vars() {
   ce_ipsec_ikeport=500
   ce_ipsec_nattport=4500
 
-  subm_ns=operators
+  subm_ns=submariner-operator
```
dfarrell07: Should we make other OLM/docs changes to match this namespace?

scripts/kind-e2e/e2e.sh
```diff
@@ -256,7 +256,8 @@ for i in 2 3; do
     verify_subm_gateway_label
 
     # Deploy SubM Operator
-    deploy_subm_operator
+    ../bin/subctl install-operator --image submariner-operator:local \
```
dfarrell07: w00t! Amazing :D

scripts/kind-e2e/e2e.sh
mangelajo: Good question. With OLM you tell the olm where to put it, in my docs I picked "openshift-submariner".  We could standardize to just submariner-operator may be, so it's uniform u/s and d/s.

If we give freedom to pick one subctl could get confused if you provide a different ns, because it could look like the operator is not installed, although it is.. (in the default namespace..).

mangelajo:
@dfarrell07 sorry for that suffering, let me add a dependency in the makefile from e2e to subctl

scripts/build
```diff
@@ -4,4 +4,5 @@ source $(dirname $0)/lib/debug_functions
 source $(dirname $0)/lib/version
 
 $(dirname $0)/build-operator
+$(dirname $0)/generate-embeddedyamls
```
mangelajo: I left the generate-embedded yamls out of the build-subctl to make sure it's not re-generated always. Although, I guess it's very quick, and probably could make sense to get rid of the makefile logic around that, and simply throw the generate inside build-subctl

dfarrell07:
> @dfarrell07 sorry for that suffering, let me add a dependency in the makefile from e2e to subctl

Awesome, thanks for making that easier/more useable @mangelajo, much appreciated!

PR #26, opened 2019-11-09
Datafile handling pkg

mangelajo:
This commit defines the data exported from the broker creation
and imported on cluster join. It contains the submariner k8s broker
URL, the token for connection, and the IPSec PSK as a k8s secret.


mangelajo:
Thanks for looking guys!

PR #27, opened 2019-11-11
Generate random Broker PSK and store in Secret

dfarrell07:
The purpose of this, is that we will be providing an unified IPSEC key between runs of
subctl, aditionally this could be used to automate rotation of IPSEC keys.

### DIFF
--- a/pkg/broker/ipsec_psk_test.go
+++ b/pkg/broker/ipsec_psk_test.go
@@ -1,0 +1,29 @@
+package broker
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const testPSKLen = 32
+
+var _ = Describe("ipsec_psk handling", func() {
+	When("generateRandonPSK is called", func() {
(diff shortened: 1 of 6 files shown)


mangelajo:
@dfarrell07 I took freedom to add some testing, fix one of your to-dos, and tiny refactor to only have one call to the module.

pkg/broker/ipsec_psk_test.go
```diff
@@ -0,0 +1,29 @@
+package broker
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const testPSKLen = 32
+
+var _ = Describe("ipsec_psk handling", func() {
```
dfarrell07: Ah, these tests are great, thanks! I was wondering about how best to do add tests, thanks for the informative example. :)

dfarrell07:
> @dfarrell07 I took freedom to add some testing, fix one of your to-dos, and tiny refactor to only have one call to the module.

That's perfect, thanks for the tweaks/tests, they look good to me. :+1:

PR #31, opened 2019-11-11
Install operator with broker and pull broker creation to it's own module

mangelajo:
Pull the broker install logic to it's own package, so the command is cleaner, and add the logic to install the operator along with the broker (which we may make optional)

### DIFF
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -1,0 +1,66 @@
+package broker
+
+import (
+	"fmt"
+
+	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/broker/ensure.go
```diff
@@ -0,0 +1,66 @@
...
+	_, err = apiext.ApiextensionsV1beta1().CustomResourceDefinitions().Create(NewEndpointsCRD())
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("error creating the Endpoint CRD: %s", err)
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("error creating the core kubernetes clientset: %s", err)
+	}
+	// Create the namespace
```
skitt: Bit of a nitpick, but this comment and a few of the following give the impression they’re attached to the code preceding them, rather than the code following them; a blank line before, and none after, would be clearer.

mangelajo: nitpick accepted :)

PR #32, opened 2019-11-11
Implements method to get SubctlData from cluster

mangelajo:
The method NewFromCluster returns a SubctlData structure from the
cluster, by embedding the API data, broker service account credentials
and IPSEC Key.

### DIFF
--- a/pkg/broker/ipsec_psk.go
+++ b/pkg/broker/ipsec_psk.go
@@ -8,7 +8,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 )
 
-const IPSECPSKSecretName = "submariner-ipsec-psk"
+const ipsecPSKSecretName = "submariner-ipsec-psk"
 
 // generateRandomPSK returns securely generated n-byte array.
 func generateRandomPSK(n int) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


pkg/broker/ipsec_psk.go
```diff
@@ -5,8 +5,11 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
 )
 
+const IPSECPSKSecretName = "submariner-ipsec-psk"
```
skitt: Can we avoid exporting this? Or is it needed for the test?

pkg/broker/rbac.go
```diff
@@ -1,15 +1,24 @@
...
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	SubmarinerBrokerNamespace = "submariner-k8s-broker"
+	SubmarinerBrokerSA        = "submariner-k8s-broker-client"
+	SubmarinerBrokerRole      = "submariner-k8s-broker-client"
```
skitt: As above, it would be better if we could avoid exporting these constants.

mangelajo: ack, thanks.

pkg/broker/ipsec_psk.go
mangelajo: I think we don't even need it for the test, we can hide it, yes. Good point.

I also took the chance to camelCase psk_secret_data , even if it's not part of this patch specifically.

pkg/broker/rbac.go
skitt: They’re still exported ;-).

PR #33, opened 2019-11-12
Add Submariner clientset

skitt:
This allows us to manipulate the CRDs managed by the operator, currently `Submariner`. We need this to tell the operator what to deploy.


mangelajo:
Thanks for looking at this, I somehow expected the operator-sdk would handle it, but it doesn't

PR #34, opened 2019-11-12
Fix getClients() return order

skitt:
Errors should always be returned last.


mangelajo:
Lol, I promise you I was adding the same change on my side.

PR #35, opened 2019-11-12
Reorganize commands and handle datafile

mangelajo:
Now we have three top-level commands:

  * deploy-broker
  * join
  * info


mangelajo:
Added a commit on top to handle newFromCluster returning an error, which we didn't check

mangelajo:
Thanks!!!

PR #36, opened 2019-11-12
Label nodes

mangelajo:
The other PR needs to be merged first, look only at the last commit of this PR


pkg/subctl/cmd/deploybroker.go
```diff
@@ -0,0 +1,52 @@
...
+
+var disableDataplane bool
+
+func init() {
+	deployBroker.PersistentFlags().BoolVarP(&disableDataplane, "no-dataplane", "n", false,
+		"Don't install the submariner dataplane on the broker")
+	rootCmd.AddCommand(deployBroker)
+}
+
+const IPSECPSKBytes = 48 // using base64 this results on a 64 character password
```
dfarrell07: We use a mix of 32 and 48 for this currently I think.

mangelajo: That's from the other patch. You mean testing vs implementation? , I'll check that on a separate PR, thanks daniel.

mangelajo:
ok, I'm rebasing this on top of master  to leave only the node-labeling code.

pkg/subctl/cmd/root.go
```diff
@@ -85,46 +89,87 @@ func handleNodeLabels() error {
...
+	if err != nil {
+		return struct{ Node string }{}, err
+	}
+	return answers, nil
+}
+
+// this function was sourced from:
+// https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/test/utils/density_utils.go#L36
+func addLabelsToNode(c kubernetes.Interface, nodeName string, labels map[string]string) error {
+	const retries = 5
```
dfarrell07: Nit: 500ms x 5 attempts, 2.5s total of waiting, seems like it might be a bit demanding. Maybe 20 attempts at 500ms? or 1s intervals?

dfarrell07:
> ok, I'm rebasing this on top of master to leave only the node-labeling code.

Awesome, much more clear now thanks! Strange how bad that diff was, GitHub...

pkg/subctl/cmd/root.go
mangelajo: yeah, may be it's a good idea to increment that. May be I can even implement the ExponentialBackoff pattern here. 

The problem with labeling nodes, is that kubernetes will update those objects every 10 seconds, with the host status, so it's likely that you could hit an issue, not sure about using patch though, but I saw the examples using retry.

mangelajo:
Oh, sorry I wanted to handle @dfarrell07 comments on this patch, I will add another PR.

skitt:
> Oh, sorry I wanted to handle @dfarrell07 comments on this patch, I will add another PR.

Ah, right, since @dfarrell07 approved the patch I thought everything was OK!

PR #38, opened 2019-11-12
Create Submariner CR

skitt:


pkg/subctl/cmd/deploybroker.go
```diff
@@ -0,0 +1,52 @@
...
+		"Don't install the submariner dataplane on the broker")
+	rootCmd.AddCommand(deployBroker)
+}
+
+const IPSECPSKBytes = 48 // using base64 this results on a 64 character password
+const brokerDetailsFilename = "broker-info.subm"
+
+var deployBroker = &cobra.Command{
+	Use:   "deploy-broker",
+	Short: "set the broker up",
```
dfarrell07: Nit: "setup the broker"?

pkg/subctl/cmd/join.go
```diff
@@ -1,13 +1,44 @@
...
+	SubmarinerNamespace = "submariner"
+)
+
 var joinCmd = &cobra.Command{
-	Use: "join",
+	Use:   "join",
+	Short: "connect a cluster to an existing broker",
+	Args:  cobra.ExactArgs(1), // exactly one, the broker data file
+	Run: func(cmd *cobra.Command, args []string) {
+
```
dfarrell07: Nit: Remove newline at start of function to be consistent with other relevant code.

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,75 @@
...
+	submariner "github.com/submariner-io/submariner-operator/pkg/apis/submariner/v1alpha1"
+	submarinerclientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/datafile"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+func Ensure(config *rest.Config, remoteNamespace string, submarinerNamespace string,
```
dfarrell07: Nit: "Ensure" is not a super-clear function name to me.

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,75 @@
...
+	clusterID string, serviceCIDR string, clusterCIDR string,
+	subctlData *datafile.SubctlData) error {
+
+	fmt.Printf("%v\n", string(subctlData.ClientToken.Data["ca.crt"]))
+
+	// Populate the Submariner CR for the operator
+	submarinerSpec := submariner.SubmarinerSpec{
```
dfarrell07: Yay! This will free up lots of Bash for removal.

pkg/subctl/operator/deploy/ensure.go
mangelajo: I accept suggestions, my vocabulary is a bit short in that area.
https://www.thesaurus.com/browse/ensure

Provide? :? 

Anyway, that's not part of @skitt patch, github is horrible with chaining patches. I'll send another PR to handle that which is merged.

Thanks daniel

pkg/subctl/cmd/deploybroker.go
mangelajo: +1

pkg/subctl/operator/deploy/ensure.go
mangelajo: Wow yeah!

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,75 @@
...
+		Version:                  "",
+		Repository:               "",
+		CeIPSecNATTPort:          0,
+		CeIPSecIKEPort:           0,
+		CeIPSecDebug:             false,
+		CeIPSecPSK:               string(subctlData.IPSecPSK.Data["psk"]),
```
mangelajo: I'm not sure if it will work as string, or if you need to do base64.URLEncoder.encode (or whatever the function is)

pkg/subctl/cmd/join.go
```diff
@@ -6,13 +6,27 @@ import (
...
+var (
+	clusterID   string
+	serviceCIDR string
+	clusterCIDR string
+)
+
 func init() {
 	rootCmd.AddCommand(joinCmd)
+	joinCmd.Flags().StringVar(&clusterID, "clusterid", "", "cluster ID used to identify the tunnels")
+	joinCmd.Flags().StringVar(&serviceCIDR, "servicecidr", "", "service CIDR")
```
mangelajo: I will add the network detection on top to avoid the cmdline flags :)

pkg/subctl/cmd/join.go
```diff
@@ -32,5 +46,8 @@ var joinCmd = &cobra.Command{
 		fmt.Printf("* Deploying the submariner operator\n")
 		err = install.Ensure(config, OperatorNamespace, operatorImage)
 		panicOnError(err)
+
+		fmt.Printf("* Deploying Submariner\n")
+		err = deploy.Ensure(config, "submariner-k8s-broker", SubmarinerNamespace, clusterID, serviceCIDR, clusterCIDR, subctlData)
```
mangelajo: nit: use the constant for the Namespace instead?

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func Ensure(config *rest.Config, remoteNamespace string, submarinerNamespace string,
+	clusterID string, serviceCIDR string, clusterCIDR string,
+	subctlData *datafile.SubctlData) error {
+
+	// Create the CRDs we need
+	apiext, err := apiextension.NewForConfig(config)
```
mangelajo: nit: we can refactor the Endpoint/Cluster CRD creation later so it's shared with the broker

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// Create a clientset for the other standard kubernetes resources
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("error creating the core kubernetes clientset: %s", err)
+	}
+
+	// Create the namespace
+	_, err = clientset.CoreV1().Namespaces().Create(NewSubmarinerNamespace())
```
mangelajo: We need to deploy the CR on the same namespace the operator is running, right?

pkg/subctl/cmd/join.go
skitt: Yes, I was hoping you’d do that ;-). We’ll probably need the flags anyway, for case where the detection doesn’t find the values.

pkg/subctl/operator/deploy/ensure.go
skitt: Yes, the CRDs aren’t broker-specific so they need to be moved out and shared.

mangelajo: And as you said, the will need to go into the operator itself eventually

pkg/subctl/cmd/join.go
mangelajo: True, I was thinking about:
0) Provided flags override anything, with a warning if they don't match detection
1) Detection -> works, use that
2) If there's a parameter missing we ask the admin, like for the gateway node.

pkg/subctl/operator/deploy/ensure.go
dfarrell07: For me the confusion is that I'm left with a question like "ensure what does what?". IMHO a more verbose function name, maybe adding a subject and/or verb, would be worth it for the extra clarity. Again, it's just a small nit and thanks!

scripts/kind-e2e/e2e.sh
```diff
@@ -194,7 +194,7 @@ function cleanup {
 function dump_clusters_on_error {
     for i in 1 2 3; do
         echo "Dumping cluster information for cluster$i ==========================="
-        kubectl --context=cluster$i cluster-info dump
+#        kubectl --context=cluster$i cluster-info dump
```
dfarrell07: Should either remove, remove comment, or move comment "#" to be indented (space on the left, not right).

scripts/kind-e2e/e2e.sh
```diff
@@ -269,10 +285,6 @@ for i in 2 3; do
     verify_subm_operator_container
 
     # FIXME: Rename all of these submariner-engine or engine, vs submariner
-    # Create SubM CR
```
dfarrell07: We can do it in a follow-up patch, but we shouldn't forget to get rid of these functions.

scripts/kind-e2e/e2e.sh
skitt: I didn’t mean to commit that ;-).

pkg/subctl/cmd/join.go
skitt: It turns out we don’t need it, it’s stored in `subctlData`!

scripts/kind-e2e/e2e.sh
skitt: Done.

dfarrell07:
Seems to be failing on this check in CI:

```
[scripts]$ kubectl get submariner submariner --namespace=submariner-operator -o jsonpath={.spec.repository}
[scripts]$ grep local
```

skitt:
> Seems to be failing on this check in CI:
> 
> ```
> [scripts]$ kubectl get submariner submariner --namespace=submariner-operator -o jsonpath={.spec.repository}
> [scripts]$ grep local
> ```

Yes, I need to add flags to `join` for a few more pieces of the CR...

pkg/subctl/operator/deploy/ensure.go
mangelajo: We can target the operator namespace, and only keep one for simplicity may be... just an idea.

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -0,0 +1,109 @@
...
+	submarinerSpec := submariner.SubmarinerSpec{
+		Repository:               repository,
+		Version:                  version,
+		CeIPSecNATTPort:          0,
+		CeIPSecIKEPort:           0,
+		CeIPSecDebug:             false,
+		CeIPSecPSK:               base64.StdEncoding.EncodeToString(subctlData.IPSecPSK.Data["psk"]),
+		BrokerK8sCA:              base64.StdEncoding.EncodeToString(subctlData.ClientToken.Data["ca.crt"]),
+		BrokerK8sRemoteNamespace: string(subctlData.ClientToken.Data["namespace"]),
+		BrokerK8sApiServerToken:  string(subctlData.ClientToken.Data["token"]),
```
mangelajo: Oh, no need for encoding on this one? :)

skitt: Surprisingly enough, no ;-).

PR #39, opened 2019-11-13
Add PSK byte/string getter fns, use in datafile

dfarrell07:
Also changes datafile from storing Secret to string.

Also marks places PSK key length vars are dependent on each other, and
where ideally we'd find ways to de-duplicate as much as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/39)

### DIFF
--- a/pkg/broker/ipsec_psk.go
+++ b/pkg/broker/ipsec_psk.go
@@ -2,6 +2,7 @@ package broker
 
 import (
 	"crypto/rand"
+	"encoding/base64"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


dfarrell07:
Any idea what's going on with this `make ci` test failure @mangelajo?

```
datafile when Getting data from cluster [It] Should produce a valid structure 
/go/src/github.com/submariner-io/submariner-operator/pkg/subctl/datafile/datafile_test.go:79

  Unexpected error:
      <*errors.errorString | 0xc000074cc0>: {
          s: "proto: Secret: illegal tag 0 (wire type 0)",
      }
      proto: Secret: illegal tag 0 (wire type 0)
  occurred

  /go/src/github.com/submariner-io/submariner-operator/pkg/subctl/datafile/datafile_test.go:81
```

pkg/broker/ipsec_psk.go
```diff
@@ -40,3 +41,25 @@ func NewBrokerPSKSecret(bytes int) (*v1.Secret, error) {
...
+
+func GetIPSECPSKBytes(clientSet clientset.Interface, brokerNamespace string) ([]byte, error) {
+	secret, err := GetIPSECPSKSecret(clientSet, brokerNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	// FIXME: Use shared var for PSK byte array length everywhere
+	psk := make([]byte, 48)
+	err = secret.Unmarshal(psk)
```
mangelajo: there must be a psk.Data["psk"] structure inside that secret with the data already in bytes, I think. I will check it.

mangelajo:
I'm not sure if we will need to do this, this could be intended for grabbing it from the broker. For now, we may need something more in the line of getting the base64 from the datafile

PR #40, opened 2019-11-13
Wait for the client token to be ready

mangelajo:
When the broker account is created, the credentials token is created by kubernetes
asynchronously. Without this patch it will fail sometimes because the token isn't created fast enough.

### DIFF
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -2,9 +2,11 @@ package broker
 
 import (
 	"fmt"
+	"time"
 
 	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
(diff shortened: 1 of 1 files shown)


pkg/broker/ensure.go
```diff
@@ -64,5 +66,35 @@ func Ensure(config *rest.Config, ipsecPSKBytes int) error {
...
+		Duration: 5 * time.Second,
+		Factor:   1.2,
+		Jitter:   1,
+	}
+
+	var lastErr error
+	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		_, lastErr = GetClientTokenSecret(clientset, SubmarinerBrokerNamespace)
+		if lastErr != nil {
+			return false, nil
```
skitt: Don’t we want

    return false, lastErr

here? Or would that involve decoding the error to figure out if we want to retry?

mangelajo: if we do that, the wait loop stops and exits https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L292

So if we have an error because the token does not exist yet, we need to continue.

mangelajo: I  must admit I also found it weird.

skitt: Right, that’s what I meant by decoding the error: some errors indicate that we should continue (“token not found” etc.), others indicate that there’s no point waiting further (if the clientset failed to communicate with the target). Let’s leave it like that for now, at least we won’t miss errors!

PR #42, opened 2019-11-13
Install SubM on broker cluster

Jaanki:
One of the test scenarios requires dynamically
adding/removing third cluster to the setup.
The idea is to use broker cluster as the third
cluster.

This patch just installs SubM on broker cluster
but does not tag any gateway nodes. Node is tagged
later in the test case to mimic adding third cluster.

Signed-Off-By: Janki Chhatbar jchhatba@redhat.com

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/42)

### DIFF
--- a/scripts/kind-e2e/cluster1-config.yaml
+++ b/scripts/kind-e2e/cluster1-config.yaml
@@ -12,3 +12,5 @@ kubeadmConfigPatches:
     dnsDomain: cluster1.local
 nodes:
 - role: control-plane
+  - role: worker
+  - role: worker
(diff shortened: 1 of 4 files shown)


mangelajo:
Shouldn't this commit go to submariner instead of submariner-operator? 

This is the e2e just for submariner-operator, although I guess if we want to run full E2E at some point they should be very similar, right?

dimaunx:
As @mangelajo mentioned this change should be in submariner repo and then back ported here if required.

Jaanki:
Similar changes are already merged in submariner repo. https://github.com/submariner-io/submariner/pull/201

mangelajo:
But I think it's not necessary here as we don't run the submariner-tree e2e here, I will close for now, and we can revisit in the future.

PR #47, opened 2019-11-15
Handle the retries of node-labeling with exponential backoff

mangelajo:
Handling the comments by @dfarrell07 about retry times on previous commits. This should make it work more safely.


pkg/subctl/cmd/root.go
```diff
@@ -154,20 +153,32 @@ func addLabelsToNode(c kubernetes.Interface, nodeName string, labels map[string]
...
 		}
-		time.Sleep(retryInterval)
+	})
+
+	if err == wait.ErrWaitTimeout {
+		return lastErr
 	}
-	return err
+
+	return nil
```
mangelajo: this should probably be "err" in this case, as it could be the "non-conflict" case, ... editing..

PR #48, opened 2019-11-15
Provide defaults, check clusterID

skitt:
The current release of the operator doesn’t fill in default values, so
we need to ensure subctl does it instead.

ClusterID needs to be a valid DNS-1123 name, enforce that in Ensure()
for now.

### DIFF
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -19,6 +19,9 @@ var (
 	clusterCIDR string
 	repository  string
 	version     string
+	nattPort    int
+	ikePort     int
+	colorCodes  string
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -74,7 +89,8 @@ func joinSubmarinerCluster(subctlData *datafile.SubctlData) {
 	panicOnError(err)
 
 	fmt.Printf("* Deploying Submariner\n")
-	err = deploy.Ensure(config, SubmarinerNamespace, repository, version, clusterID, serviceCIDR, clusterCIDR, subctlData)
+	err = deploy.Ensure(config, SubmarinerNamespace, repository, version,
```
mangelajo: At some point I propose we turn this into a structure, it's growing out of control, lol :D

pkg/subctl/operator/deploy/ensure.go
```diff
@@ -52,12 +53,22 @@ func Ensure(config *rest.Config, submarinerNamespace string, repository string,
...
 	}
 
+	if len(repository) == 0 {
+		// Default repository; this is handled in the operator after 0.0.1
+		repository = "quay.io/submariner"
+	}
+
+	if len(version) == 0 {
+		// Default version; this is handled in the operator after 0.0.1
+		version = "0.0.1"
```
mangelajo: we need to use 0.0.2,

0.0.1 is the initial rancher release for submariner, right?

pkg/subctl/cmd/join.go
skitt: I wonder what structure would be suitable ;-).

PR #50, opened 2019-11-15
Discovery for canal-flannel, and avoid crash on join

mangelajo:
Crash on join happened if networkDiscovery was nil, which can
still happen even for err == nil.

### DIFF
--- a/pkg/discovery/network/canal_test.go
+++ b/pkg/discovery/network/canal_test.go
@@ -1,0 +1,74 @@
+package network
+
+import (
+	v1 "k8s.io/api/core/v1"
+	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


pkg/discovery/network/canal_test.go
```diff
@@ -0,0 +1,74 @@
...
+		"net-conf.json": `{
+			"Network": "10.0.0.0/8",
+            "SubnetLen": 20,
+            "SubnetMin": "10.10.0.0",
+            "SubnetMax": "10.99.0.0",
+            "Backend": {
+                "Type": "udp",
+                "Port": 7890
+            }
+		}`,
```
skitt: Mix of tabs and spaces?

mangelajo: oops also something failed on CI, thanks.

mangelajo:
error was a timeout on weave.

mangelajo:
Tab-mess fixed :)

PR #51, opened 2019-11-15
Handle OpenShift's privileged SCC

mangelajo:
When installing on top of OpenShift the operator service
account must be added to the privileged scc, otherwise it
won't be able to create privileged pods.

### DIFF
--- a/pkg/subctl/operator/install/serviceaccount/ensure.go
+++ b/pkg/subctl/operator/install/serviceaccount/ensure.go
@@ -15,7 +15,7 @@ import (
 
 //go:generate go run generators/yamls2go.go
 
-const OperatorServiceAccout = "submariner-operator"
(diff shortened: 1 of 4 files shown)


pkg/subctl/operator/install/serviceaccount/ensure.go
```diff
@@ -15,6 +15,8 @@ import (
 
 //go:generate go run generators/yamls2go.go
 
+const OperatorServiceAccout = "submariner-operator"
```
skitt: Typo (no particular need to fix it, don’t update the PR just for this): `Account`.

mangelajo: oops, I can do ti quickly, np

mangelajo: Done, just use squash when merging.

PR #54, opened 2019-11-19
Ask for missing information

skitt:
If the user hasn’t provided a cluster id or color codes on the command
line, ask for them.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/54)

### DIFF
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -52,14 +52,6 @@ var joinCmd = &cobra.Command{
 	Short: "connect a cluster to an existing broker",
 	Args:  cobra.ExactArgs(1), // exactly one, the broker data file
 	Run: func(cmd *cobra.Command, args []string) {
-		// Make sure the clusterid is a valid DNS-1123 string
-		if match, _ := regexp.MatchString("^[a-z0-9][a-z0-9.-]*[a-z0-9]$", clusterID); !match {
-			fmt.Printf("Cluster IDs must be valid DNS-1123 names, with only lowercase alphanumerics,\n"+
-				"'.' or '-' (and the first and last characters must be alphanumerics).\n"+
-				"%s doesn't meet these requirements\n", clusterID)
-			return
-		}
-
 		subctlData, err := datafile.NewFromFile(args[0])
 		panicOnError(err)
 		fmt.Printf("* %s says broker is at: %s\n", args[0], subctlData.BrokerURL)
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -69,6 +61,51 @@ var joinCmd = &cobra.Command{
 
 func joinSubmarinerCluster(subctlData *datafile.SubctlData) {
 
+	// Missing information
+	var qs = []*survey.Question{}
+
+	if len(clusterID) == 0 {
```
mangelajo: can we change it to...  len(clusterID)==0 or !isValidClusterID(str)  ?

In such way we avoid the final panic condition if the cmdline was wrong.

pkg/subctl/cmd/join.go
```diff
@@ -69,6 +61,51 @@ var joinCmd = &cobra.Command{
...
+			Validate: func(val interface{}) error {
+				str, ok := val.(string)
+				if !ok {
+					return nil
+				}
+				return isValidClusterID(str)
+			},
+		})
+	}
+	if len(colorCodes) == 0 {
```
mangelajo: can we also add a default color in the cobra cmdline variable?, that way we avoid asking the admin for this field. 

This is something rancher internal I think we even talked to chris once about removal... so i'd better not to ask the admin for unnecessary stuff. They could use labels/annotations on the clusters for that.

pkg/subctl/cmd/join.go
```diff
@@ -69,6 +61,51 @@ var joinCmd = &cobra.Command{
...
+		if len(answers.ClusterID) > 0 {
+			clusterID = answers.ClusterID
+		}
+		if len(answers.ColorCodes) > 0 {
+			colorCodes = answers.ColorCodes
+		}
+	}
+
+	// Make sure the clusterid is valid (we check twice because this might come from the command line)
+	err := isValidClusterID(clusterID)
```
mangelajo: can be removed with the top if condition change.

pkg/subctl/cmd/join.go
skitt: Ah yes, I like that.

pkg/subctl/cmd/join.go
skitt: Yes, indeed.

mangelajo: ack thanks !! :)

skitt: I’ve left the colour question here, but with #55 it won’t be asked unless the user explicitly specifies a blank colour code on the command line.

PR #56, opened 2019-11-19
Enable NAT by default, since it's the most common case.

mangelajo:
It can be disabled with --disable-nat, and also non-nat environments will also work with nat support enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/56)


mangelajo:
timed out deploying netshot, so re-running.

Issue #64, opened 2019-11-25
We need cross-compilation for subctl (OSX, windows, linux32, etc..)


mangelajo:
https://github.com/submariner-io/submariner-operator/releases

skitt:
Sorry, I hadn’t noticed that this was assigned to @dimaunx!

PR #65, opened 2019-11-25
Cross-build subctl for Linux/x86, Windows, macOS

skitt:
Fixes: #64

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/65)


dimaunx:
The only thing maybe to add a git commit that the version was build from. main.COMMIT=${COMMIT} we could output better info from version command. LGTM.

Issue #67, opened 2019-11-27
Trying to use subctl with gke/gcp results on error


mangelajo:
miguelangel@cloudshell:~ (kubenet)$ ./subctl-v0.0.2-linux-amd64 deploy-broker
* Deploying broker
panic: error setting up the engine requirements: error creating the api extensions client: No Auth Provider found for name "gcp"

patches which look similar: https://gitlab.cncf.ci/kubernetes/kubernetes/commit/1e7fa1f6cd2046678f8c37ab3328cdda2a05bf07

PR #68, opened 2019-11-28
Fixes subctl compatibility issue with GKE kubeconfig files

mangelajo:
This enables compilation of the required auth plugins used
in google cloud.

Fixes Issue: #67

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/68)

### DIFF
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/root.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
skitt: We could fix this once and for all with

    _ "k8s.io/client-go/plugin/pkg/client/auth"

See [the client-go](https://github.com/kubernetes/client-go/tree/master/examples#auth-plugins) example for details.

mangelajo: ohh, does that work like that?, I'll do it, Thanks skitt! :)

mangelajo: Done, and tested.

PR #71, opened 2019-11-29
Add licensing information

skitt:
This makes it clear that the code is under ASL-2.0 (same as
Submariner) and adds appropriate headers to the files we’ve created or
extensively modified.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/71)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -1,3 +1,19 @@
+/*
+© 2019 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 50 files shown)


mangelajo:
@skitt tiny issue on verification: https://travis-ci.com/submariner-io/submariner-operator/builds/138728056#L1125

it complais because of a comment directive.

cmd/manager/main.go
```diff
@@ -1,3 +1,19 @@
+/*
```
dfarrell07: Does it matter if this file was created entirely by the operator-sdk? If I go back and look at the old scripted generation logic, it seems like we delete/recreate this code every time without edits. I see @mangelajo made one commit to it after it was moved, but that was just to change these paths:

```
-       "github.com/submariner-operator/submariner-operator/pkg/apis"
-       "github.com/submariner-operator/submariner-operator/pkg/controller"
+       "github.com/submariner-io/submariner-operator/pkg/apis"
+       "github.com/submariner-io/submariner-operator/pkg/controller"
```

Non-blocker

mangelajo: It should be fine, as far as I remember operator ask does not impose a license .

mangelajo: But good point

Issue #69, opened 2019-11-28
The subctl deploy-broker syntax is confusing to the user


mangelajo:
Right now you could do:

```
export KUBECONFIG=clusterA
subctl deploy-broker  --clusterID=clusterA

export KUBECONFIG=clusterB
subctl join broker-info.subm
```

And that would be enough to connect the dataplane of both clusters, but people seems to be confused by the fact that "deploy-broker" will also deploy the dataplane (unless you say --no-dataplane).

For the shake of simplicity I believe we may want to change to:
```
export KUBECONFIG=clusterA
subctl deploy-broker
subctl join broker-info.subm  --clusterID=clusterA

export KUBECONFIG=clusterB
subctl join broker-info.subm  --clusterID=cluster B
```

manosnoam:
Yes, that will be clearer, since "deploy-broker" command is not "deploy-broker-and-join".

skitt:
I agree, I always use `--no-dataplane`. If we want we can always provide an explicit all-in-one command (see #62).

PR #70, opened 2019-11-28
Separate deploy-broker and join

skitt:
deploy-broker’s ability to set the dataplane up is confusing, it’s
best to leave that to an explicit join (or the all-in-one deploy which
is explicitly all-in-one).

Fixes: #69
Signed-off-by: Stephen Kitt <skitt@redhat.com>

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/70)

PR #75, opened 2019-12-02
Flip the dataplane default

skitt:
This changes the broker deployment command so that it doesn’t join the
cluster by default, and adds a --dataplane option to join (instead of
the --no-dataplane option to disable joining).

This is an alternative to #70, and would also fix: #69.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/75)


mangelajo:
Thanks @skitt !!! wow :) that was quick

PR #79, opened 2019-12-03
Allow consecutive runs of subctl join

mangelajo:
This will make sure that we update the Submariner operator resource in the
server

Fixes issue: #49

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/79)

PR #84, opened 2019-12-04
Fix update error in deploy.Ensure

tpantelis:
Fixes issue outlined in https://github.com/submariner-io/submariner/issues/242.
Call Create instead of Update and check for IsAlreadyExists.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/84)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ deploy/crds/submariner_endpoints_crd.yaml
 output/
 vendor/
 *.coverprofile
+.idea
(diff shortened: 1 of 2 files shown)


skitt:
I think this is dealt with more comprehensively in #79, isn’t it?

tpantelis:
> I think this is dealt with more comprehensively in #79, isn’t it?

yeah I didn't realize there was another patch to address this - we can close this.

PR #87, opened 2019-12-06
Allow --no-dataplane

skitt:
To simplify documentation, and because it’s used in demos, we should
keep --no-dataplane for now. This patch allows it, sets it by default,
but effectively ignores it.

---
This change is [Reviewable](https://reviewable.io/reviews/submariner-io/submariner-operator/87)

### DIFF
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -26,10 +26,15 @@ import (
 )
 
 var enableDataplane bool
+var disableDataplane bool
 
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&enableDataplane, "dataplane", false,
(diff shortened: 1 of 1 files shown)


manosnoam:
Does it means that for now we'll have 2 optional flags:
--no-dataplane
--dataplane

Where "--no-dataplane" is also the default action, if not specifying anything ?

skitt:
> Does it mean that for now we'll have 2 optional flags:
> --no-dataplane
> --dataplane
> 
> Where "--no-dataplane" is also the default action, if not specifying anything ?

Yes, but `--no-dataplane` is hidden. The point of this PR is that, since we’ve started publishing content using `--no-dataplane`, it’s a shame to make it suddenly break execution when we don’t need it to.

PR #97, opened 2019-12-17
Provide explicit error if no non-master nodes

skitt:
When attempting to label a gateway, if subctl doesn’t find any
non-master nodes, it errors out with an obscure error message (“please
provide options to select from”, produced by the survey library).
This patch handles this situation explicitly, with a friendlier error
message (“there are no non-master nodes in the cluster”).


mangelajo:
At some point we also need to support when there is just a "master" node, and that's all, where pods can be scheduled (may be we can detect that by taints) . That could be a possible common case for testing AIO environment, or edge AIO deployments.

PR #106, opened 2019-12-18
Skip labeling if there are no nodes

skitt:
In subctl, if we don’t find any candidate nodes for labeling, skip
instead of exiting. This sets up Submariner, which will then wait for
a gateway to be labeled.

This also adds a --no-label option to skip labeling in all cases.


mangelajo:
Thanks @sridhargaddam  !!

Issue #112, opened 2019-12-19
subctl crashes detecting network on small k3s


mangelajo:
$ ./subctl-v0.0.3-linux-amd64 join broker-info.subm 
* broker-info.subm says broker is at: https://149.202.169.255:6443
? What is your cluster ID? k3s-ovh-1
* There are 1 labeled nodes in the cluster:
  - k3s-test2
* Deploying the submariner operator
* Created operator CRDs.
* Created operator namespace: submariner-operator
* Created operator service account and role
* Deployed the operator successfully
* The operator is up and running
* Discovering network details
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x12019a0]

goroutine 1 [running]:
github.com/submariner-io/submariner-operator/pkg/subctl/cmd.getServiceCIDR(0x0, 0x0, 0x0, 0x1e, 0x0, 0x0, 0x0)
	/go/src/github.com/submariner-io/submariner-operator/pkg/subctl/cmd/join.go:192 +0x1c0
github.com/submariner-io/submariner-operator/pkg/subctl/cmd.joinSubmarinerCluster(0xc000499bc0)
	/go/src/github.com/submariner-io/submariner-operator/pkg/subctl/cmd/join.go:147 +0x22f
github.com/submariner-io/submariner-operator/pkg/subctl/cmd.glob..func2(0x215d2e0, 0xc000291b80, 0x1, 0x1)
	/go/src/github.com/submariner-io/submariner-operator/pkg/subctl/cmd/join.go:85 +0x162
github.com/spf13/cobra.(*Command).execute(0x215d2e0, 0xc000291b60, 0x1, 0x1, 0x215d2e0, 0xc000291b60)
	/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/spf13/cobra/command.go:830 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0x215cde0, 0x0, 0x12bb040, 0xc00004c118)
	/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/spf13/cobra/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/spf13/cobra/command.go:864
github.com/submariner-io/submariner-operator/pkg/subctl/cmd.Execute(...)
[...]

PR #113, opened 2019-12-19
Avoid crash when no network discovery info

mangelajo:
Fixes issue: #112


mangelajo:
merging since it's an obvious tiny fix

Issue #86, opened 2019-12-05
submariner-operator strongswan connection 10 minutes delay


manosnoam:
After running subctl join on the 2nd cluster, it took ~10 minutes to establish connection.
subctl should indicate about this delay, and print when strongswan status is up.

```
[nmanos@nmanos submariner-operator]$ kubconf_a  subctl join --clusterid subm-cluster-a ./broker-info.subm --ikeport 501 --nattport 4501
* ./broker-info.subm says broker is at: https://api.nmanos-cluster-a.devcluster.openshift.com:6443
* There are 1 labeled nodes in the cluster:
  - ip-10-0-89-164.ec2.internal
* Deploying the submariner operator
* Created operator CRDs.
* Created operator namespace: submariner-operator
* Created operator service account and role
* Updated the privileged SCC
* Deployed the operator successfully
* The operator is up and running
* Discovering network details
    Discovered network details:
        Network plugin:  OpenShift
        ClusterIP CIDRs: [172.30.0.0/16]
        Pod CIDRs:       [10.128.0.0/14]
* Deploying Submariner
* Submariner is up and running


[nmanos@nmanos submariner-operator]$ kubconf_a  oc get crds | grep submariner
clusters.submariner.io                                      2019-12-05T20:23:31Z
endpoints.submariner.io                                     2019-12-05T20:23:31Z
submariners.submariner.io                                   2019-12-05T20:24:24Z


[nmanos@nmanos submariner-operator]$   kubconf_b subctl join --clusterid subm-cluster-b ./broker-info.subm --ikeport 501 --nattport 4501
* ./broker-info.subm says broker is at: https://api.nmanos-cluster-a.devcluster.openshift.com:6443
* There are 1 labeled nodes in the cluster:
  - nmanos-cl2-tsdqz-worker-tdkpm
* Deploying the submariner operator
* Created operator CRDs.
* Created operator namespace: submariner-operator
* Created operator service account and role
* Updated the privileged SCC
* Deployed the operator successfully
* The operator is up and running
* Discovering network details
    Discovered network details:
        Network plugin:  OpenShift
        ClusterIP CIDRs: [100.96.0.0/16]
        Pod CIDRs:       [10.252.0.0/14]
* Deploying Submariner
* Submariner is up and running


[nmanos@nmanos submariner-operator]$ kubconf_b  oc get crds | grep submariner
clusters.submariner.io                                      2019-12-05T20:25:39Z
[...]

PR #114, opened 2019-12-19
KIND spinners

skitt:
These commits shouldn’t be squashed on merge (I’d like to keep the “import” commit standalone, so that our changes can be more easily rebased if necessary).


skitt:
> This looks cool, will it show spinner status during subctl commands,
> and only when submariner is ready - return to shell ?
> 
> Will it also resolve #86 ?

It doesn’t change the existing behaviour, only the display: while `subctl` is working, a spinner informs the user that _something_ is still happening.

Issue #72, opened 2019-12-01
Please add subctl --version


manosnoam:
At the moment this is the output:

[nmanos@nmanos submariner-operator]$ subctl --version
Error: unknown flag: --version
Usage:
  subctl [command]

Available Commands:
  deploy-broker set the broker up
  help          Help about any command
  info          Get information on your cluster related to submariner
  join          connect a cluster to an existing broker

Flags:
  -h, --help                help for subctl
      --kubeconfig string   absolute path(s) to the kubeconfig file(s) (default "/home/nmanos/.kube/config")

Use "subctl [command] --help" for more information about a command.

panic: unknown flag: --version

goroutine 1 [running]:
main.main()
	/go/src/github.com/submariner-io/submariner-operator/pkg/subctl/main.go:8 +0x84

mangelajo:
Makes sense! :)

mangelajo:
Added:

```subctl version```

as openshift-install version, ...

PR #117, opened 2019-12-20
Add subctl version command

mangelajo:
Adds the version command, where the version tag and the
commit id of the build is exposed for binary traceability.

Fixes: #72


pkg/subctl/cmd/join.go
```diff
@@ -38,7 +38,7 @@ var (
 	serviceCIDR     string
 	clusterCIDR     string
 	repository      string
-	version         string
+	imageVersion    string
```
mangelajo: This field is modified to avoid overlap with the pkg import happening on version.go over the same package:

"github.com/submariner-io/submariner-operator/pkg/version"

mangelajo:
I added a missing file pkg/version.go

mangelajo:
> I imagine dumping the cluster info is a good idea, but the 30k LoC log is a bit rough lol. Finally found the error:

Oh, it's not, I added it but it's horrible, we should get rid of it ':D I agree
> 
> > panic: unknown flag: --version

hhmm, let me check thanks

mangelajo:
ping @dfarrell07 this should be ready now

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-openapi/spec v0.19.5
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/golang/groupcache v0.0.0-20191027212112-611e8accdfc9 // indirect
+	github.com/google/btree v1.0.0 // indirect
```
dfarrell07: I guess this is for my learning as much as anything else, but - how do we know this is necessary, and in general, do we have a way to verify all our dependencies are still necessary? I removed this line locally and was still able to successfully run `make build && make test validate && make ci && make e2e status=keep`. Thanks!

skitt: `go build` adds missing dependencies if necessary, so removing indirect dependencies from `go.mod` doesn’t affect the build. Indirect dependencies are transitive dependencies pulled in by dependencies listed in `go.mod` which don’t yet use modules themselves; they are added to our `go.mod` to ensure that it provides a full record of our module dependencies. See [this entry in the modules wiki](https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod) for details.

To address your main question, `go mod tidy` is the reference tool for maintaining `go.mod`. It will take care of removing obsolete dependencies.

PR #119, opened 2019-12-20
Remove IPSEC secret storage in broker

mangelajo:
Subctl should not store the IPSEC secret in the broker, because
even if the broker is compromised, still without the IPSEC factor
an attacking actor could not connect to the remote clusters.

To minimize the risk of IPSEC key loss the broker-info.subm file
is backed up, and if the file is found locally, it's IPSEC PSK key
is reused. Also can be provided via a cmdline flag "--ipsec-psk-from"
Fixes: #93

### DIFF
--- a/pkg/subctl/datafile/backup.go
+++ b/pkg/subctl/datafile/backup.go
@@ -1,0 +1,17 @@
+package datafile
+
+import (
+	"os"
+	"strings"
+	"time"
+)
+
+func BackupIfExists(filename string) (string, error) {
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return "", nil
+	}
+	now := time.Now()
+	nowStr := strings.Replace(now.Format(time.RFC3339), ":", "_", -1)
+	newFilename := filename + "." + nowStr
+	return newFilename, os.Rename(filename, newFilename)
+}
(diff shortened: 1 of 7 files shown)


pkg/subctl/datafile/backup.go
```diff
@@ -0,0 +1,17 @@
...
+	"os"
+	"strings"
+	"time"
+)
+
+func BackupIfExists(filename string) (string, error) {
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return "", nil
+	}
+	now := time.Now()
```
dfarrell07: Nit: seems cleaner just to embed this in the next line vs having a dedicated var

mangelajo:
> Can you explain more about the use case ?
> Before this change, was the secret stored elsewhere (besides broker-info.subm) ?
> Is it now User responsibility to backup broker-info.subm ?

Correct,

This was discussed on the upstream meeting, the idea is that the broker should not have a copy of the PSK (which we do now on broker deploy).

If you store the PSK in the broker, then, if your broker is compromised, anybody can connect to your clusters, ... but if you don't store it (this change) then, it's still not possible to connect to the cluster networks because you're missing the second piece, the PSK.

PR #129, opened 2020-01-10
Install the KubeFed operator

skitt:


mangelajo:
We could also modify the object once the yaml is loaded. The idea was to make it easy to update stuff if the yaml changed because of the operator-sdk. But... this also makes sense to me.

pkg/subctl/operator/install/deployment/ensure.go
```diff
@@ -41,19 +40,57 @@ func Ensure(restConfig *rest.Config, namespace string, image string) (bool, erro
...
-	deployment := &appsv1.Deployment{}
-
-	err = embeddedyamls.GetObject(embeddedyamls.Operator_yaml, &deployment)
-
-	deployment.Spec.Template.Spec.Containers[0].Image = image
-
+	replicas := int32(1)
+	imagePullPolicy := v1.PullAlways
 	// If we are running with a local development image, don't try to pull from registry
-	if image == "submariner-operator:local" {
```
mangelajo: Oh  I though this was the reason, what's the advantage of converting it to explicit?

skitt: I thought it would be interesting to have a single function for all operators (Submariner and KubeFed).

skitt:
BTW please don’t squash the commits when merging this, they are supposed to be separate units.

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -49,14 +65,19 @@ func main() {
...
 
 		_, err = out.WriteString("\t" + constName(f) + " = `")
 		panicOnErr(err)
 
 		fmt.Println(f)
-		f, _ := os.Open(path.Join(yamlsDirectory, f))
-		_, err = io.Copy(out, f)
+		contents, err := ioutil.ReadFile(path.Join(yamlsDirectory, f))
+		panicOnErr(err)
+
```
mangelajo: Can we add a comment explaining why we do this? I guess it’s related to comments , strings or annotations or something?

pkg/subctl/operator/common/operatorpod/ensure.go
```diff
@@ -14,10 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package deployment
+package operatorpod
```
mangelajo: Nit: it’s it more like operator deployment? :)

PR #132, opened 2020-01-14
Add lighthouse option to subctl

vthapar:
1. Adds 'service-discovery' option to subctl when deploying broker
2. Adds the parameter to subm file so can be used when joining clusters

### DIFF
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -20,11 +20,11 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-	lighthouse "github.com/submariner-io/submariner-operator/pkg/subctl/lighthouse/deploy"
 
 	"github.com/submariner-io/submariner-operator/pkg/broker"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/datafile"
+	lighthouse "github.com/submariner-io/submariner-operator/pkg/subctl/lighthouse/deploy"
 )
 
 var (
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -28,14 +28,14 @@ import (
 
 var enableDataplane bool
 var disableDataplane bool
-var enableLighthouse bool
+var serviceDiscovery bool
 
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&enableDataplane, "dataplane", false,
 		"Install the Submariner dataplane on the broker")
 	deployBroker.PersistentFlags().BoolVar(&disableDataplane, "no-dataplane", true,
 		"Don't install the Submariner dataplane on the broker (default)")
-	deployBroker.PersistentFlags().BoolVar(&enableLighthouse, "service-discovery", false,
+	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", false,
 		"Enable Multi Cluster Service Discovery")
 	err := deployBroker.PersistentFlags().MarkHidden("no-dataplane")
 	// An error here indicates a programming error (the argument isn’t declared), panic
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@ deploy/crds/submariner_endpoints_crd.yaml
 output/
 vendor/
 *.coverprofile
+.idea
```
skitt: Please leave this out from this PR.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -28,12 +28,15 @@ import (
 
 var enableDataplane bool
 var disableDataplane bool
+var enableLighthouse bool
```
skitt: Since the user-facing option is generic, I’d recommend keeping the variable generic too:

    var enableServiceDiscovery bool

pkg/subctl/datafile/datafile.go
```diff
@@ -32,6 +32,7 @@ type SubctlData struct {
 	BrokerURL   string     `json:"brokerURL"`
 	ClientToken *v1.Secret `omitempty,json:"clientToken"`
 	IPSecPSK    *v1.Secret `omitempty,json:"ipsecPSK"`
+	Lighthouse  bool `omitempty,json:"serviceDiscovery"`
```
skitt: Same here, let’s make the variable generic.

pkg/subctl/datafile/datafile.go
```diff
@@ -73,7 +74,7 @@ func NewFromFile(filename string) (*SubctlData, error) {
 	return NewFromString(string(dat))
 }
 
-func NewFromCluster(restConfig *rest.Config, brokerNamespace string) (*SubctlData, error) {
+func NewFromCluster(restConfig *rest.Config, brokerNamespace string, enableLighthouse bool) (*SubctlData, error) {
```
skitt: This is supposed to retrieve the information stored in the cluster, which doesn’t cover service discovery; I think we should avoid adding a parameter here, and set the value explicitly after the call to `NewFromCluster`:

    subctlData, err := datafile.NewFromCluster(config, broker.SubmarinerBrokerNamespace)
    subctlData.ServiceDiscovery = enableServiceDiscovery

skitt:
BTW I really like the idea of making this `--service-discovery` rather than `--lighthouse`!

pkg/subctl/datafile/datafile.go
skitt: Either that, or make the function find out from the broker that it needs to enable service discovery.

pkg/subctl/cmd/deploybroker.go
vthapar: Was in two minds, decided to go this way to get inputs in review.

.gitignore
vthapar: Rebase takes care of it.

pkg/subctl/datafile/datafile.go
vthapar: Going with setting value after call. Need to add it to subctlData so it shows up in `broker-info.subm`

pkg/subctl/cmd/deploybroker.go
```diff
@@ -20,15 +20,20 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+	lighthouse "github.com/submariner-io/submariner-operator/pkg/subctl/lighthouse/deploy"
```
skitt: This should be grouped with the Submariner imports in the following section.

vthapar: I expected the plugin in IDE to take care of this, looks like it doesn't. Done.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -24,16 +24,28 @@ import (
...
+)
 
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&enableDataplane, "dataplane", false,
 		"Install the Submariner dataplane on the broker")
 	deployBroker.PersistentFlags().BoolVar(&disableDataplane, "no-dataplane", true,
 		"Don't install the Submariner dataplane on the broker (default)")
+	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", false,
+		"Enable Multi Cluster Service Discovery")
+	deployBroker.PersistentFlags().StringVar(&serviceDiscoveryImageRepo, "service-discovery-repo", "",
```
skitt: We’ll have to specify defaults here, but that can wait until we have the images we want in the first release.

mangelajo: seems ok as a start, we can model the flags more later when we have a better idea of the experience, and eventually this can be disable-service-discovery ;)

PR #134, opened 2020-01-21
Add lighthouse option to e2e

dfarrell07:


scripts/kind-e2e/e2e.sh
```diff
@@ -69,10 +70,15 @@ function setup_custom_cni(){
 }
 
 function setup_broker() {
+
     if kubectl --context=cluster1 get crd clusters.submariner.io > /dev/null 2>&1; then
         echo Submariner CRDs already exist, skipping broker creation...
     else
         echo Installing broker on cluster1.
+        if [[ $lighthouse = true ]]; then
```
skitt: There’s no need to run `deploy-broker` twice when deploying with Lighthouse, so

     sd=
     ../bin/subctl --kubeconfig ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster1 deploy-broker ${sd}

should work.

The option also needs to be added to the `join` commands.

vthapar: Done

Issue #116, opened 2019-12-20
Spinner pollutes CI log sometimes.


mangelajo:
I believe kind disables that on the absence of a real tty:

https://travis-ci.com/submariner-io/submariner-operator/builds/141770889#L3914

```
[scripts]$ ../bin/subctl join --operator-image submariner-operator:local --kubeconfig /go/src/github.com/submariner-io/submariner-operator/output/kind-config/dapper/kind-config-cluster3 --clusterid cluster3 --repository local --version local --nattport 4500 --ikeport 500 --colorcodes blue --disable-nat broker-info.subm

* broker-info.subm says broker is at: https://172.17.0.3:6443

* There are 1 labeled nodes in the cluster:

  - cluster3-worker

[...]

PR #138, opened 2020-01-23
Force subctl output through a pipe to avoid spinner in logs

mangelajo:
Also remove some duplication around subctl usage in the e2e.sh script.

Fixes: #116


mangelajo:
big mixup, don't merge yet

mangelajo:
ok, should be good now

PR #140, opened 2020-01-23
Update Submariner Engine to a DaemonSet

tpantelis:
See https://github.com/submariner-io/submariner-charts/pull/13 for details.


tpantelis:
> This means we need to bump the CRD version (`v1alpha2`).



> This means we need to bump the CRD version (`v1alpha2`).

Is that really necessary at this stage in development? It wouldn't affect any downstream until we create a new release with these changes (unless a downstream is using master). Also it is alpha so API changes should be expected, plus I wouldn't expect anyone's actually using the operator for real yet.

skitt:
> Is that really necessary at this stage in development? It wouldn't affect any downstream until we create a new release with these changes (unless a downstream is using master). Also it is alpha so API changes should be expected, plus I wouldn't expect anyone's actually using the operator for real yet.

I think we need to handle this if we want to publish an update to OperatorHub, since the CRD has been published there. Another option would be to not change the CRD and ignore the `count` value...

mangelajo:
I wanted to raise the concern, now, as Tom says, may be is safe to do without hurting anyone. I was more meaning that may be we should delete existing Deployment after deploying "Daemonset", to make sure they won't run side-by-side on an updated environment.

I think the operatorhub/OLM options should be ditched. That's not really usable.

I was talking the other day to the storage guys, and they showed me they have the upstream operator, and then, a meta-operator.

Our "meta-operator" is more or less subctl...  may be at some point we can create some sort of top-level operator that will be mostly subctl with an interface.

skitt:
Right, the reconciliation should delete any existing deployment if it finds one.

I agree that we won’t be hurting anyone, I’m fine with dropping the old version.

tpantelis:
> I wanted to raise the concern, now, as Tom says, may be is safe to do without hurting anyone. I was more meaning that may be we should delete existing Deployment after deploying "Daemonset", to make sure they won't run side-by-side on an updated environment.
> 

So do this in the operator controller ...  yeah I can do that.
 
> I was talking the other day to the storage guys, and they showed me they have the upstream operator, and then, a meta-operator.
> 
> Our "meta-operator" is more or less subctl... may be at some point we can create some sort of top-level operator that will be mostly subctl with an interface.

yeah you would expect users to use _subctl_ rather than directly using the Submariner CRD.

I'll create a v1alpha2 if you guys really think it's necessary but I don't think it is since it is alpha.

mangelajo:
@tpantelis @skitt 
I tested as-is, and it does not break or cause any harm.

I verified that the CRD is updated, and that the operator runs happily after.

We only need to add the removal of the old deployment item, I would do a) add daemonset b) remove deployment

See the following output, k8s is smart enough to not duplicate the pods, as both the daemonset and the deployment are satisfied:

```
[majopela@bluehat submariner-operator]$ kubectl get daemonsets -n submariner-operator
NAME                    DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                AGE
submariner              1         1         0       1            0           submariner.io/gateway=true   5s
submariner-routeagent   2         2         2       2            2           <none>                       24m
[majopela@bluehat submariner-operator]$ oc get pods
NAME                       READY   STATUS    RESTARTS   AGE
netshoot-785ffd8c8-ql2vv   1/1     Running   0          23m
[majopela@bluehat submariner-operator]$ kubectl get deployments -n submariner-operator
NAME                  READY   UP-TO-DATE   AVAILABLE   AGE
submariner            1/1     1            1           24m
submariner-operator   1/1     1            1           24m

```


Also verification:

```
[majopela@bluehat submariner-operator]$ kubectl get crd submariners.submariner.io -o yaml | grep clusterID
            clusterID:
          - clusterID
[majopela@bluehat submariner-operator]$ kubectl get crd submariners.submariner.io -o yaml | grep count
[majopela@bluehat submariner-operator]$ kubectl get submariner submariner -n submariner-operator | grep count
[majopela@bluehat submariner-operator]$ 

``
No kitties were harmed during the testing of this patch.

**Comment by skitt** [comment]
> I'll create a v1alpha2 if you guys really think it's necessary but I don't think it is since it is alpha.

No need from my perspective. Having thought about this a bit more, I now think we need to figure out our upgrade story in detail, with the interactions between `subctl`, the operator, and Submariner (and Lighthouse, and KubeFed, and ...), before we embark on piecemeal changes.

**Comment by tpantelis** [comment]
Submitted https://github.com/submariner-io/submariner-operator/pull/140/commits/c65d0db93d87780239fb573329c63b89ae283add to delete engine deployment.

File: pkg/controller/submariner/submariner_controller.go [hunk 1]
```diff
@@ -193,39 +170,32 @@ func (r *ReconcileSubmariner) reconcileRouteagentDaemonSet(instance *submarinerv
...
-	maxSurge := intstr.FromInt(1)
-	maxUnavailable := intstr.FromInt(0)
+	maxUnavailable := intstr.FromInt(1)
 
-	deployment := &appsv1.Deployment{
+	deployment := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:    labels,
 			Namespace: cr.Namespace,
 			Name:      "submariner",
```
mangelajo: Sorry for the late comment, it's something we can handle in a follow up PR as it's unrelated.

Some time ago we mentioned D/S and U/S about changing submariner to submariner-gateway to make it more evident, as some people was confused by what the gateways where. 

I will add a follow up PR to here and the helm charts for uniformity before next release.

sridhargaddam: Submariner-gateway would be apter. @mangelajo ++

PR #143, opened 2020-01-24
Ensure embedded YAMLs are generated if yamls2go changes

skitt:
Currently, we rebuild the embedded YAMLs if the source YAMLs change;
but since the list of YAMLs is contained in yamls2go, we need to watch
that too.


mangelajo:
Silly me I should have not merged this here until the other base patch is merged...

PR #145, opened 2020-01-28
Rename submariner engine daemonset to submariner-gateyway

mangelajo:
In previous upstream discussions we talked about renaming submariner
engine "deployment" (now daemonset) to submariner-gateway to make
identification of the gateway easier to admins and avoid confusion.

Also see: https://github.com/submariner-io/submariner-charts/pull/14


pkg/controller/submariner/submariner_controller.go
```diff
@@ -204,7 +204,7 @@ func newEngineDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet {
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:    labels,
 			Namespace: cr.Namespace,
-			Name:      "submariner",
+			Name:      "submariner-gateway",
```
tpantelis: UTs need to be updated too. Perhaps we should make this a constant?

mangelajo: Oh, I had a look on UTs and didn't see a check on the name.

It's better not to use a constant for UTs to reference. That provides double validation (if you make a mistake in one of the sides, or change the constant without noticing, the UTs will chime at you)

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
 	github.com/json-iterator/go v1.1.8 // indirect
 	github.com/kr/pty v1.1.8 // indirect
-	github.com/mattn/go-isatty v0.0.10
+	github.com/mattn/go-isatty v0.0.11
```
mangelajo: oops, that's unrelated.

tpantelis: yeah - I keep getting the same staged changes locally too but I was not checking it in. Since your getting it too might as well check it in to avoid the hassle.

mangelajo:
ping @sridhargaddam @skitt @dfarrell07

go.mod
skitt: Or we could merge https://github.com/submariner-io/submariner-operator/pull/136 ...

mangelajo:
Oh right, I must remove that, Is unrelated sorry for the ping.

I’ll ping you as soon it’s fixed.

Enviado desde mi iPhone

mangelajo:
Cleaned up the go.mod parts.

PR #152, opened 2020-01-30
Support specifying contexts

skitt:
... as well as the kubeconfig file. This will be necessary for KIND
upgrades since KIND (and Kubernetes in general) is moving away from
separate kubeconfig files.


mangelajo:
This chimes in CI 
```
pkg/subctl/cmd/join.go:142:4: ineffectual assignment to `clusterID` (ineffassign)
			clusterID = answers.ClusterID

```
but I didn't check why

skitt:
> Makes sense, does it also work if you don't provide the context?

Yes, we only override the context if one is provided.

> And an idea for the future: we could also support the
> 
> KUBECONFIG=kubecfg1:kubecfg2:kubecfg3 notation, see [1]
> 
> [1] https://github.com/submariner-io/submariner/blob/5af282d10b1005e26fce514822109b30e6397fb4/test/e2e/framework/util.go#L18

Ah yes, I considered that and didn’t realise it was already supported in Submariner ;-). I’ll do that in a follow-up patch.

PR #153, opened 2020-01-31
kubefed join

aswinsuryan:
adds kubefed join functionality to subctl

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,10 +4,13 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner-operator DAPPER_DOCKER_SOCKET=true \
     TRASH_CACHE=${DAPPER_SOURCE}/.trash-cache HOME=${DAPPER_SOURCE} DAPPER_OUTPUT=output \
-    OPERATOR_SDK_VERSION=0.12.0 GOROOT=/usr/lib/golang
+    OPERATOR_SDK_VERSION=0.12.0 GOROOT=/usr/lib/golang \
+    KUBEFED_VERSION=0.1.0-rc3
 
 RUN curl -Lo /usr/bin/operator-sdk "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu" && \
-    chmod a+x /usr/bin/operator-sdk
+    chmod a+x /usr/bin/operator-sdk && \
+    curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v${KUBEFED_VERSION}/kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz" && \
+    tar -xzf kubefedctl-${KUBEFED_VERSION}-linux-${ARCH}.tgz && cp kubefedctl /usr/bin/ && chmod a+x /usr/bin/kubefedctl
 
 WORKDIR ${DAPPER_SOURCE}
 
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"os/exec"
 	"regexp"
 	"strings"
 
@@ -36,18 +37,19 @@ import (
 )
 
 var (
-	clusterID       string
-	serviceCIDR     string
-	clusterCIDR     string
-	repository      string
-	imageVersion    string
-	nattPort        int
-	ikePort         int
-	colorCodes      string
-	disableNat      bool
-	ipsecDebug      bool
-	submarinerDebug bool
-	noLabel         bool
+	clusterID                 string
+	serviceCIDR               string
+	clusterCIDR               string
+	repository                string
+	imageVersion              string
+	nattPort                  int
+	ikePort                   int
+	colorCodes                string
+	disableNat                bool
+	ipsecDebug                bool
+	submarinerDebug           bool
+	noLabel                   bool
+	kubefedHostClusterContext string
 )
 
 func init() {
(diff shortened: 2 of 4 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -71,6 +73,7 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ipsecDebug, "ipsec-debug", false, "Enable IPsec debugging (verbose logging)")
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
+	cmd.Flags().StringVar(&kubefedController,  "kubefed-controller", "" , "Kubefed Controller Context Name")
```
vthapar: Use same descrption as used by kubefedctl:
 `--host-cluster-context string   Host cluster context`
It isn't clear that it needs the name of the host.

Also, this might be better option:
`--host-cluster-name string      If set, overrides the use of host-cluster-context name in resource names created in the target cluster. This option must be used when the context name has characters invalid for kubernetes resources like "/" and ":".`

>

pkg/subctl/cmd/join.go
```diff
@@ -163,6 +166,12 @@ func joinSubmarinerCluster(subctlData *datafile.SubctlData) {
 		err = lighthouse.Ensure(status, config, "", "", false)
 		status.End(err == nil)
 		exitOnError("Error deploying multi cluster service discovery", err)
+
+		status.Start("Joining to Kubefed control plane")
+		cmd := exec.Command("kubefedctl", "join", "--kubefed-namespace", "kubefed-operator", clusterID, "--host-cluster-context", kubefedController)
```
vthapar: You can use something like:
`_, err := exec.Command("kubefedctl", "join", "--kubefed-namespace", "kubefed-operator", clusterID, "--host-cluster-context", kubefedController).Output()

Since you're running single command, you don't really need CombinedOutput

pkg/subctl/cmd/join.go
```diff
@@ -36,18 +37,19 @@ import (
...
+	repository        string
+	imageVersion      string
+	nattPort          int
+	ikePort           int
+	colorCodes        string
+	disableNat        bool
+	ipsecDebug        bool
+	submarinerDebug   bool
+	noLabel           bool
+	kubefedController string
```
vthapar: Add parameter to e2e.sh also.

pkg/subctl/cmd/join.go
aswinsuryan: Doesn't that look little hard to read, since the entire command is a set of strings than variable?

pkg/subctl/cmd/join.go
aswinsuryan: Done.

pkg/subctl/cmd/join.go
aswinsuryan: Done.

Dockerfile.dapper
```diff
@@ -7,7 +7,9 @@ ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPE
     OPERATOR_SDK_VERSION=0.12.0 GOROOT=/usr/lib/golang
 
 RUN curl -Lo /usr/bin/operator-sdk "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu" && \
-    chmod a+x /usr/bin/operator-sdk
+    chmod a+x /usr/bin/operator-sdk && \
+    curl -LO "https://github.com/kubernetes-sigs/kubefed/releases/download/v0.1.0-rc3/kubefedctl-0.1.0-rc3-linux-amd64.tgz" && \
```
skitt: Could you make the version a variable, as in `Dockerfile.dapper-base`?

pkg/subctl/cmd/join.go
```diff
@@ -71,6 +72,7 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ipsecDebug, "ipsec-debug", false, "Enable IPsec debugging (verbose logging)")
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
+	cmd.Flags().StringVar(&kubefedController, "kubefed-controller", "", "overrides the use of host-cluster-context name in resource names created in the target cluster.")
```
tpantelis: The name "kubefed-controller" seems out of context with the description - how about we name it "kubefed-host-cluster-context"? Similarly the variable "kubefedHostClusterContext".

pkg/subctl/cmd/join.go
vthapar: It is same thing, just combining Command and CombinedOutput() in single line.

Dockerfile.dapper
aswinsuryan: Done.

pkg/subctl/cmd/join.go
aswinsuryan: Done.

pkg/subctl/cmd/join.go
aswinsuryan: Done.

pkg/subctl/cmd/join.go
```diff
@@ -71,6 +73,7 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ipsecDebug, "ipsec-debug", false, "Enable IPsec debugging (verbose logging)")
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
+	cmd.Flags().StringVar(&kubefedHostClusterContext, "kubefedHostClusterContext", "", "overrides the use of host-cluster-context name in resource names created in the target cluster.")
```
```suggestion
	cmd.Flags().StringVar(&kubefedHostClusterContext, "kubefed-host-cluster-context", "", "overrides the use of host-cluster-context name in resource names created in the target cluster.")
```

To follow CLI arg naming conventions.

aswinsuryan: Oh just forgot to push the updated one. Done.

scripts/kind-e2e/e2e.sh
```diff
@@ -77,7 +77,7 @@ function setup_broker() {
         echo Installing broker on cluster1.
          sd=
          [[ $lighthouse = true ]] && sd=--service-discovery
-         ../bin/subctl --kubeconfig ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster1 deploy-broker ${sd} |& cat
+         ../bin/subctl --kubeconfig ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster1 deploy-broker --service-discovery ${sd} |& cat
```
vthapar: This is not needed. $sd is set to service discovery in previous like already if lighthouse is true.

aswinsuryan: Done.

Issue #126, opened 2020-01-06
make build-subctl doesn’t run the embedded generator


skitt:
rm pkg/subctl/operator/install/embeddedyamls/yamls.go
    make build-subctl

fails, it shouldn’t.

skitt:
cc @Jaanki

mangelajo:
@skitt I believe you fixed this in one of your last refactors/PRs, am I right?

mangelajo:
please close if it was really fixed.

skitt:
No, it’s not fixed yet. (The workaround is to run `make bin/subctl` which does the right thing.)

mangelajo:
Ah, but I think, the right way should be make bin/subctl, which handles the dependency logic, may be we should remove build-subctl from the runnable scripts through make, I'd do that.

manosnoam:
Also we should mention in Readme that it is required to get vendor modules before the build:
```
GO111MODULE="on" go mod vendor
make bin/subctl
```

skitt:
> Also we should mention in Readme that it is required to get vendor modules before the build:
> 
> ```
> GO111MODULE="on" go mod vendor
> make bin/subctl
> ```

No need, that was fixed in #174.

PR #157, opened 2020-02-03
Describe subctl dependencies on build-subctl

skitt:
Targets can have multiple sets of dependencies, so we can specify
build-subctl’s dependencies, and specify that bin/subctl depends on
that.

(This can’t be extended too far, otherwise we end up with too many
phony dependencies in the chain, and everything gets rebuilt all the
time.)

Fixes: #126

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,9 @@ $(TARGETS): .dapper
 shell: .dapper
 	./.dapper -s -m bind
 
-bin/subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
-	$(MAKE) build-subctl
+build-subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
+
+bin/subctl: build-subctl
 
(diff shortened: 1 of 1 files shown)


skitt:
> Won’t that always force it to run because ‘build-subctl’ will never exist?

Yes, `make bin/subctl` will always rebuild the binary. So we could repeat the prerequisites to avoid that, I guess...

The real issue is that we’re mixing the build scripts and file-based rules. `build` runs the generators, `build-cross` and `build-subctl` don’t, and `ci` ends up running them twice!

mangelajo:
> > Won’t that always force it to run because ‘build-subctl’ will never exist?
> 
> Yes, `make bin/subctl` will always rebuild the binary. So we could repeat the prerequisites to avoid that, I guess...
> 
> The real issue is that we’re mixing the build scripts and file-based rules. `build` runs the generators, `build-cross` and `build-subctl` don’t, and `ci` ends up running them twice!

Yeah, I was trying to turn it more into file based, but I'm not sure if it's a doable target or not..

mangelajo:
Can we may be completely remove build-subctl as a possible make target at all?


So if you want to run build-subctl... you use make bin/subctl

May be instead of listing scripts and making them available to make, we should move into adding them explicitly instead.

PR #161, opened 2020-02-04
Set the DNS operator up for Lighthouse

skitt:

### DIFF
--- a/pkg/subctl/lighthouse/dns/ensure.go
+++ b/pkg/subctl/lighthouse/dns/ensure.go
@@ -1,0 +1,299 @@
+/*
+© 2020 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package lighthousedns
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/subctl/lighthouse/dns/ensure.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	clientsetifc "k8s.io/client-go/kubernetes/typed/apps/v1"
+	"k8s.io/client-go/rest"
+
+	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
+)
+
+const (
+	operatorImage           = "quay.io/aswinsuryan/lighthouse-cluster-dns-operator:v0.0.1"
```
mangelajo: Did we also need to modify the cluster-dns-operator? :)

pkg/subctl/lighthouse/dns/ensure.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+func Ensure(status *cli.Status, config *rest.Config) error {
+	clientSet, err := clientset.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	// Disable ClusterVersionOperator
+	_, err = scaleDeployment(clientSet.AppsV1().Deployments("openshift-cluster-version"), "cluster-version-operator", 0)
```
mangelajo: Isn't this very openshift specific?, what about others? Can we add comments about the reasoning for this?

pkg/subctl/lighthouse/dns/ensure.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
+	}
+	status.QueueSuccessMessage("Disabled the cluster version operator")
+
+	// Fix up cluster role
+	clusterRole, err := clientSet.RbacV1().ClusterRoles().Get("openshift-dns-operator", metav1.GetOptions{})
```
mangelajo: Same here. We need to think about how to do this in other cases, not only satisfy openshift, although I understand probably for a POC is what we have, we cannot consider this feature finished or ready until we handle something that would work community wide.

skitt: Yes, this whole patch is OpenShift-specific and very hacky. We shouldn’t have to deal with any of this stuff, and the cluster ends up somewhat broken as a result...

skitt:
This should now be transparent on non-OpenShift deployments (I’ll check tomorrow).

PR #169, opened 2020-02-13
Check kubefedctl

skitt:


pkg/subctl/lighthouse/deploy/ensure.go
```diff
@@ -17,22 +17,77 @@ limitations under the License.
...
+		v := version.Info{}
+		err = json.Unmarshal([]byte(output), &v)
+		if err != nil {
+			return err
+		}
+		if v.Version != kubefedVersion {
+			return fmt.Errorf("invalid kubefedctl version %s, please use %s", v.Version, kubefedVersion)
+		}
+	}
+	return errors.New("")
```
vthapar: why not return nil?

skitt: Hah, thanks, I was using that to check the new code and forgot to remove it!

pkg/subctl/lighthouse/deploy/ensure.go
```diff
@@ -17,22 +17,76 @@ limitations under the License.
...
 	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/kubefed"
 )
 
 const (
-	DefaultControllerImageName    = "lighthouse-controller"
-	DefaultControllerImageRepo    = "quay.io/submariner"
-	DefaultControllerImageVersion = "0.0.1"
+	defaultControllerImageName    = "lighthouse-controller"
+	defaultControllerImageRepo    = "quay.io/submariner"
+	defaultControllerImageVersion = "0.0.1"
```
mangelajo: Can we bump the version to 0.1.0 in lighthouse so we sync them all?

skitt: Yes, once the Lighthouse release is tagged, we’ll need a version-bump PR in the operator.

vthapar:
> Just a tiny comment about lighthouse version, can we bump that now to 0.1.0 ? is it ready?

That will happen in separate patch and rebase can take care of it.

mangelajo:
conflict hopefully resolved ok

skitt:
> I changed it to 0.1.0 but... Isn't that replicated with the yamls?

Please don’t, the version bump has nothing to do with checking for `kubefedctl`!

skitt:
> > I changed it to 0.1.0 but... Isn't that replicated with the yamls?
> 
> Please don’t, the version bump has nothing to do with checking for `kubefedctl`!

Ah, sorry, that was to fix the conflict with the merged bump patch, thanks!

mangelajo:
> > > I changed it to 0.1.0 but... Isn't that replicated with the yamls?
> > 
> > 
> > Please don’t, the version bump has nothing to do with checking for `kubefedctl`!
> 
> Ah, sorry, that was to fix the conflict with the merged bump patch, thanks!

Well you were actually right. But in the end I was lucky and it makes sense after all. :)

PR #170, opened 2020-02-14
subctl: add globalnet options to deploybroker

vthapar:
subctl: add globalnet options to deploybroker
    
Adds following options to allow configuration for
globalnet (overlapping cidrs) when deploying broker:
    
  1. globalnet-enable: Allow clusters with overlapping CIDRs
  2. globalnet-cidr-range: GlobalCIDR supernet range. Default
      169.254.0.0/16
  3. globalnet-cluster-size: Default size of globalnet Subnet in each cluster
      Default is 8192. Should not be more than half of supernet size.


pkg/subctl/cmd/deploybroker.go
```diff
@@ -42,6 +46,13 @@ func init() {
 	deployBroker.PersistentFlags().BoolVar(&disableDataplane, "no-dataplane", true,
 		"Don't install the Submariner dataplane on the broker (default)")
 	lighthouse.AddFlags(deployBroker, "service-discovery")
+
+	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet-enable", false,
+		"Enable support for Overlapping CIDRs in connecting clusters")
+	deployBroker.PersistentFlags().StringVar(&globalnetCidrRange, "globalnet-cidr-range", "169.254.0.0/16",
+		"Global CIDR supernet range for allocating GlobalCIDRs to each cluster")
+	deployBroker.PersistentFlags().Uint8Var(&globalnetClusterSize, "globalnet-cluster-size", 8,
```
mangelajo: Those defaults are problematic, it means that each cluster can only have 256 pods+services.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -42,6 +46,13 @@ func init() {
 	deployBroker.PersistentFlags().BoolVar(&disableDataplane, "no-dataplane", true,
 		"Don't install the Submariner dataplane on the broker (default)")
 	lighthouse.AddFlags(deployBroker, "service-discovery")
+
+	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet-enable", false,
```
sridhargaddam: One option is to get rid of this flag and depend on globalnet-cidr-range alone (i.e., if its present, it means GN is enabled otherwise not). What's your thought on this?
CC: @tpantelis @mangelajo @skitt @mkolesnik

pkg/subctl/cmd/deploybroker.go
sridhargaddam: IPv4 Link-local address is 169.254.0.0/16, so we cannot expand it further.
I think we can do the following, the default value of globalnet-cidr-range can be 169.254.0.0/16 whereas globalnet-cluster-size can be modified to 19 bits.

With this configuration, we can have 8 connected clusters via Submariner and each cluster can have 8K Pods+Services (which I feel is a decent value for defaults). Thoughts?

pkg/subctl/cmd/deploybroker.go
sridhargaddam: Discussed with @vthapar in today's sync call and he suggested that by having globalnet-enable, we can avoid asking the user for globalnet-cidr-range/globalnet-cluster-size as inputs if the user is ok with the default values, which is a good point.

pkg/subctl/cmd/deploybroker.go
sridhargaddam: If user wants to override the defaults, instead of asking the input in networkMask/bits for the cluster, let the user specify the clustersize (i.e., typical number of Pods/Services that he intends to deploy in his cluster). 

For example:
If a user specifies that his clusterSize would be 200, subctl can allocate a globalCIDR of "/8" for that cluster.
If a user specifies his clusterSize is 7500, subctl would allocate the nearest upper bound - i.e., "/19"
and so on...

sridhargaddam: The way I see these flags in use...
globalnet-enable and globalnet-cidr-range are applicable during Broker installation.
globalnet-cluster-size is applicable when joining data clusters.

sridhargaddam: A clarification: globalnet-cluster-size is also applicable during broker installation, but it can be overridden during subctl join operation.

pkg/subctl/cmd/deploybroker.go
vthapar: This is for cases when user just wants to enable globalnet without having to manually specify any CIDR.

mangelajo:
I resolved the conflict, let's see how it takes CI

pkg/subctl/cmd/deploybroker.go
sridhargaddam: Similar to displaying default globalnet-cidr-range and globalnet-cluster-size can you also display that by default globalnet-enable is false.

Current output:
Usage:
  subctl deploy-broker [flags]

Flags:
      --broker-cluster-context string      Broker cluster context
      --clustercidr string                 cluster CIDR
      --clusterid string                   cluster ID used to identify the tunnels
      --colorcodes string                  color codes (default "blue")
      --dataplane                          Install the Submariner dataplane on the broker
      --disable-nat                        Disable NAT for IPsec
      --globalnet-cidr-range string        Global CIDR supernet range for allocating GlobalCIDRs to each cluster (default "169.254.0.0/16")
      --globalnet-cluster-size uint        Default cluster size for GlobalCIDR allocated to each cluster (default 8192)
      --globalnet-enable                   Enable support for Overlapping CIDRs in connecting clusters
  -h, --help                               help for deploy-broker
      --ikeport int                        IPsec IKE port (default 500)
      --ipsec-debug                        Enable IPsec debugging (verbose logging)

pkg/subctl/cmd/deploybroker.go
```diff
@@ -40,6 +44,13 @@ func init() {
 		"Don't install the Submariner dataplane on the broker (default)")
 	// TODO (skitt) make this generic for potentially multiple plugins (see below too)
 	lighthouse.AddFlags(deployBroker, "service-discovery")
+
+	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet-enable", false,
+		"Enable support for Overlapping CIDRs in connecting clusters")
+	deployBroker.PersistentFlags().StringVar(&globalnetCidrRange, "globalnet-cidr-range", "169.254.0.0/16",
+		"Global CIDR supernet range for allocating GlobalCIDRs to each cluster")
+	deployBroker.PersistentFlags().UintVar(&globalnetClusterSize, "globalnet-cluster-size", 8192,
```
sridhargaddam: Can you update the commit message as well (it's stale)

i.e., globalnet-cluster-size: Size in bits of each cluster. Default is 8

pkg/subctl/cmd/deploybroker.go
vthapar: It is picks default from parameters passed and looks like it doesn't show boolean defaults for anything.

sridhargaddam: You can add it as part of the description. i.e., "Enable support for Overlapping CIDRs in connecting clusters (default disabled)"

vthapar: Done

PR #173, opened 2020-02-14
Fix a missing constant replacement

skitt:


mangelajo:
@skitt thanks I have included your patch as part of the other PR. good catch.

PR #182, opened 2020-02-18
subctl: Check for overlapping CIDRs

vthapar:
When joining a new cluster, check if serviceCIDR
or clusterCIDR overlaps with any existing cluster.

This code will also be reused for
Globalnet (overlapping cidrs).

### DIFF
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -1,0 +1,82 @@
+/*
+© 2020 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package globalnet
+
+import (
+	"fmt"
+	"net"
+
+	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GlobalNetwork struct {
+	GlobalCIDRs  []string
+	ServiceCIDRs []string
+	ClusterId    string
+}
+
+func (gn *GlobalNetwork) Show() {
+	if gn == nil {
+		fmt.Println("    No global network details discovered")
+	} else {
+		fmt.Printf("    Discovered global network details for Cluster %s:\n", gn.ClusterId)
+		fmt.Printf("        ServiceCidrs: %v\n", gn.ServiceCIDRs)
+		fmt.Printf("        Global CIDRs: %v\n", gn.GlobalCIDRs)
+
+	}
+}
+
+func ShowNetworks(networks map[string]*GlobalNetwork) {
+	for _, network := range networks {
+		network.Show()
+	}
+}
+
+func Discover(client *submarinerClientset.Clientset, namespace string) (map[string]*GlobalNetwork, error) {
+	clusters, err := client.SubmarinerV1().Clusters(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var globalNetworks = make(map[string]*GlobalNetwork)
+	for _, cluster := range clusters.Items {
+		globalNetwork := GlobalNetwork{
+			GlobalCIDRs:  cluster.Spec.GlobalCIDR,
+			ServiceCIDRs: cluster.Spec.ServiceCIDR,
+			ClusterId:    cluster.Spec.ClusterID,
+		}
+		globalNetworks[cluster.Spec.ClusterID] = &globalNetwork
+	}
+	return globalNetworks, nil
+}
+
+func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
+	_, newNet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return false, err
+	}
+	for _, v := range cidrList {
+		_, baseNet, err := net.ParseCIDR(v)
+		if err != nil {
+			return false, err
+		}
+		if baseNet.Contains(newNet.IP) || newNet.Contains(baseNet.IP) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -27,6 +27,7 @@ import (
 type GlobalNetwork struct {
 	GlobalCIDRs  []string
 	ServiceCIDRs []string
+	ClusterCIDRs []string
 	ClusterId    string
 }
 
@@ -36,6 +37,7 @@ func (gn *GlobalNetwork) Show() {
 	} else {
 		fmt.Printf("    Discovered global network details for Cluster %s:\n", gn.ClusterId)
 		fmt.Printf("        ServiceCidrs: %v\n", gn.ServiceCIDRs)
+		fmt.Printf("        ClusterCidrs: %v\n", gn.ClusterCIDRs)
 		fmt.Printf("        Global CIDRs: %v\n", gn.GlobalCIDRs)
 
 	}
@@ -57,6 +59,7 @@ func Discover(client *submarinerClientset.Clientset, namespace string) (map[stri
 		globalNetwork := GlobalNetwork{
 			GlobalCIDRs:  cluster.Spec.GlobalCIDR,
 			ServiceCIDRs: cluster.Spec.ServiceCIDR,
+			ClusterCIDRs: cluster.Spec.ClusterCIDR,
 			ClusterId:    cluster.Spec.ClusterID,
 		}
 		globalNetworks[cluster.Spec.ClusterID] = &globalNetwork
--- a/pkg/discovery/globalnet/globalnet_test.go
+++ b/pkg/discovery/globalnet/globalnet_test.go
@@ -1,0 +1,99 @@
+/*
+© 2020 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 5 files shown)


pkg/discovery/globalnet/globalnet.go
```diff
@@ -0,0 +1,82 @@
...
+			GlobalCIDRs:  cluster.Spec.GlobalCIDR,
+			ServiceCIDRs: cluster.Spec.ServiceCIDR,
+			ClusterId:    cluster.Spec.ClusterID,
+		}
+		globalNetworks[cluster.Spec.ClusterID] = &globalNetwork
+	}
+	return globalNetworks, nil
+}
+
+func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
```
skitt: This needs a unit test. AFAICT, it only checks whether any of the members of `cidrList` _contains_ `cidr`, not whether they overlap: `IsOverlappingCIDR([]string{"10.10.10.0/24"}, "10.10.0.0/16")` would indicate that they don’t overlap.

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,42 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+
 	status.Start("Deploying Submariner")
 	err = submarinercr.Ensure(config, OperatorNamespace, populateSubmarinerSpec(subctlData))
 	status.End(err == nil)
 	exitOnError("Error deploying Submariner", err)
 }
 
+func checkOverlappingServiceCidr(networks map[string]*globalnet.GlobalNetwork) {
+	for k, v := range networks {
+		yes, err := globalnet.IsOverlappingCIDR(v.ServiceCIDRs, serviceCIDR)
```
skitt: `overlaps` instead of `yes` would make the code more explicit.

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,42 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	clusterCIDR, err = getPodCIDR(clusterCIDR, networkDetails)
 	exitOnError("Error determining the pod CIDR", err)
 
+	fmt.Printf("* Discovering multi cluster details\n")
```
skitt: Please use the status.

pkg/discovery/globalnet/globalnet.go
skitt: Fixing this is easy enough, you can check both `IP` members against the opposite `IPNet`.

vthapar: Your conclusion is correct but reasoning is little off :) going by your example, it is basically checking if 10.10.10.0/24 contains 10.0.0.1 or not. If you were to use `IsOverlappingCIDR([]string{"10.10.0.0/16"}, "10.10.10.0/24")`it would return overlap to be true. I missed a check which I'll add in subsequent patch and it will then work. I will add UTs too.

skitt: My point is that `IsOverlappingCIDR` should check whether they overlap, not whether one contains the other. It shouldn’t matter what the argument order is: while 10.10.10.0/24 doesn’t contain 10.10.0.0/16, it does overlap.

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,45 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		}
+	}
+	return nil
+}
+
+func getGlobalNetworks(subctlData *datafile.SubctlData) map[string]*globalnet.GlobalNetwork {
+
+	brokerConfig := subctlData.GetBrokerConfig()
+	brokerSubmClient, err := submarinerClientset.NewForConfig(brokerConfig)
+	exitOnError("Unable to setup submariner connection to broker cluster", err)
```
sridhargaddam: If I understand it correctly, it means the user has passed an invalid/old broker-info.subm file. If it is so, shall we give a message on those lines?

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,45 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	brokerNamespace := string(subctlData.ClientToken.Data["namespace"])
+	globalNetworks, err := globalnet.Discover(brokerSubmClient, brokerNamespace)
+	exitOnError("Error trying to discover multicluster network details", err)
```
sridhargaddam: nit: multi-cluster

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,45 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+
 	status.Start("Deploying Submariner")
 	err = submarinercr.Ensure(config, OperatorNamespace, populateSubmarinerSpec(subctlData))
 	status.End(err == nil)
 	exitOnError("Error deploying Submariner", err)
 }
 
+func checkOverlappingServiceCidr(networks map[string]*globalnet.GlobalNetwork) error {
+	for k, v := range networks {
+		overlap, err := globalnet.IsOverlappingCIDR(v.ServiceCIDRs, serviceCIDR)
```
sridhargaddam: Why are we checking ONLY serviceCIDR? Ideally AFAIU (when not using globalnet) we should be checking both Service CIDR and Cluster CIDR. Otherwise, there could be issues when SM route-agent program rules where traffic destined to Overlapping Cluster/POD CIDR could be redirected over the tunnel. Even though LH supports the discovery of Services, we should check both POD/Service CIDRs.

pkg/discovery/globalnet/globalnet_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+var _ = Describe("IsOverlappingCidr", func() {
+	When("There are no base CIDRs", func() {
+		overlapping, err := IsOverlappingCIDR([]string{}, "10.10.10.0/24")
+		It("Should return false", func() {
+			Expect(overlapping).To(BeFalse())
+		})
+		It("Should not return error", func() {
+			Expect(err).ToNot(HaveOccurred())
+		})
```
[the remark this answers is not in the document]
vthapar: Doesn't really matter, but will do.

pkg/subctl/cmd/join.go
vthapar: Will do. I refactored the code a bit but forgot to update this msg from a generic one.

pkg/subctl/cmd/join.go
vthapar: Wasn't sure if we want to check ClusterCIDRs or not. Will add.

pkg/discovery/globalnet/globalnet_test.go
```diff
@@ -0,0 +1,99 @@
...
+*/
+
+package globalnet
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("IsOverlappingCidr", func() {
```
sridhargaddam: Are these tests not run as part of CI?
I do not see them in the CI output - https://travis-ci.com/submariner-io/submariner-operator/builds/149776758

pkg/discovery/globalnet/globalnet_test.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+
+	When("At least one Base CIDR is subset of new CIDR", func() {
+		overlapping, err := IsOverlappingCIDR([]string{"10.10.0.0/16", "10.20.30.0/24"}, "10.10.10.0/24")
+		It("Should not return error", func() {
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("Should return true", func() {
+			Expect(overlapping).To(BeTrue())
+		})
```
sridhargaddam: Looks like you missed removing this test/updating it. Its same as test case at L36

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,63 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	clusterCIDR, err = getPodCIDR(clusterCIDR, networkDetails)
 	exitOnError("Error determining the pod CIDR", err)
 
+	status.Start("* Discovering multi cluster details")
```
sridhargaddam: nit: multi-cluster

pkg/subctl/cmd/join.go
```diff
@@ -185,12 +189,63 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		if overlap && k != clusterID {
+			return fmt.Errorf("invalid service CIDR: %s overlaps with cluster %s", serviceCIDR, k)
+		}
+	}
+	return nil
+}
+
+func checkOverlappingClusterCidr(networks map[string]*globalnet.GlobalNetwork) error {
+	for k, v := range networks {
+		overlap, err := globalnet.IsOverlappingCIDR(v.ClusterCIDRs, clusterCIDR)
```
sridhargaddam: This API is very similar to checkOverlappingServiceCidr. Shall we have a single API and pass different args?

sridhargaddam:
@vthapar, also please update the commit message as we are now checking both ServiceCIDRs and ClusterCIDRs in this PR. Thanks.

pkg/discovery/globalnet/globalnet_test.go
vthapar: yeah, missed cleanup after copy paste.

pkg/subctl/cmd/join.go
vthapar: Actually we don't need different args at all. I'll refactor this as single function.

pkg/discovery/globalnet/globalnet_test.go
vthapar: https://travis-ci.com/submariner-io/submariner-operator/builds/149776758#L1220

pkg/subctl/cmd/join.go
mangelajo: Please remove the "* " at the start, this is handled by the status library.

pkg/subctl/cmd/join.go
mangelajo: It'd be cool also to find a way to check for also the worker node IPs, may be we can add annotations about that info into the Cluster objects at some point, but we don't have it now.

mangelajo:
I'm merging it for the v0.1.1 release, from the perspective that this adds extra checks, although it's related to globalnet CIDRs, that functionality is not enforced or exposed, and we can benefit from the additional checks on the UX.

Issue #180, opened 2020-02-18
subctl doesn’t pass K8s configs/contexts to kubectl and kubefedctl


skitt:
In scenarios where `subctl` needs other tools (`kubectl` and `kubefedctl`, as used for Lighthouse), it doesn’t pass parameters to specify the configuration to use, so the whole setup only works if `KUBECONFIG` etc. are set appropriately externally.

PR #183, opened 2020-02-19
Provide config and context arguments to commands

skitt:
Currently, subctl runs kubectl and kubefedctl in the default context:
whichever kubeconfig is the default, and whichever context is the
default. If the user specifies --kubeconfig and/or --kubecontext on
the subctl command-line, this causes mix-ups since subctl and the
other tools don’t touch the same cluster.

With this patch, kubectl and kubefedctl are given the appropriate
flags as necessary. This isn’t particularly elegant but it will all be
removed once KubeFed is removed.

Fixes: #180


mangelajo:
Opps, sorry, conflicts @skitt can you look at it?

PR #185, opened 2020-02-19
Add configurable CableType and DriverImpl

roytman:
In order to support pluggable VPN, we need to propagate these configuration options as an input parameters.


pkg/apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -44,6 +44,8 @@ type SubmarinerSpec struct {
 	ServiceCIDR              string `json:"serviceCIDR"`
 	ClusterCIDR              string `json:"clusterCIDR"`
 	Namespace                string `json:"namespace"`
+	CableType                string `json:"cableType"`
```
mangelajo: What do you think about using may be just CableType, which could be in the format:

cable[/driver]  ?

mangelajo: I'd also add the omitempty, which makes it an optional field, for backwards compatibility I think we need it.

pkg/controller/submariner/submariner_controller.go
```diff
@@ -277,6 +277,8 @@ func newEnginePodTemplate(cr *submarinerv1alpha1.Submariner) corev1.PodTemplateS
 						{Name: "SUBMARINER_DEBUG", Value: strconv.FormatBool(cr.Spec.Debug)},
 						{Name: "SUBMARINER_NATENABLED", Value: strconv.FormatBool(cr.Spec.NatEnabled)},
 						{Name: "SUBMARINER_BROKER", Value: cr.Spec.Broker},
+						{Name: "SUBMARINER_CABLE", Value: cr.Spec.CableType},
+						{Name: "SUBMARINER_DRIVER", Value: cr.Spec.DriverImpl},
```
mangelajo: Again we could leave it back to convention over a single env var to reduce complexity/stuff moving around.

pkg/subctl/cmd/join.go
```diff
@@ -77,6 +79,9 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
 	cmd.Flags().StringVar(&brokerClusterContext, "broker-cluster-context", "", "Broker cluster context")
+	cmd.Flags().StringVar(&cableType, "cable-type", "ipsec", "Cable type: ipsec or wireguard")
```
mangelajo: On an initial merge I would not talk of wireguard, until we finally merge it and have support, otherwise it could look like we already support wireguard, while we not. (We could accept it for development purposes and making our devel lifes easier, but not announce it :-) )

mangelajo:
Some comments, thanks!, in general it looks good though.

pkg/subctl/cmd/join.go
dfarrell07: +1

pkg/apis/submariner/v1alpha1/submariner_types.go
roytman: Done

pkg/subctl/cmd/join.go
roytman: Done

roytman:
Thanks for the comments, all required changes are done.

dfarrell07:
You may want to add checks for these new spec fields in `scripts/kind-e2e/lib_operator_verify_subm.sh`, for example in the `verify_subm_cr`, `verify_subm_crd`, `verify_subm_engine_pod`, and `verify_subm_engine_container` functions.

mangelajo:
@roytman see comments by @dfarrell07 ^ , it's about the e2e checks that we perform after deploy

roytman:
@dfarrell07 , @mangelajo thanks, I added e2e tests.

dfarrell07:
Note that this needs some manual work to resolve a conflict in `pkg/apis/submariner/v1alpha1/zz_generated.openapi.go`.

pkg/subctl/cmd/join.go
```diff
@@ -77,6 +78,8 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
 	cmd.Flags().StringVar(&brokerClusterContext, "broker-cluster-context", "", "Broker cluster context")
+	cmd.Flags().StringVar(&cableDriver, "cable-driver", "ipsec/strongswan", "cable[/driver] - Cable type and optional its driver implementation")
```
```suggestion
	cmd.Flags().StringVar(&cableDriver, "cable-driver", "ipsec/strongswan", "cable[/driver] - Cable type and optional driver implementation")
```

roytman:
> Note that this needs some manual work to resolve a conflict in `pkg/apis/submariner/v1alpha1/zz_generated.openapi.go`.

Oops, I missed it, will fix

pkg/subctl/cmd/join.go
roytman: Done

pkg/subctl/cmd/join.go
```diff
@@ -80,6 +81,8 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
 	cmd.Flags().StringVar(&brokerClusterContext, "broker-cluster-context", "", "Broker cluster context")
+	cmd.Flags().StringVar(&cableDriver, "cable-driver", "ipsec/strongswan", "cable[/driver] - Cable type and optional driver implementation")
```
tpantelis: I think we should specify "" default value and let the engine backend decide what default driver to use. That way if we change the default driver in the backend in a future release, we don't have to change it here and upgrades will automatically pick up the new default.

roytman: Done

scripts/kind-e2e/e2e.sh
```diff
@@ -130,6 +130,7 @@ function create_subm_vars() {
   subm_colorcodes=blue
   subm_debug=false
   subm_broker=k8s
+  subm_cabledriver=ipsec/strongswan
```
mangelajo: I think we need to change this to strongswan in the end.
```suggestion
  subm_cabledriver=strongswan
```

pkg/subctl/cmd/join.go
```diff
@@ -85,6 +86,7 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&submarinerDebug, "subm-debug", false, "Enable Submariner debugging (verbose logging)")
 	cmd.Flags().BoolVar(&noLabel, "no-label", false, "skip gateway labeling")
 	cmd.Flags().StringVar(&brokerClusterContext, "broker-cluster-context", "", "Broker cluster context")
+	cmd.Flags().StringVar(&cableDriver, "cable-driver", "", "Cable driver implementation")
```
mangelajo:
```suggestion
	cmd.Flags().StringVar(&cableDriver, "cable-driver", "strongswan", "Cable driver implementation")
```

mangelajo: We need a default

tpantelis: The default is decided in the backend when empty string is passed. As per my previous comment:

> I think we should specify "" default value and let the engine backend decide what default driver to use. That way if we change the default driver in the backend in a future release, we don't have to change it here and upgrades will automatically pick up the new default.

scripts/kind-e2e/e2e.sh
tpantelis: Or leave it blank.

roytman: I think the `strongswan` string is better, because we can check that user can define the driver name and the name will be propagated into the `submariner` CRD, otherwise the relevant CRD filed will not exist.
What do you think ?

tpantelis: Good point - agree.

pkg/subctl/cmd/join.go
mangelajo: Ok, makes sense. Migration from one driver to another will need to be checked. We will also need to warn the customer of the change of driver when that's about to happen.

PR #204, opened 2020-02-25
Rename ClusterIP CIDRs to Service CIDRs

sridhargaddam:
Currently when we query the cluster info using "subctl info ..." command
the output displays ClusterIP Cidrs and Pod CIDRs. In K8s, POD CIDR is
usually referred to as Cluster CIDR. So, using ClusterIP CIDR for Service
CIDR is really confusing. This patch modifies the output accordingly.


dfarrell07:
> makes sense, @dfarrell07 be aware of this on your side for the discovery part you needed in helm.

ACK, okay, I'll also keep this in mind - there might be other places we can clarify these names in the scripts.

PR #224, opened 2020-03-03
e2e: Add globalnet to e2e

vthapar:


.travis.yml
```diff
@@ -13,7 +13,7 @@ git:
 services:
   - docker
 script:
-  - make ci e2e status=keep
+  - make ci e2e status=keep globalnet=true
```
vthapar: E2e with Globalnet build: https://travis-ci.com/submariner-io/submariner-operator/builds/151560337

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,96 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+			IP:   uintToIP(last + 1),
+			Mask: mask,
+		}
+		cidr = nextNet.String()
+		last, err = allocateByCidr(cidr)
+		if err != nil && last == 0 {
+			return "", fmt.Errorf("allocation not available")
+		}
+	}
+	return cidr, nil
```
skitt: This would be easier to understand if it was in a single loop.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,96 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+}
+
+func AllocateGlobalCIDR(networks map[string]*GlobalNetwork, subctlData *datafile.SubctlData) (string, error) {
+	globalCidr = GlobalCIDR{allocatedCount: 0, cidr: subctlData.GlobalnetCidrRange}
+	_, network, _ := net.ParseCIDR(globalCidr.cidr)
```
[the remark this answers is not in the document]
vthapar: This validation was already done when we populated `broker-info.subm` during `deploy-broker`

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,96 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+	globalCidr.net = network
+	for _, network := range networks {
```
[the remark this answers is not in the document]
vthapar: Will rename it.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,96 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+			return clusterCidr.lastIp, fmt.Errorf("%s overlaps with allocated %s\n", cidr, allocated.network)
+		}
+	}
+	globalCidr.allocatedClusters = append(globalCidr.allocatedClusters, &clusterCidr)
+	globalCidr.allocatedCount++
+	return 0, nil
+}
+
+func allocateByClusterSize(numSize uint) (string, error) {
+	bitSize := int(math.Log2(float64(numSize)))
```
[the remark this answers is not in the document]
vthapar: Thanks, will definitely use this :)

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,96 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+	ones, total := network.Mask.Size()
+	clusterSize := uint(total - ones)
+	firstIpInt := ipToUint(network.IP)
+	lastIpUint := (firstIpInt + 1<<clusterSize) - 1
+	return lastIpUint
+}
+
+func allocateByCidr(cidr string) (uint, error) {
+	requestedIp, requestedNetwork, err := net.ParseCIDR(cidr)
+	if err != nil || !globalCidr.net.Contains(requestedIp) {
```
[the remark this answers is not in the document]
vthapar: That gets checked in subsequent steps.

pkg/subctl/cmd/join.go
```diff
@@ -242,13 +243,23 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		status.End(err == nil)
+		exitOnError("Error validating overlapping ClusterCIDRs", err)
+	} else if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
+		globalCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
+		status.End(err == nil)
+		status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalCIDR))
+		exitOnError("Globalnet failed", err)
+	} else {
+		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
+		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: ", globalNetworks[clusterID].GlobalCIDRs[0]))
```
sridhargaddam: %s missing in the printf message is causing build failures

vthapar: Done.

pkg/subctl/cmd/join.go
```diff
@@ -242,13 +243,23 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	if subctlData.GlobalnetCidrRange == "" {
+		// Globalnet not enabled
+		err = checkOverlappingServiceCidr(globalNetworks)
+		status.End(err == nil)
+		exitOnError("Error validating overlapping ServiceCIDRs", err)
+		err = checkOverlappingClusterCidr(globalNetworks)
+		status.End(err == nil)
+		exitOnError("Error validating overlapping ClusterCIDRs", err)
+	} else if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
+		globalCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
```
sridhargaddam: Can you add a note/comment on when this case will hit?

pkg/controller/submariner/submariner_controller.go
```diff
@@ -392,6 +402,75 @@ func newRouteAgentDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet
...
+	allowPrivilegeEscalation := true
+	privileged := true
+	readOnlyFileSystem := false
+	runAsNonRoot := false
+	security_context_all_cap_allow_escal := corev1.SecurityContext{
+		Capabilities:             &corev1.Capabilities{Add: []corev1.Capability{"ALL"}},
+		AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+		Privileged:               &privileged,
+		ReadOnlyRootFilesystem:   &readOnlyFileSystem,
+		RunAsNonRoot:             &runAsNonRoot,
```
sridhargaddam: In the future, we can identify the minimum caps and try to give only those permission. This is ok for now as we are doing similar things even for other submariner pods.

.travis.yml
sridhargaddam: We should remove this check (i.e., globalnet=true) before merging this PR right?

scripts/kind-e2e/e2e.sh
```diff
@@ -145,8 +153,34 @@ function create_subm_vars() {
   subm_broker_ns=submariner-k8s-broker
 }
 
+function get_globalip() {
+    svcname=$1
+    gip=
+    attempt_counter=0
+    max_attempts=5
```
sridhargaddam: @tpantelis gave a comment about 5 seconds on my PR, which I think is valid. Let's increase the max_attempts to 30 secs.

sridhargaddam: you can also take a look at the other comments and make necessary changes - https://github.com/submariner-io/submariner/pull/363/files

scripts/kind-e2e/e2e.sh
```diff
@@ -328,10 +367,13 @@ deploy_nginx_cluster3
...
-../bin/subctl verify-connectivity ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster2 \
-                                  ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster3 \
-                                  --verbose
+# dataplane E2E need to be modified for globalnet
+if [[ $globalnet = false ]]; then
+    # run dataplane E2e tests between the two clusters
+    ../bin/subctl verify-connectivity ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster2 \
+                                      ${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster3 \
+                                      --verbose
+fi
```
sridhargaddam: this check is not required once my PR merges (but do we need an official SM release before we can consume it here?) - https://github.com/submariner-io/submariner/pull/363/files

.travis.yml
vthapar: > We should remove this check (i.e., globalnet=true) before merging this PR right?

Yes.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,102 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+	}
+	globalCidr.net = network
+	for _, globalNetwork := range globalNetworks {
+		for _, otherCluster := range globalNetwork.GlobalCIDRs {
+			otherClusterCIDR := NewCIDR(otherCluster)
+			globalCidr.allocatedClusters = append(globalCidr.allocatedClusters, &otherClusterCIDR)
+			globalCidr.allocatedCount++
+		}
+	}
+	return allocateByClusterSize(subctlData.GlobalnetClusterSize)
```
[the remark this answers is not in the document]
vthapar: We check it against the CIDR given in globalCidrRange. So, for /24 it can't go more than 255.
https://github.com/submariner-io/submariner-operator/blob/master/pkg/subctl/cmd/deploybroker.go#L143

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,102 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+	}
+	globalCidr.allocatedClusters = append(globalCidr.allocatedClusters, &clusterCidr)
+	globalCidr.allocatedCount++
+	return 0, nil
+}
+
+func allocateByClusterSize(numSize uint) (string, error) {
+	bitSize := bits.LeadingZeros(0) - bits.LeadingZeros(numSize-1)
+	_, totalbits := globalCidr.net.Mask.Size()
+	clusterPrefix := totalbits - bitSize
```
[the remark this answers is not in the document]
vthapar: Yes, will do, makes it simpler.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,102 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+	}
+	clusterCidr := NewCIDR(cidr)
+	if !globalCidr.net.Contains(uintToIP(clusterCidr.lastIp)) {
+		return 0, fmt.Errorf("%s not a valid subnet of %v\n", cidr, globalCidr.net)
+	}
+	for i := 0; i < globalCidr.allocatedCount; i++ {
+		allocated := globalCidr.allocatedClusters[i]
+		if allocated.network.Contains(requestedIp) {
+			//subset of already allocated, try next
+			return allocated.lastIp, fmt.Errorf("%s subset of already allocated %v\n", cidr, allocated.network)
```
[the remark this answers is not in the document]
vthapar: done. Used globalCidr instead of network.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -83,3 +101,102 @@ func IsOverlappingCIDR(cidrList []string, cidr string) (bool, error) {
...
+		}
+	}
+	return cidr, nil
+}
+
+func AllocateGlobalCIDR(globalNetworks map[string]*GlobalNetwork, subctlData *datafile.SubctlData) (string, error) {
+	globalCidr = GlobalCIDR{allocatedCount: 0, cidr: subctlData.GlobalnetCidrRange}
+	_, network, err := net.ParseCIDR(globalCidr.cidr)
+	if err != nil {
+		return "", fmt.Errorf("invalid GlobalCIDR %s in broker-info.subm", globalCidr.cidr)
```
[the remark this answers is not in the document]
vthapar: It might add a cyclic dependency so wanted to avoid. Will just remove file name and use generic message.

vthapar:
CI with globlanet enabled: https://travis-ci.com/submariner-io/submariner-operator/builds/152585060

With all comments addressed, will disable globalnet=enable before merge.

scripts/kind-e2e/e2e.sh
vthapar: yes.

pkg/discovery/globalnet/globalnet.go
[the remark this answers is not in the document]
sridhargaddam: Right, but the error message could be really confusing to the user. 
Say user has specified clusterSize as 33000 for a globalCidrRange of "/16", the error message which will be displayed will be
"Cluster size 65536 should be <= 32768"

They could be wondering from where this 65536 came into the picture. You can propose a separate PR with proper message, just sharing it for your info.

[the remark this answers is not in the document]
vthapar: That will need to be done on deploy broker code, which is out of scope of this patch. Can you raise an issue, I'll do a PR to address this.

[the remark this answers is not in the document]
sridhargaddam: https://github.com/submariner-io/submariner-operator/issues/258

deploy/cluster_role.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
mangelajo: nit/note for follow up: we should rename this to globalnet_cluster_role so the filename it's clearer

deploy/cluster_role_binding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
mangelajo: nit: same

pkg/controller/submariner/submariner_controller.go
```diff
@@ -393,6 +403,75 @@ func newRouteAgentDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet
 	return routeAgentDaemonSet
 }
 
+func newGlobalnetDaemonSet(cr *submarinerv1alpha1.Submariner) *appsv1.DaemonSet {
```
mangelajo: follow-up: Please move the specific globalnet logic to a separate .go file

deploy/cluster_role.yaml
vthapar: This was also discussed with Stephen. Currently we're using the same SA as submariner engine and routeagent. So kept the name generic in case need to add more permissions to cluster roles. Common name also conveys message that this cluster role applies to engine and route-agent too, as using same SA. Using globalnet_cluster_role without a different SA can give misleading impression that this applies only to globalnet.

IMO, this should be renamed/refactored when we move to using individual and specific SAs for engine, routeagent and globalnet.

mangelajo: Makes sense, thanks for the explanation, agreed

pkg/controller/submariner/submariner_controller.go
mangelajo: True

PR #227, opened 2020-03-04
Don’t set DNS up on broker

skitt:
The broker doesn’t need DNS overrides (and we don’t want to replace
its CoreDNS and break its CVO, or ask the user about it).

(cherry picked from commit f527a314985d5a13fc53951775f4be646091e98d)


mangelajo:
Based on PR#218 votes I will merge this one.

Issue #235, opened 2020-03-05
subctl flag to accept "Enabling service discovery on OpenShift"


manosnoam:
Please add a flag to auto-accept "Enabling service discovery on OpenShift will disable OpenShift updates, do you want to continue?" when running subctl deploy with --service-discovery.

PR #239, opened 2020-03-05
Add flag to disable the OpenShift CVO

mangelajo:
When subctl determines that the OpenShift CVO needs to be disabled, it
asks for confirmation; this patch allows skipping this check (and
forcing the CVO to be disabled, if necessary) with --disable-cvo.

Fixes: #235
(cherry picked from commit 52a6b94d9436f6264b09ffd928bc224fe77c77d2)


mangelajo:
Thank you guys for the quick reviews.

Issue #236, opened 2020-03-05
subctl: Deploying multi cluster service discovery can fail "dns-operator" failed update


mangelajo:
The dns-operator object needs to be updated with a reload/retry mechanism as it happens for many other k8s objects.

```
[majopela@bluehat subm-demo]$ subctl deploy-broker --kubecontext cluster-a --kubeconfig cluster-a/auth/kubeconfig --dataplane --service-discovery --broker-cluster-context cluster-a --clusterid my-cluster-a
 ✓ Deploying broker 
 ✓ Creating broker-info.subm file
 ✓ A new IPSEC PSK will be generated for broker-info.subm
 ✓ Deploying Service Discovery controller 
 ✓ Created operator CRDs
 ✓ Created operator namespace: kubefed-operator
 ✓ Created operator service account
 ✓ Deployed the operator successfully
 ✓ Created lighthouse CRDs
 ✓ Created lighthouse controller
? Enabling service discovery on OpenShift will disable OpenShift updates, do you want to continue? Yes
? Which node should be used as the gateway? ip-10-0-137-75.eu-west-3.compute.internal
 ✓ Deploying the Submariner operator 
 ✓ Created operator CRDs
 ✓ Created operator namespace: submariner-operator
 ✓ Created operator service account and role
 ✓ Updated the privileged SCC
 ✓ Deployed the operator successfully
 ✗ Deploying multi cluster service discovery 
 ✓ Added Lighthouse entry in the openshift-dns role
 ✓ Disabled the cluster version operator
 ✓ Added Lighthouse entry in the openshift-dns-operator role
 ✓ Updated DNS operator deployment
Error deploying multi cluster service discovery: error setting DNS up: Operation cannot be fulfilled on deployments.apps "dns-operator": the object has been modified; please apply your changes to the latest version and try again
subctl version: dev
built from git commit id: 9f04472d9cd9642f644af75a3401c3fea6f29da3
```

***workaround**
A second run of subctl will make it work:
```
subctl deploy-broker --kubecontext cluster-a --kubeconfig cluster-a/auth/kubeconfig --dataplane --service-discovery --broker-cluster-context cluster-a --clusterid my-cluster-a
 ✓ Deploying broker 
 ✓ Creating broker-info.subm file
 ✓ Reusing IPSEC PSK from existing broker-info.subm
 ✓ Backed up previous broker-info.subm to broker-info.subm.2020-03-05T14_50_06+01_00
[...]

PR #242, opened 2020-03-05
Retry dns-operator updates on conflicts

skitt:
This uses the standard K8s retry pattern. It’s not the only place in
subctl where we need to do this, I’ll follow up with another PR to fix
the rest.

Fixes: #236

### DIFF
--- a/pkg/subctl/lighthouse/dns/ensure.go
+++ b/pkg/subctl/lighthouse/dns/ensure.go
@@ -28,6 +28,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	clientsetifc "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 )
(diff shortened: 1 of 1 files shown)


pkg/subctl/lighthouse/dns/ensure.go
```diff
@@ -227,28 +228,31 @@ func setupOpenShift(status *cli.Status, clientSet *clientset.Clientset, repo str
...
-		if container.Name == "dns-operator" {
-			deployment.Spec.Template.Spec.Containers[i].Image = repo + operatorImage + ":" + version
-			for j, env := range container.Env {
-				if env.Name == "IMAGE" {
-					deployment.Spec.Template.Spec.Containers[i].Env[j].Value = repo + openShiftCoreDNSImage + ":" + version
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		deployment, err := deployments.Get("dns-operator", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
```
tpantelis: This should/will return `errors.IsNotFound` if it doesn't exist so I think we should check for that and return nil.  Also there's no need for the nil check on L236 - it won't return nil if no error is returned (in fact it won't return nil either way).

On a side note, there's also no need for the nil check on L227 above as that call never returns nil.

pkg/subctl/lighthouse/dns/ensure.go
```diff
@@ -227,28 +228,31 @@ func setupOpenShift(status *cli.Status, clientSet *clientset.Clientset, repo str
...
+		if deployment == nil {
+			// Assume we're not on OpenShift
+			return nil
+		}
```
mangelajo:
```suggestion
		if errors.IsNotFound(err) {
			// Assume we're not on OpenShift
			return nil
		} else if err != nil {
			return err
		}
```

pkg/subctl/lighthouse/dns/ensure.go
skitt: Done, thanks!

Issue #244, opened 2020-03-06
subctl sometimes fails at join with: x509: certificate signed by unknown authority


mangelajo:
Discovering multi cluster details Error trying to discover multi-cluster network details: Get https://api.majopela-mar6-a.devcluster.openshift.com:6443/apis/submariner.io/v1/clusters: x509: certificate signed by unknown authority

If we look at the broker-info.subm file, we find the wrong token: "submariner-k8s-broker-client-dockercfg-v6jlj"

which are the docker pull secrets.

If we look at the service account we see:
```
- apiVersion: v1
  imagePullSecrets:
  - name: submariner-k8s-broker-client-dockercfg-v6jlj
  kind: ServiceAccount
  metadata:
    creationTimestamp: "2020-03-06T10:23:19Z"
    name: submariner-k8s-broker-client
    namespace: submariner-k8s-broker
    resourceVersion: "19684"
    selfLink: /api/v1/namespaces/submariner-k8s-broker/serviceaccounts/submariner-k8s-broker-client
    uid: 8002d982-5f94-11ea-8766-0eb83f56231e
  secrets:
  - name: submariner-k8s-broker-client-dockercfg-v6jlj
  - name: submariner-k8s-broker-client-token-k976b

```

But our code assumes that secrets[0] will be the "-token" 

https://github.com/submariner-io/submariner-operator/blob/master/pkg/broker/rbac.go#L90

Such code must be flexible

PR #246, opened 2020-03-06
Fix issue with wrong token stored in subm file (OpenShift)

mangelajo:
Sometimes the service account inverts the order of the secrets,
and our code is not smart enough, this patch fixes it looking
for the right secret.

Fixes: #244
(cherry picked from commit aa8f09fb5a3189e136044c2a9a8311dfb5621395)

### DIFF
--- a/pkg/broker/rbac.go
+++ b/pkg/broker/rbac.go
@@ -18,6 +18,7 @@ package broker
 
 import (
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/versions/versions.go
```diff
@@ -2,8 +2,8 @@ package versions
 
 const (
 	DefaultSubmarinerRepo            = "quay.io/submariner"
-	DefaultSubmarinerOperatorVersion = "0.1.0"
-	DefaultSubmarinerVersion         = "0.1.0"
-	DefaultLighthouseVersion         = "0.1.0"
+	DefaultSubmarinerOperatorVersion = "0.1.1"
+	DefaultSubmarinerVersion         = "0.1.1"
+	DefaultLighthouseVersion         = "0.1.0'"
```
mangelajo: this creeped in... fixing

PR #248, opened 2020-03-06
WIP: Convert e2e to use shared scripts from image

dfarrell07:
Instead of using the e2e.sh and subsequent scripts in this repo, use the
shared /opt/kind-e2e logic now packaged in the dapper-base image built
in submariner-io/submariner.

Allows de-duplication of deployment scripting, preventing fragmentation
and easing maintains.

Any special logic added to the submariner-operator e2e scripting after
it was copied from the main submariner repo will need to be ported back
to the main submariner repo.

Relates to: #369

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

PR #261, opened 2020-03-13
Adjust kind clusters for 0.6.1

skitt:
This is a port of
https://github.com/submariner-io/submariner/commit/c2fc6b5e6db363d9ab555320827bf044b5c690e3
to submariner-operator, taking into account the new cluster naming
with recent versions of kind.

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -5,6 +5,21 @@ source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 
 ### Functions ###
 
+function kind_fixup_config() {
+    cluster=$1
+    master_ip=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' ${cluster}-control-plane | head -n 1)
+    sed -i -- "s/server: .*/server: https:\/\/$master_ip:6443/g" $KUBECONFIG
+    sed -i -- "s/user: kind-.*/user: ${cluster}/g" $KUBECONFIG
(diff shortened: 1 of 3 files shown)


mangelajo:
> LGTM, but the deployment of lighthouse is failing :/

Yeah, I suspect may be lighthouse timeout is too low in travis? ...

scripts/kind-e2e/e2e.sh
```diff
@@ -5,6 +5,21 @@ source $(git rev-parse --show-toplevel)/scripts/lib/debug_functions
 
 ### Functions ###
 
+function kind_fixup_config() {
```
dfarrell07: Maybe we can make these changes in submariner-io/submariner e2e.sh, then consume the fixes via https://github.com/submariner-io/submariner-operator/pull/248? I think the goal is to get rid of these scripts in submariner-operator ASAP. We could also try to get in https://github.com/submariner-io/submariner/pull/425 in first to take advantage of that refactoring.

dfarrell07:
```
[scripts]$ kubectl get submariner submariner --namespace=submariner-operator -o jsonpath={.spec.serviceCIDR}
[scripts]$ grep 100.95.0.0/16
[submariner-operator]$ ./scripts/e2e keep true true
FATA[0996] exit status 1
```

scripts/kind-e2e/e2e.sh
dfarrell07: Or second-thinking, maybe it's better just to fix it here for now and do this and other migrations when we move to shipyard.

mkolesnik:
Closing since we moved to shipyard

Issue #211, opened 2020-02-26
E2E deployment for submariner-operator repo using shipyard


mangelajo:

mkolesnik:
This has been done

PR #272, opened 2020-03-18
Use latest version of dapper-base

dfarrell07:
Un-pin the version of dapper-base so we can pull in the latest changes,
in particular the new shared e2e scripting.

Relates to: #211


dfarrell07:
> See #261 which already does this (and tries to address the resulting problems).

Oh wonderful, thank you! I knew I saw some other relevant work to this float by but couldn't find it.

Issue #273, opened 2020-03-18
Rename "globalnet-enable" flag to "global-net"


manosnoam:
Just like we have the "service-discovery" flag (not "servicediscovery-enable"),
can we change the "globalnet-enable" flag to "global-net" ?

PR #274, opened 2020-03-18
Rename “globalnet-enable” to “globalnet”

skitt:
Fixes: #273

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -80,7 +80,7 @@ function setup_broker() {
          sd=
          [[ $lighthouse = true ]] && sd=--service-discovery
          gn=
-         [[ $globalnet = true ]] && gn=--globalnet-enable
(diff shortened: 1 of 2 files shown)


scripts/kind-e2e/e2e.sh
```diff
@@ -80,7 +80,7 @@ function setup_broker() {
          sd=
          [[ $lighthouse = true ]] && sd=--service-discovery
          gn=
-         [[ $globalnet = true ]] && gn=--globalnet-enable
+         [[ $globalnet = true ]] && gn=--globalnet
```
manosnoam: I thought to be consistent with "service-discovery" and call it "global-net" (with a dash). What do you think ?

vthapar: `Globalnet` is a the feature name unlike `Service Discovery`. We use Globalnet everywhere, not Global Net.

PR #276, opened 2020-03-18
Update operator versions for 0.2.0-rc0 release

mangelajo:


mangelajo:
globalip + LH seems to fail, but releasing rc0 , we can release rc1 later... and start testing manually and checking what's wrong.

Issue #255, opened 2020-03-10
subctl needs to retry on conflicts


skitt:
General K8s object updates need to handle conflicts (as mentioned in #242).

PR #278, opened 2020-03-20
Ensure that update operations are retried

skitt:
All Kubernetes update operations should be retried on conflicts,
i.e. the object being changed has changed on the server since it was
retrieved.

Fixes: #255

### DIFF
--- a/pkg/utils/createorupdate.go
+++ b/pkg/utils/createorupdate.go
@@ -1,0 +1,162 @@
+/*
+© 2020 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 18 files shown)


skitt:
TODO:

[ ] Add a full set of tests
[ ] Handle the Submariner CR and SCC

pkg/utils/createorupdate.go
```diff
@@ -0,0 +1,162 @@
...
+			clusterRole.ResourceVersion = existingClusterRole.ResourceVersion
+			// Potentially retried
+			_, err = clientSet.RbacV1().ClusterRoles().Update(clusterRole)
+			if err != nil {
+				return fmt.Errorf("failed to update pre-existing cluster role %s : %v", clusterRole.Name, err)
+			}
+			return nil
+		})
+		return false, retryErr
+	}
```
tpantelis: I think we should use ` controllerutil.CreateOrUpdate` as is done [here](https://github.com/submariner-io/submariner-operator/blob/master/pkg/controller/submariner/submariner_controller.go#L178) but perhaps that should be in another PR.

skitt: I tried to do that, but `controllerutil` works in the context of a controller (with a plan client and context), it doesn’t work with a clientSet. I imagine it’s possible to get a client and context from a clientSet, but I haven’t figured out how (yet).

tpantelis: right - forgot about that. We can look into that later. It's a little harder to do it generically with client sets.

PR #290, opened 2020-03-31
Migrate cluster deployment to shipyard

mkolesnik:
Cluster deployment is available in shipyard so use that instead of
reimplementing it here


scripts/kind-e2e/e2e.sh
```diff
@@ -1,76 +1,26 @@
...
-                if [[ $? -ne 0 && $? -ne 127 ]]; then
-                    echo Cluster $i creation failed:
-                    cat ${logs[$i]}
-                fi
-                rm -f ${logs[$i]}
-            fi
-        done
-    fi
-    KUBECONFIG=${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster1:${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster2:${PRJ_ROOT}/output/kind-config/dapper/kind-config-cluster3 \
-    kubectl config view --flatten > ${PRJ_ROOT}/output/kind-config/dapper/kind-config-merged
```
skitt: I suspect we still need this somewhere...

mkolesnik: I kept it, it's in the main section

skitt: Indeed, I’m blind and I don’t know how to use Ctrl+F!

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -79,45 +79,35 @@ function verify_subm_cr() {
...
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.brokerK8sApiServerToken}' | grep $SUBMARINER_BROKER_TOKEN
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.brokerK8sCA}' | grep $SUBMARINER_BROKER_CA
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.brokerK8sRemoteNamespace}' | grep $SUBMARINER_BROKER_NS
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.ceIPSecDebug}' | grep $ce_ipsec_debug
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.ceIPSecIKEPort}' | grep $ce_ipsec_ikeport
-  kubectl get submariner $deployment_name --namespace=$subm_ns -o jsonpath='{.spec.ceIPSecNATTPort}' | grep $ce_ipsec_nattport
+  json_file=/tmp/${deployment_name}.json
+  kubectl get submariner $deployment_name --namespace=$subm_ns -o json | tee $json_file
+
+  validate_equals '.metadata.namespace}' $subm_ns
```
skitt: Shouldn’t these drop the closing bracket?

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -125,44 +125,59 @@ function verify_subm_op_pod() {
...
+function validate_pod_container_has() {
+  local json_filter=".spec.containers[].${1}"
+  local expected=$2
+  [[ $(jq -r -M $json_filter $json_file) =~ $expected ]]
+}
+
+function validate_pod_container_env() {
+  local var_name=$1
+  local expected=$2
+  [[ $(jq -r -M ".spec.containers[].env[] | select(.name==\"${var_name}\").value" $json_file) = $expected ]]
```
skitt: In `kubectl jsonpath` this can be written `.spec.containers.env[?(@.name="${var_name}")].value`, not sure whether that would work here.

scripts/kind-e2e/lib_operator_verify_subm.sh
mkolesnik: right, keen eye! some bad vim-ing on my part :)

scripts/kind-e2e/lib_operator_verify_subm.sh
mkolesnik: Doesn't seem like they use the same underlying library

```
[root@04239f4bedaa submariner-operator]# jq '.spec.containers[].env[?(@.name="SUBMARINER_NAMESPACE")].value' //tmp/pod.json
jq: error: syntax error, unexpected '?' (Unix shell quoting issues?) at <top-level>, line 1:
.spec.containers[].env[?(@.name="SUBMARINER_NAMESPACE")].value                       
```

Makefile
```diff
@@ -11,8 +13,17 @@ TARGETS := $(shell ls scripts)
 	@./.dapper.tmp -v
 	@mv .dapper.tmp .dapper
 
+cleanup: .dapper
+	./.dapper -m bind $(SCRIPTS_DIR)/cleanup.sh
+
+clusters: build
```
mangelajo, line 19: don't we need package after build too?

mangelajo: oh as you clarified we don't have a separate action in the submariner-operator repo

PR #292, opened 2020-04-01
Add basic operator status

skitt:
This adds a standard status object returning parameters which the user
might want to check.

### DIFF
--- a/pkg/apis/submariner/v1alpha1/submariner_types.go
+++ b/pkg/apis/submariner/v1alpha1/submariner_types.go
@@ -54,6 +54,13 @@ type SubmarinerSpec struct {
 // SubmarinerStatus defines the observed state of Submariner
 // +k8s:openapi-gen=true
 type SubmarinerStatus struct {
+	NatEnabled  bool   `json:"natEnabled"`
+	ColorCodes  string `json:"colorCodes,omitempty"`
+	ClusterID   string `json:"clusterID"`
+	ServiceCIDR string `json:"serviceCIDR"`
+	ClusterCIDR string `json:"clusterCIDR"`
+	GlobalCIDR  string `json:"globalCIDR,omitempty"`
+	CableDriver string `json:"cableDriver,omitempty"`
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
--- a/deploy/crds/submariner.io_submariners_crd.yaml
+++ b/deploy/crds/submariner.io_submariners_crd.yaml
@@ -88,6 +88,26 @@ spec:
           type: object
         status:
           description: SubmarinerStatus defines the observed state of Submariner
+          properties:
+            cableDriver:
+              type: string
+            clusterCIDR:
+              type: string
+            clusterID:
+              type: string
+            colorCodes:
+              type: string
+            globalCIDR:
+              type: string
+            natEnabled:
+              type: boolean
+            serviceCIDR:
+              type: string
+          required:
+          - clusterCIDR
+          - clusterID
+          - natEnabled
+          - serviceCIDR
           type: object
       type: object
   version: v1alpha1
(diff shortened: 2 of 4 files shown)


pkg/apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -54,6 +54,13 @@ type SubmarinerSpec struct {
 // SubmarinerStatus defines the observed state of Submariner
 // +k8s:openapi-gen=true
 type SubmarinerStatus struct {
+	NatEnabled  bool   `json:"natEnabled"`
```
mangelajo, line 57: I think it's better not to replicate the same parameters we pass to the Submariner CRD as status

I see that's what the Network operator does, but I don't understand the reasoning. I wouldn't follow it blindly.



I think we need to assume that what we provided as configuration is what the operator has applied, we could export some details about the success of any change,

May be a list of conditions: https://docs.openshift.com/container-platform/4.2/operators/olm-status.html

is more helpful, and also eventually a list of connections (And the status of each). It's the patch I'm working on for submariner gateway.

mangelajo: Other useful things we could report:
0) List of conditions (last N? )

beyond of the scope of this initial patch:
1) The list of gateway node(s) (we could get this from submariner-engine status api endpoint WIP)
2) the active gateway node(s) (ditto)
3) The number of route agents (indication if any is unhealthy)
3) The list of cables, including status and remote ip/cidrs (ditto)

skitt: I thought the pattern was still relevant for us, at least as a start, so that users can confirm that their settings are applied. (We know they’re applied blindly, users don’t, and some of them might be changeable in future.)

I’m working on another patch which adds daemonset status, to see how that gets represented in OCP.

mangelajo: > I thought the pattern was still relevant for us, at least as a start, so that users can confirm that their settings are applied. (We know they’re applied blindly, users don’t, and some of them might be changeable in future.)
> 
> I’m working on another patch which adds daemonset status, to see how that gets represented in OCP.

I see a possible value if the operator was not running for example, since our changes will be almost instantaneous.

But IMO is marginal, I still think that a list of conditions would be better, and easier to read, we could reflect there that last changes were processed and other objects updated on the reconcile function. I admit this is subjective, let's wait for other to chime in, may be we can look for more examples of other operators in the wild?

vthapar: IMO it is good to have these fields in status, at least for ones that may or may not be applied. e.g. What if the NAT flag specified is not supported by cloud provider or OnPrem? Having field in status is confirmation user passed valid parameters and they got applied correctly.

mangelajo: ceph: https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go#L148

mangelajo: > IMO it is good to have these fields in status, at least for ones that may or may not be applied. e.g. What if the NAT flag specified is not supported by cloud provider or OnPrem? Having field in status is confirmation user passed valid parameters and they got applied correctly.

But the nat flag is only internal of how submariner discovers the IP, does not depend on the cloud provider.

BUT, you brought a good point, for example, if we leave the CIDRs for auto-discover, we could report the auto-discovered parameters  in the status.

@skitt I start to buy it.. :) , Do you think it'd be reasonable to move all those "config based" status fields under a whole structure inside status?, so it'd be easier to read/navigate.

skitt: If we take the NETCONF nomenclature, would it make sense to encapsulate this in an `Operational` structure? That’s what we’re trying to provide: the operational view of the requested configuration (available in the spec)...

skitt: I still need to check what the resulting status looks like in OpenShift, I’m not sure how it presents sub-structures.

skitt:
With this and the follow-up daemonset status patch, the status looks like

            "status": {
                "clusterCIDR": "10.243.0.0/16",
                "clusterID": "cluster3",
                "colorCodes": "blue",
                "engineDaemonSetStatus": {
                    "currentNumberScheduled": 1,
                    "desiredNumberScheduled": 1,
                    "numberAvailable": 1,
                    "numberMisscheduled": 0,
                    "numberReady": 1,
                    "observedGeneration": 1,
                    "updatedNumberScheduled": 1
                },
                "natEnabled": false,
                "routeAgentDaemonSetStatus": {
                    "currentNumberScheduled": 2,
                    "desiredNumberScheduled": 2,
                    "numberAvailable": 2,
                    "numberMisscheduled": 0,
                    "numberReady": 2,
                    "observedGeneration": 1,
                    "updatedNumberScheduled": 2
                },
                "serviceCIDR": "100.93.0.0/16"
            }

PR #295, opened 2020-04-02
[WIP] E2E: Deploy with less nodes

mkolesnik:
On the broker cluster we don't need any nodes, and on the other clusters
one node should be enough for each.


mkolesnik:
This was incorporated in another change, closing

PR #303, opened 2020-04-09
Add managed resources to the operator status

skitt:
This adds the various daemon sets to the operator status: the engine
daemonset, the route agent daemonset, and the globalnet daemonset.

Currently the status of these daemon sets isn’t updated outside the
reconciliation loop, which means the status is inaccurate.


deploy/crds/submariner.io_submariners_crd.yaml
```diff
@@ -97,10 +97,14182 @@ spec:
               type: string
             colorCodes:
               type: string
+            engineDaemonSet:
```
dfarrell07, line 100: We really need this 14k LoC?

skitt: Indeed, we don’t, I was thinking of only including the daemonset status instead.

skitt:
In addition to Daniel’s comment, this needs changes to register the daemonsets, otherwise their status is never updated.

deploy/crds/submariner.io_submariners_crd.yaml
tpantelis: This is generated code - the result of completely embedding a Pod object.  I was a little wary of this too  Perhaps we should embed just the Pod's Status?

tpantelis: Sorry my comment crossed with @skitt comment 😄

PR #306, opened 2020-04-14
Add Gateway CRD

skitt:
This will be used to store and report the cluster-local engine
status.

### DIFF
--- a/pkg/subctl/operator/submarinerop/crds/ensure.go
+++ b/pkg/subctl/operator/submarinerop/crds/ensure.go
@@ -37,7 +37,7 @@ func Ensure(restConfig *rest.Config) (bool, error) {
 		return false, err
 	}
 
-	crd, err := getSubmarinerCRD()
+	submarinerCrd, err := getSubmarinerCRD()
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/broker/rbac.go
```diff
@@ -54,7 +54,7 @@ func NewBrokerAdminRole() *rbacv1.Role {
 			rbacv1.PolicyRule{
 				Verbs:     []string{"create", "get", "list", "watch", "patch", "update", "delete"},
 				APIGroups: []string{"submariner.io"},
-				Resources: []string{"clusters", "endpoints"},
+				Resources: []string{"clusters", "endpoints", "gateways"},
```
mangelajo, line 57: Those rbacs are for the broker cluster, we should not handle gateways on the broker cluster. This admin role is for subctl over the broker.

pkg/broker/rbac.go
```diff
@@ -81,7 +81,7 @@ func NewBrokerClusterRole() *rbacv1.Role {
...
```
mangelajo, line 84: Those rbacs are for the broker cluster, we should not handle gateways on the broker cluster.

deploy/cluster_role.yaml
```diff
@@ -21,7 +21,10 @@ rules:
   resources:
   - endpoints
   - clusters
+  - gateways
```
mangelajo, line 24: This cluster_role.yaml is only for globalnet I think. It may not need to change the permissions

skitt: OK, I’ll remove these.

PR #307, opened 2020-04-14
Publish subctl binary files on releases via GH actions

mangelajo:


mangelajo:
Let's merge and see if it works..

PR #314, opened 2020-04-17
Fix the build using the Shipyard infrastructure

skitt:
This restores the status Make variable which is needed for now, and
renames the build script (and target) to build-all to avoid a conflict
with the build directory.


mkolesnik:
I was actually working on removing the status flag last week but didnt finish, the PR is live at #315 now though this can be merged before that and we'll adjust the other one accordingly

mangelajo:
I re-ran one of the builds which failed for something unrelated:

https://travis-ci.com/github/submariner-io/submariner-operator/jobs/321143507#L2289

PR #315, opened 2020-04-19
Remove the status flag from e2e

mkolesnik:
Following removal of the status in submariner [1], it makes sense to
remove here.

[1] https://github.com/submariner-io/submariner/pull/487


mangelajo:
Looks good, I clicked on re-run for one of the jobs which failed on weavenet.

mangelajo:
The failure of this one is due to weave install...

mangelajo:
It's running but honestly I think this patch doesn't modify anything different for the globalnet one

mkolesnik:
> It's running but honestly I think this patch doesn't modify anything different for the globalnet one

Yeah KIND and weave are a bit flaky so sometimes they fail at random, maybe we can retry kind deploys in shipyard automagically but not sure about the weave ones since theyre harder to diagnose

Issue #298, opened 2020-04-06
Provision to specify globalnet-cidr as part of subctl join


sridhargaddam:
Currently --globalnet-cidr-range (default being 169.254.0.0/16) can be specified during broker-installation. Subsequently, while joining clusters, subctl will try to auto allocate a globalCIDR for the new cluster from the unallocated portion of globalnet-cidr-range.

Subctl should be enhanced to support an option where we can explicitly specify the globalCIDR if required. This will be useful for CI/KIND deployments. Along with this, it will also help in scenarios where we can choose a CIDR that is completely outside of the globalnet-cidr-range.

PR #323, opened 2020-04-21
Specify globalCIDR while joining cluster

Jaanki:
Enhance 'subctl join' to be able specify globalCIDR
for each cluster while joining. The logic implemented is

```
if globalcidr specified by user
  error if it overlaps with other clusters
  if globalcidr is preconfigured
    take pre-configured value
  else take specified value

if not specified by user
  if globalcidr is preconfigured
    take pre-configured value
  else automatically assign value
```
It also checks if the specified CIDR is valid and that both globalcidr and globalnet cluster size are not specified.

Closes: #298
Signed-off-by: Janki Chhatbar

### DIFF
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"regexp"
 	"strings"
@@ -52,7 +53,7 @@ var (
 	clusterID            string
 	serviceCIDR          string
 	clusterCIDR          string
-	globalCIDR           string
+	globalnetCIDR        string
 	repository           string
 	imageVersion         string
 	nattPort             int
@@ -95,6 +96,8 @@ func addJoinFlags(cmd *cobra.Command) {
 		"disable OpenShift's cluster version operator if necessary, without prompting")
 	cmd.Flags().UintVar(&globalnetClusterSize, "globalnet-cluster-size", 0,
 		"Cluster size for GlobalCIDR allocated to this cluster (amount of global IPs)")
+	cmd.Flags().StringVar(&globalnetCIDR, "globalnet-cidr", "",
+		"GlobalCIDR to be allocated to the cluster")
 }
 
 const (
@@ -125,6 +128,8 @@ func checkArgumentPassed(args []string) error {
 	}
 	return nil
 }
+
+var status = cli.NewStatus()
 
 func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData) {
 
@@ -191,20 +196,14 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 			colorCodes = answers.ColorCodes
 		}
 	}
-	if subctlData.GlobalnetCidrRange != "" && globalnetClusterSize != 0 && globalnetClusterSize != subctlData.GlobalnetClusterSize {
-		clusterSize, err := globalnet.GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
-		if err != nil || clusterSize == 0 {
-			exitOnError("Invalid globalnet-cluster-size", err)
-		}
-		subctlData.GlobalnetClusterSize = clusterSize
-	}
+
+	validateGlobalnetConfiguration(subctlData)
 
 	if !noLabel {
 		err := handleNodeLabels(config)
 		exitOnError("Unable to set the gateway node up", err)
 	}
 
-	status := cli.NewStatus()
 	status.Start("Deploying the Submariner operator")
 	err := submarinerop.Ensure(status, config, OperatorNamespace, operatorImage)
 	status.End(err == nil)
@@ -237,16 +236,31 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 		err = checkOverlappingClusterCidr(globalNetworks)
 		status.End(err == nil)
 		exitOnError("Error validating overlapping ClusterCIDRs", err)
-	} else if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
-		// Globalnet enabled, no globalCidr configured on this cluster
-		globalCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
-		status.End(err == nil)
-		status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalCIDR))
-		exitOnError("Globalnet failed", err)
+	} else if globalnetCIDR == "" {
+		// Globalnet enabled, GlobalCIDR not specified by the user
+		if cidrIsPreConfigured(clusterID, globalNetworks) {
+			// globalCidr already configured on this cluster
+			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
+			status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		} else {
+			// no globalCidr configured on this cluster
+			globalnetCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
+			status.End(err == nil)
+			status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalnetCIDR))
+			exitOnError("Globalnet failed", err)
+		}
 	} else {
-		// Globalnet enabled, globalCidr already configured on this cluster
-		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		// Globalnet enabled, globalnetCIDR specified by user
+		if cidrIsPreConfigured(clusterID, globalNetworks) {
+			// globalCidr pre-configured on this cluster
+			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
+			status.QueueSuccessMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
+		} else {
+			// globalCidr as specified by the user
+			err = checkOverlappingGlobalCidr(globalNetworks)
+			exitOnError(fmt.Sprintf("Error validating overlapping GlobalCIDRs %s", globalnetCIDR), err)
+			status.QueueSuccessMessage(fmt.Sprintf("GlobalCIDR is: %s", globalnetCIDR))
+		}
 	}
 
 	status.Start("Creating SA for cluster")
@@ -283,6 +297,19 @@ func checkOverlappingClusterCidr(networks map[string]*globalnet.GlobalNetwork) e
 		}
 		if overlap && k != clusterID {
 			return fmt.Errorf("invalid ClusterCIDR: %s overlaps with cluster %s", clusterCIDR, k)
+		}
+	}
+	return nil
+}
+
+func checkOverlappingGlobalCidr(networks map[string]*globalnet.GlobalNetwork) error {
+	for k, v := range networks {
+		overlap, err := globalnet.IsOverlappingCIDR(v.GlobalCIDRs, globalnetCIDR)
+		if err != nil {
+			return fmt.Errorf("unable to validate overlapping GlobalCIDR: %s", err)
+		}
+		if overlap && k != clusterID {
+			return fmt.Errorf("invalid globalnet CIDR: %s overlaps with cluster %s", globalnetCIDR, k)
 		}
 	}
 	return nil
@@ -388,6 +415,45 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
 	return scale.Spec.Replicas > 0, nil
 }
 
+func cidrIsPreConfigured(clusterID string, globalNetworks map[string]*globalnet.GlobalNetwork) bool {
+	// GlobalCIDR is not pre-configured
+	if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
+		return false
+	}
+	// GlobalCIDR is pre-configured
+	return true
+}
+
+func validateGlobalnetConfiguration(subctlData *datafile.SubctlData) {
+	if subctlData.GlobalnetCidrRange != "" && globalnetClusterSize != 0 && globalnetClusterSize != subctlData.GlobalnetClusterSize {
+		clusterSize, err := globalnet.GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
+		if err != nil || clusterSize == 0 {
+			exitOnError("Invalid globalnet-cluster-size", err)
+		}
+		subctlData.GlobalnetClusterSize = clusterSize
+	}
+
+	if globalnetCIDR != "" && globalnetClusterSize != 0 {
+		err := errors.New("Both globalnet-cluster-size and globalnet-cidr can't be specified. Specify either one.\n")
+		exitOnError("Invalid configuration", err)
+	}
+
+	if globalnetCIDR != "" {
+		_, _, err := net.ParseCIDR(globalnetCIDR)
+		exitOnError("Specified globalnet-cidr is invalid", err)
+	}
+
+	if subctlData.GlobalnetCidrRange == "" {
+		if globalnetCIDR != "" {
+			status.QueueSuccessMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cidr\n")
+			globalnetCIDR = ""
+		} else if globalnetClusterSize != 0 {
+			status.QueueSuccessMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cluster-size\n")
+			globalnetClusterSize = 0
+		}
+	}
+}
+
 func populateSubmarinerSpec(subctlData *datafile.SubctlData) submariner.SubmarinerSpec {
 	brokerURL := subctlData.BrokerURL
 	if idx := strings.Index(brokerURL, "://"); idx >= 0 {
@@ -429,9 +495,8 @@ func populateSubmarinerSpec(subctlData *datafile.SubctlData) submariner.Submarin
 		CableDriver:              cableDriver,
 		ServiceDiscoveryEnabled:  subctlData.ServiceDiscovery,
 	}
-
-	if globalCIDR != "" {
-		submarinerSpec.GlobalCIDR = globalCIDR
+	if globalnetCIDR != "" {
+		submarinerSpec.GlobalCIDR = globalnetCIDR
 	}
 	return submarinerSpec
 }


Jaanki:
The output looks like this (irrelevant portion is omitted)

1. Deploy broker on cluster1
2. Join cluster2 specifying globalcidr
```
subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster2 broker-info.subm --clusterid c2 --global-cidr 170.23.2.1/21
....
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.95.0.0/16]
        Cluster CIDRs: [10.245.0.0/16]
 ✓ Discovering multi cluster details
 ✓ GlobalCIDR is: 170.23.2.1/21
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner
```
3. Rejoin cluster2 specifying different globalcidr (note change in mask)
```
subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster2 broker-info.subm --clusterid c2 --global-cidr 170.23.2.1/22
...
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.95.0.0/16]
        Cluster CIDRs: [10.245.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [170.23.2.1/21]
 ✓ Discovering multi cluster details
 ✓ GlobalCIDR is: 170.23.2.1/21
 ✗ Pre-configured GlobalCIDR 170.23.2.1/21 detected. Not changing it.
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner
```
4. Again rejoin to verify globalCIDR submarinerspec hasn't changed
```
subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster2 broker-info.subm --clusterid c2 --global-cidr 170.23.2.1/22
...
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.95.0.0/16]
        Cluster CIDRs: [10.245.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [170.23.2.1/21]    <---- this is not changed
 ✓ Discovering multi cluster details
 ✓ GlobalCIDR is: 170.23.2.1/21
 ✗ Pre-configured GlobalCIDR 170.23.2.1/21 detected. Not changing it.
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner
```
5. Join cluster3 without specifying globalcidr so that it is automatically assigned
```
subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster3 broker-info.subm --clusterid c3
...
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.96.0.0/16]
        Cluster CIDRs: [10.246.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [170.23.2.1/21]
 ✓ Discovering multi cluster details
 ✓ Creating SA for cluster
 ✓ Allocated GlobalCIDR: 169.254.0.0/19   <-----
* Submariner is up and running
 ✓ Deploying Submariner
```
6. Rejoin cluster3 with globalcidr specified
```
subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster3 broker-info.subm --clusterid c3 --global-cidr 170.24.2.1/27
...
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.96.0.0/16]
        Cluster CIDRs: [10.246.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [170.23.2.1/21]
    Discovered global network details for Cluster c3:
        ServiceCidrs: [100.96.0.0/16]
        ClusterCidrs: [10.246.0.0/16]
        Global CIDRs: [169.254.0.0/19]
 ✓ Discovering multi cluster details
 ✓ User specifeid GlobalCIDR is: 169.254.0.0/19
 ✗ Pre-configured GlobalCIDR 169.254.0.0/19 detected. Not changing it.
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner
```

pkg/subctl/cmd/join.go
```diff
@@ -98,6 +98,8 @@ func addJoinFlags(cmd *cobra.Command) {
 		"disable OpenShift's cluster version operator if necessary, without prompting")
 	cmd.Flags().UintVar(&globalnetClusterSize, "globalnet-cluster-size", 0,
 		"Cluster size for GlobalCIDR allocated to this cluster (amount of global IPs)")
+	cmd.Flags().StringVar(&globalCIDR, "global-cidr", "",
```
sridhargaddam, line 101: Please change it to globalnet-cidr to be aligned with the existing flags.

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +270,28 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		}
 	} else {
-		// Globalnet enabled, globalCidr already configured on this cluster
-		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		// Globalnet enabled, globalCIDR specified by user
+		if cidrispreconfigured(clusterID, globalNetworks) {
+			// globalCidr pre-configured on this cluster
+			globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
+			status.QueueFailureMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalCIDR))
```
sridhargaddam, line 291: Please change the names in the debug/log messages to match the new flag (i.e., instead of GlobalCIDR, use globalnet-cidr to avoid confusion to the user)

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +270,28 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		}
+		// globalCidr as specified by the user
+		status.QueueSuccessMessage(fmt.Sprintf("GlobalCIDR is: %s", globalCIDR))
```
sridhargaddam, line 294: If user explicitly specifies the globalnet-cidr, we hit this case. Where are we validating that this globalnet-cidr does not match/overlap with any existing allocations?

@vthapar is experienced in this area of code. Vishal, can you please check if this is handled?

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +270,28 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 		err = checkOverlappingClusterCidr(globalNetworks)
 		status.End(err == nil)
 		exitOnError("Error validating overlapping ClusterCIDRs", err)
-	} else if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
-		// Globalnet enabled, no globalCidr configured on this cluster
-		globalCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
-		status.End(err == nil)
-		status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalCIDR))
-		exitOnError("Globalnet failed", err)
+	} else if globalCIDR == "" {
```
vthapar, line 273: Rework these checks. Before these checks you should check if user has specified size and cidr both. There are some checks from line 207-213. Can move them to a validateGlobalnetConfiguration() and add these checks. Throw an error that both parameters can't be used together and only one of the two should be used. If using CIDR, it should also validate it is a valid CIDR. Aim is if user supplied invalid input, we should catch it before we do expensive process of discovering networks and doing our checks. Fail fast.

Then check if it is already configured and use existing if present.

Finally handle case where cidr is specified. I think this way it will be easier to read and understand the flow.

pkg/subctl/cmd/join.go
```diff
@@ -459,10 +482,8 @@ func populateSubmarinerSpec(subctlData *datafile.SubctlData) submariner.Submarin
 		Namespace:                SubmarinerNamespace,
 		CableDriver:              cableDriver,
 		ServiceDiscoveryEnabled:  subctlData.ServiceDiscovery,
+		GlobalCIDR:               globalCIDR,
```
vthapar, line 485: Keep it the way it was. We only fill the filled if it was configured else leave it at nil. If globalnet is not configured globalCIDR will be "" and we don't want an empty non-nil field. That is assuming CR treats nil and empty fields differently.

pkg/subctl/cmd/join.go
vthapar: This is missing following checks:
1. Global-cidr specified should be a valid CIDR, that should probably be done in some validation of arguments passed before we even discover networks.
2. Check if overlaps with existing global and throw error if it does.

mkolesnik: Check #2 could be prone to races.
In any case if the user specified his own CIDR I agree it is good to do input validation that it's a valid CIDR but I wouldn't bother with checking overlap. We don't want to babysit our users, if they specify one explicitly they can just as well make sure they are specifying a distinct one.

pkg/subctl/cmd/join.go
Jaanki: GlobalCIDR has been used since before so I didnt change it.

pkg/subctl/cmd/join.go
Jaanki: you mean it does not match for 2 clusters?

sridhargaddam: We have another issue to handle the race conditions by using a ConfigMap on the Broker cluster. So let's validate the overlap check.

Jaanki: > If user explicitly specifies the globalnet-cidr, we hit this case. Where are we validating that this globalnet-cidr does not match/overlap with any existing allocations?
> 
you mean allocations for 2 clusters are same?
> @vthapar is experienced in this area of code. Vishal, can you please check if this is handled?

Jaanki: > Check #2 could be prone to races.
> In any case if the user specified his own CIDR I agree it is good to do input validation that it's a valid CIDR but I wouldn't bother with checking overlap. We don't want to babysit our users, if they specify one explicitly they can just as well make sure they are specifying a distinct one.

Agree. It is upto the user to check for non-overlapping allocations

mkolesnik: > We have another issue to handle the race conditions by using a ConfigMap on the Broker cluster. So let's validate the overlap check.

Doesn't matter, the whole idea here is to let the user specify whatever CIDR he wants so lets keep it simple.
Adding this check is IMHO unnecessary complication which doesn't add real value.

sridhargaddam: > you mean allocations for 2 clusters are same?

Currently when subctl auto-allocates globalnet-cidr from the --globalnet-cidr-range it ensures that its unique and does not overlap with existing allocations. Since this PR is adding support to specify --globalnet-cidr as part of join, before we consider the input as valid, subctl should validate the same checks on the user-provided input.

sridhargaddam: > > We have another issue to handle the race conditions by using a ConfigMap on the Broker cluster. So let's validate the overlap check.
> 
> Doesn't matter, the whole idea here is to let the user specify whatever CIDR he wants so lets keep it simple.
> Adding this check is IMHO unnecessary complication which doesn't add real value.

Noooo, it's not complicated. We already have APIs to check if there is overlap in the CIDRs. I don't understand why you are saying that lets blindly trust the user input when we can easily validate it (and save time in debugging failures that happen at run-time).

sridhargaddam: > > > We have another issue to handle the race conditions by using a ConfigMap on the Broker cluster. So let's validate the overlap check.
> > 
> > 
> > Doesn't matter, the whole idea here is to let the user specify whatever CIDR he wants so lets keep it simple.
> > Adding this check is IMHO unnecessary complication which doesn't add real value.
> 
> Noooo, it's not complicated. We already have APIs to check if there is overlap in the CIDRs. I don't understand why you are saying that lets blindly trust the user input when we can easily validate it (and save time in debugging failures that happen at run-time).

Trust me, debugging any issues which are a result of overlapping globalCIDRs will be extremely hard. This is a simple validation IMO @mkolesnik

mkolesnik: > > you mean allocations for 2 clusters are same?
> 
> Currently when subctl auto-allocates globalnet-cidr from the --globalnet-cidr-range it ensures that its unique and does not overlap with existing allocations. Since this PR is adding support to specify --globalnet-cidr as part of join, before we consider the input as valid, subctl should validate the same checks on the user-provided input.

This is expected though since its part of the basics of the operating mechanism of the way it allocates CIDRs.

> Noooo, it's not complicated. We already have APIs to check if there is overlap in the CIDRs. I don't understand why you are saying that lets blindly trust the user input when we can easily validate it (and save time in debugging failures that happen at run-time).

Why do you think you're smarter than the user in this case?
Perhaps the user wants to do it for a certain reason, it has nothing to do with validation being easy or hard.

sridhargaddam: > Why do you think you're smarter than the user in this case?

lol, this is not the point of discussion.

> Perhaps the user wants to do it for a certain reason, it has nothing to do with validation being easy or hard.

User can provide any input (sometimes intentional and sometimes unintentional) and we have checks in code to ensure that we gracefully handle wrong input and warn him at a very early stage. The question is - does Submariner Globalnet support "Overlapping" globalCIDRs. The plain answer is NO, so we should warn him at the deployment stage itself.

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +270,28 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 		exitOnError("Error validating overlapping ClusterCIDRs", err)
-	} else if globalNetworks[clusterID] == nil || globalNetworks[clusterID].GlobalCIDRs == nil || len(globalNetworks[clusterID].GlobalCIDRs) <= 0 {
-		// Globalnet enabled, no globalCidr configured on this cluster
-		globalCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
-		status.End(err == nil)
-		status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalCIDR))
-		exitOnError("Globalnet failed", err)
+	} else if globalCIDR == "" {
+		// Globalnet enabled, GlobalCIDR not specified by the user
+		if cidrispreconfigured(clusterID, globalNetworks) {
```
mangelajo, line 275: nit: use cammelcase to separate words : cidrIsPreConfigured so it's easier to read the function name.

pkg/subctl/cmd/join.go
mkolesnik: > User can provide any input (sometimes intentional and sometimes unintentional) and we have checks in code to ensure that we gracefully handle wrong input and warn him at a very early stage. The question is - does Submariner Globalnet support "Overlapping" globalCIDRs. The plain answer is NO, so we should warn him at the deployment stage itself.

OK than I propose a settlement - that we can do a "detection" if you desire, but not block the operation. Print a warning so the user knows, so he can know and decide if that's what he wanted or not.

mangelajo: I agree with sridhar here, we must do validation. If we have a way of detecting how the admin is shooting his feet, lets prevent it. Think of this as a gun safelock... 

What we couldn't prevent for example is the admin editing the ConfigMap on the broker (when we have that..) and that will be his way to override things if the admin knows what he's doing

skitt: If an admin forces the use of a conflicting CIDR, what would the consequences be? If it could break a running cluster, then it seems to me we should prevent it by default...

vthapar: Checking user input for invalid inputs is UX 101. Adding a cluster with overlapping CIDR can break things. Not letting user shoot itself in the foot is always good design. Not doing validations coz there is a race condition sounds more like a race condition problem that should be fix than validation problem.

mkolesnik: By your guys standards then `rm` should prevent running on `/` if `-r` is specified, since the user is stupid and can seriously screw up something.
Yet, we have many systems and languages that allow users more flexibility and dont babysit him, even if he can "shoot himself in the foot".

So if you want to be holier than the pope, by all means cotton wrap the user so the poor thing doesnt break anything by accident. I prefer to walk in the footsteps of giants and let him shoot himself if the input is valid :)

* And yes, even windows which is much more restrictive to a "dumb user" will allow him to format his OS drive quite easily..

mangelajo: The consequences I believe would be conflicting iptable rules, and route rules, may be the route agent would end in a crash loop (not sure of how error handling happens there)... @sridhargaddam probably you have more insight.

Something that came up while talking with mike about this, is that... at a minimun we may need to support removal of globalCIDRs from clusters, ... 

For example, if you keep adding and removing clusters, ... we don't handle cleanups properly... or what if you gave more CIDRs to a cluster than you wanted?, there must be a way to fix it.

sridhargaddam: > The consequences I believe would be conflicting iptable rules, and route rules, may be the route agent would end in a crash loop (not sure of how error handling happens there)... @sridhargaddam probably you have more insight.

If the user-provides a globalnet-cidr that overlaps with globalnet-cidr of remote cluster, then cross-cluster connectivity will fail. I dont think it will affect local cluster traffic. 
Regarding a crash in route-agent, I don't think so based on my understanding of route-agent code. When overlapping globalCIDRs are used during deployment, it would make debugging lot more difficult as there will not be any crash but traffic could go in totally different ways than what we anticipate/program. So, it's better to ensure that we are validating the input at deployment time.

> 
> Something that came up while talking with mike about this, is that... at a minimun we may need to support removal of globalCIDRs from clusters, ...

IMHO, I would say that we should start looking at cleanup (of submariner resources) in general which is missing today. 

> 
> For example, if you keep adding and removing clusters, ... we don't handle cleanups properly... or what if you gave more CIDRs to a cluster than you wanted?, there must be a way to fix it.

Right, this is more about updating the globalCIDR of a cluster where updation can be expanding/shrinking/allocating-newCIDR. Yes, we should support this use-case as a new PR.

dfarrell07: +1 to not giving the user the option to shoot themselves.

pkg/subctl/cmd/join.go
Jaanki: Added check for invalid CIDR and overlapping CIDR

Jaanki:
Test for overlapping GlobalCIDR specified

1. Deploy cluster2 with globalcidr 10.24.2.1/28
2. Deploy cluster3 with the same globalcidr
```
$ ../submariner-operator/bin/subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster3 broker-info.subm --clusterid c3 --globalnet-cidr 10.24.2.1/28
.....
* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.96.0.0/16]
        Cluster CIDRs: [10.246.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [10.24.2.1/28]
**Error validating overlapping GlobalCIDRs: invalid globalnet CIDR: 10.24.2.1/28 overlaps with cluster c2**
subctl version: dev
built from git commit id: bc3c106bcfdb8a8a538126f2439d6eda21bfddad
```
3. Deploy cluster3 with non-overlapping global CIDR
```
../submariner-operator/bin/subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster3 broker-info.subm --clusterid c3 --globalnet-cidr 100.24.2.1/28

* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.96.0.0/16]
        Cluster CIDRs: [10.246.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [10.24.2.1/28]    <------------ for c2
 ✓ Discovering multi cluster details
 ✓ GlobalCIDR is: 100.24.2.1/28   <------------ for c3
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner
```
4. Rejoin cluster3 (c3) with the separate non-overlapping global cidr
```
$ ../submariner-operator/bin/subctl join --kubeconfig output/kind-config/local-dev/kind-config-cluster3 broker-info.subm --clusterid c3 --globalnet-cidr 100.24.2.1/29

* Discovering network details
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs: [100.96.0.0/16]
        Cluster CIDRs: [10.246.0.0/16]
    Discovered global network details for Cluster c2:
        ServiceCidrs: [100.95.0.0/16]
        ClusterCidrs: [10.245.0.0/16]
        Global CIDRs: [10.24.2.1/28]
    Discovered global network details for Cluster c3:
        ServiceCidrs: [100.96.0.0/16]
        ClusterCidrs: [10.246.0.0/16]
        Global CIDRs: [100.24.2.1/28]
 ✓ Discovering multi cluster details
 ✓ GlobalCIDR is: 100.24.2.1/28
 ✗ Pre-configured GlobalCIDR 100.24.2.1/28 detected. Not changing it.    <-----------
 ✓ Creating SA for cluster
* Submariner is up and running
 ✓ Deploying Submariner

```

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +266,31 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+			status.End(err == nil)
+			status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalCIDR))
+			exitOnError("Globalnet failed", err)
+		}
 	} else {
-		// Globalnet enabled, globalCidr already configured on this cluster
-		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		// Globalnet enabled, globalCIDR specified by user
+		err = checkOverlappingGlobalCidr(globalNetworks)
```
vthapar, line 284: Make this function also take globalCidr as argument like other overlappingCIdr functions and pass it. Will be useful when we allow multiple globalCIdrs in future.

pkg/subctl/cmd/join.go
vthapar: Agree with Sridhar. If there is error, user needs to know which input parameter caused error. It was fine earlier because there was no user parameter for globalcidr, it was an internal variable so it was more for debugging than informing user. But now, we should tell the exact parameter that was passed wrong.

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +266,31 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		exitOnError("Error validating overlapping GlobalCIDRs", err)
```
vthapar, line 285: Include `globalnet-cidr: %s` and pass cidr passed by user as value in error message.

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +445,35 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+		clusterSize, err := globalnet.GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
+		if err != nil || clusterSize == 0 {
+			exitOnError("Invalid globalnet-cluster-size", err)
+		}
+		subctlData.GlobalnetClusterSize = clusterSize
+	}
+
+	if globalCIDR != "" && globalnetClusterSize != 0 {
+		err := errors.New("Both Globalnet CIDR and Globalnet Cluster size can't be specified. Specify either one")
+		exitOnError("Error:", err)
```
vthapar, line 468: nitpick: Instead of "Error" use "Invalid configuration". exitOnError

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +445,35 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+	if subctlData.GlobalnetCidrRange != "" && globalnetClusterSize != 0 && globalnetClusterSize != subctlData.GlobalnetClusterSize {
+		clusterSize, err := globalnet.GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
+		if err != nil || clusterSize == 0 {
+			exitOnError("Invalid globalnet-cluster-size", err)
+		}
+		subctlData.GlobalnetClusterSize = clusterSize
+	}
+
+	if globalCIDR != "" && globalnetClusterSize != 0 {
+		err := errors.New("Both Globalnet CIDR and Globalnet Cluster size can't be specified. Specify either one")
```
vthapar, line 467: Use param names passed by user in error message instead of variable names. globalnet-cluster-size and globalnet-cidr

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +445,35 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+		exitOnError("Error:", err)
+	}
+
+	if globalCIDR != "" {
+		_, _, err := net.ParseCIDR(globalCIDR)
+		exitOnError("Specified CIDR is invalid", err)
```
vthapar, line 473: Use parameter name globalnet-cidr in error message. Something like "globalnet-cidr is invalid"

pkg/subctl/cmd/join.go
mangelajo: Also the code becomes clearer, I know he value is already a global, but code will be more self documented in that way.

pkg/subctl/cmd/join.go
Jaanki: Done

pkg/subctl/cmd/join.go
Jaanki: That's a separate function you are referring as per our discussion on slack. Not changing this for the time being. Will revisit if needed.

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +266,31 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+			status.End(err == nil)
+			status.QueueSuccessMessage(fmt.Sprintf("Allocated GlobalCIDR: %s", globalnetCIDR))
+			exitOnError("Globalnet failed", err)
+		}
 	} else {
-		// Globalnet enabled, globalCidr already configured on this cluster
-		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		// Globalnet enabled, globalnetCIDR specified by user
+		err = checkOverlappingGlobalCidr(globalNetworks)
```
sridhargaddam, line 284: Move this check to else part of Line 287
i.e., if the cidrIsPreConfigured for the cluster, no need to validate for overlappingGlobalCIdr()

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +445,35 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+func validateGlobalnetConfiguration(subctlData *datafile.SubctlData) {
+	if subctlData.GlobalnetCidrRange != "" && globalnetClusterSize != 0 && globalnetClusterSize != subctlData.GlobalnetClusterSize {
+		clusterSize, err := globalnet.GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
+		if err != nil || clusterSize == 0 {
+			exitOnError("Invalid globalnet-cluster-size", err)
+		}
+		subctlData.GlobalnetClusterSize = clusterSize
+	}
+
+	if globalnetCIDR != "" && globalnetClusterSize != 0 {
```
sridhargaddam, line 466: Similarly, add a check that if subctlData.GlobalnetCidrRange is empty (i.e., Globalnet is not enabled during deploy-broker), but during join operation, user provides globalnet-cluster-size (and/or globalnet-cidr), then these values are ignored.
for example: Globalnet is not enabled, so ignoring globalnet-cluster-size (or globalnet-cidr)

@vthapar @mangelajo  Ignoring should be fine right? Or do you suggest that we exitOnError?

pkg/subctl/cmd/join.go
Jaanki: That was my initial implementation which somehow didnot work. So moved it here.

pkg/subctl/cmd/join.go
Jaanki: I think ignoring with a log line should be fine.

pkg/subctl/cmd/join.go
Jaanki: And also this checks if the ip overlaps with the preconfigured value as well (which won't be the case currently) but might be useful in future when we allow passing globalnet-cidr by means other than join

pkg/subctl/cmd/join.go
vthapar: Ignore with a log should be fine.

pkg/subctl/cmd/join.go
vthapar: Could you clarify what didn't work? validation for overlapping or preconfigured check?

Jaanki: Rechecked again...seems to work fine. Don't remember what was the exact issue last time.

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +266,30 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		}
 	} else {
-		// Globalnet enabled, globalCidr already configured on this cluster
-		globalCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-		status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+		// Globalnet enabled, globalnetCIDR specified by user
+		if cidrIsPreConfigured(clusterID, globalNetworks) {
+			// globalCidr pre-configured on this cluster
+			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
+			status.QueueFailureMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
```
sridhargaddam, line 287: If "Pre-configured GlobalCIDR is detected", why to show it as failure?

pkg/subctl/cmd/join.go
```diff
@@ -268,16 +266,30 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		}
+		// globalCidr as specified by the user
+		err = checkOverlappingGlobalCidr(globalNetworks)
+		exitOnError(fmt.Sprintf("Error validating overlapping GlobalCIDRs %s", globalnetCIDR), err)
+		status.QueueSuccessMessage(fmt.Sprintf("GlobalCIDR is: %s", globalnetCIDR))
```
sridhargaddam, line 292: I think this should be in the else part

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +444,41 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+		exitOnError("Invalid configuration:", err)
+	}
+
+	if globalnetCIDR != "" {
+		_, _, err := net.ParseCIDR(globalnetCIDR)
+		exitOnError("Specified globalnet-cidr is invalid", err)
+	}
+
+	if subctlData.GlobalnetCidrRange == "" && (globalnetCIDR != "" || globalnetClusterSize != 0) {
+		fmt.Printf("* globalnet is not enabled in broker. Ignoring specified globalnet-cidr or globalnet-cluster-size\n")
```
sridhargaddam, line 476: Do we get fmt.Printf messages when we run subctl? 
Should we actually use status.QueueSuccessMessage here?

vthapar, line 476: @skitt thoughts on using Printf? I know you wanted to avoid using Printf in subctl code but we don't have a logging equivalent for warning messages.

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +444,41 @@ func isOpenShiftCVOEnabled(config *rest.Config) (bool, error) {
...
+		err := errors.New("Both globalnet-cluster-size and globalnet-cidr can't be specified. Specify either one\n")
+		exitOnError("Invalid configuration:", err)
+	}
+
+	if globalnetCIDR != "" {
+		_, _, err := net.ParseCIDR(globalnetCIDR)
+		exitOnError("Specified globalnet-cidr is invalid", err)
+	}
+
+	if subctlData.GlobalnetCidrRange == "" && (globalnetCIDR != "" || globalnetClusterSize != 0) {
```
sridhargaddam, line 475: The log message should reflect the exact input.
So, please change it as follows.
   if globalnetCIDR != "" {
   ... 
   // Globanet not enabled on Broker, ignoring globalnet-cidr
   } else if globalnetClusterSize != 0 {
   ...
   // Globanet not enabled on Broker, ignoring globalnet-cluster-size
  }
}

pkg/subctl/cmd/join.go
Jaanki: yes we get printf output

Jaanki: eg: fmt.Printf("* Discovering network details\n")

Jaanki: fmt.Printf("* Discovering network details\n") :P

pkg/subctl/cmd/join.go
Jaanki: Failure is just for logging. the process does not actually fail. This prints 'x' in red so that it is highlighted and user is aware of the ignorance.

Jaanki: like this 
`✗ Pre-configured GlobalCIDR 100.24.2.1/28 detected. Not changing it.`

pkg/subctl/cmd/join.go
skitt: Yes, this should use either `status.QueueSuccessMessage` or `status.QueueFailureMessage`, depending on the desired format.

pkg/subctl/cmd/join.go
Jaanki: I don't think this differentiation is needed. User can identify what's wrong very easily.

pkg/subctl/cmd/join.go
Jaanki: This should be fine too as the if block is just a warning and does not afffect the completion of command.

pkg/subctl/cmd/join.go
sridhargaddam: thanks @skitt 

I'd prefer to use status.QueueSuccessMessage here as its not a failure. If we show it as a failure, user might think that something is broken as part of the deployment. @vthapar your comments on this?

pkg/subctl/cmd/join.go
sridhargaddam: Same as my comment below. I would prefer to use status.QueueSuccessMessage instead of showing it as a failure. The message is self-explanatory as it says - since pre-configured globalCIDR is detected, we are not changing it.
CC: @vthapar

pkg/subctl/cmd/join.go
sridhargaddam: Since we already found a pre-configured globalCIDR and are planning to re-use it, why do we have to check for OverlappingGlobalCIDR (for the user-supplied value)? So, this validation should be in else.

pkg/subctl/cmd/join.go
sridhargaddam: I'm a user and I would get confused when I see a log message that says "Ignoring specified globalnet-cidr or globalnet-cluster-size"

pkg/subctl/cmd/join.go
Jaanki: Changed it to status.QueueSuccessMessage

pkg/subctl/cmd/join.go
Jaanki: DOne

pkg/subctl/cmd/join.go
Jaanki: DOne

pkg/subctl/cmd/join.go
Jaanki: DOne

Jaanki: Done

mangelajo:
Thank you for the patch and addressing all comments! :)

pkg/subctl/cmd/join.go
skitt: > By your guys standards then `rm` should prevent running on `/` if `-r` is specified, since the user is stupid and can seriously screw up something.

`rm` (at least, GNU `rm`) does check for that ;-).

PR #328, opened 2020-04-23
Include Gateway information in the operator status

skitt:
This also drops the CableDriver entry since the cable driver in use is
reflected (accurately) in the Gateway information.


mangelajo:
@skitt I wonder if at this point makes sense to completely replicate the gateway list. Would it make more sense to report an overall status of the gateways?, like, for example programatically going over them, and reporting, which one is the active, and if all connections are operational... or if there is any failing connection?, so the admin can go and check at the gateway?

PR #330, opened 2020-04-23
Fix deploy-broker status msg for service discovery

vthapar:
Setup Service Discovery step of deploy-broker gives incorrect status
msg. We no longer deploy controller on broker.


vthapar:
https://github.com/submariner-io/submariner-operator/pull/331

PR #337, opened 2020-04-27
Refactor function to check overlapping CIDRs

Jaanki:
checkOverlappingServiceCidr and checkOverlappingClusterCidr
replaced by checkOverlappingCidr.

### DIFF
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -18,10 +18,12 @@ package globalnet
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/bits"
 	"net"
 
+	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/datafile"
 	submarinerClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/discovery/globalnet/globalnet.go
```diff
@@ -32,6 +32,7 @@ type GlobalNetwork struct {
 	ServiceCIDRs []string
 	ClusterCIDRs []string
 	ClusterId    string
+	//Cidrlist     []string
```
vthapar, line 35: I had same idea but using a custom CidrType and using switch case. Something like this:
https://play.golang.org/p/cKj4srS3dnl

pkg/subctl/cmd/join.go
```diff
@@ -262,10 +262,11 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 
 	if subctlData.GlobalnetCidrRange == "" {
 		// Globalnet not enabled
-		err = checkOverlappingServiceCidr(globalNetworks)
+		err = checkOverlappingCidr(globalNetworks, "service")
```
mangelajo, line 265: passing a string as a parameter to specify what to check is error prone, the developer could
pass a wrong string and compiler won't detect anything.

At least I'd use consts, 
```go
type OverlapCheck int

const (
             ClusterOverlap OverlapCheck = iota
             ServiceOverlap
             GlobalnetOverlap
)
```

pkg/subctl/cmd/join.go
```diff
@@ -262,10 +262,11 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 		status.End(err == nil)
 		exitOnError("Error validating overlapping ServiceCIDRs", err)
-		err = checkOverlappingClusterCidr(globalNetworks)
+		// err = checkOverlappingClusterCidr(globalNetworks)
```
mangelajo, line 268:
```suggestion
```

PR #339, opened 2020-04-27
Use Shipyard's deploy target

mkolesnik:
Use shipyards target for deployment and provide it with extra flags to
affect the deployment.


.travis.yml
```diff
@@ -13,8 +13,12 @@ git:
 jobs:
   include:
     - env: CMD="make ci"
-    - env: CMD="make e2e lighthouse=true"
-    - env: CMD="make e2e lighthouse=true globalnet=true" DEPLOY=true
+    - env: CMD="make e2e"
+           DEPLOY_ARGS="--deploytool_broker_args '--service-discovery'"
```
mangelajo, line 17: strings within strings :exploding_head: ;-)

README.md
```diff
@@ -143,10 +143,9 @@ make e2e cleanup
 ## Setup development environment 
 You will need docker installed in your system, and at least 8GB of RAM. Run:
 
- ```
- make e2e
- ```
- 
+```
+make deploy
```
mangelajo, line 147: Probably we should move the "print out" of the KUBECONFIG details from the kind-e2e script to the end of deploy, or better, to a common function that we can call from both...

So it prints out at the end of e2e, and the end of deploy, where the user most likely see it..

tpantelis: I put that into a util function with https://github.com/submariner-io/shipyard/pull/113. I was thinking it should be printed at the end of _clusters.sh_ since that's where the clusters are created and maybe also at the end of the e2e script.

mkolesnik: Ideally it would be best if it printed once at the end of the run, but perhaps thats not easily possible with Make.

PR #351, opened 2020-04-30
Use specific cluster functions

mkolesnik:
After introduction of cluster functions [1] we can use them to make our
code nicer and smarter.

[1] https://github.com/submariner-io/shipyard/pull/127


mkolesnik:
This one needs https://github.com/submariner-io/shipyard/pull/127 merged first, so it'll fail until it is

Issue #355, opened 2020-05-06
subctl version should give release versions


vthapar:
subctl version only shows commit ID of git commit from which it was built. It should also include release version information e.g. 0.2.0, 0.3.0-rc1 etc. It will make it easier to troubleshoot without having to match commit to particular release.

PR #356, opened 2020-05-06
Use the closest git reference instead of the commit hash

skitt:
In subctl version, display the output of "git describe --tags" instead
of the full commit hash. On development versions, this will show the
closest tag, followed by the number of commits since the tag, and the
short hash of the current commit:

	$ bin/subctl version
	subctl version: dev
	built from git ref: v0.3.0-rc2-2-g95d2c84

When subctl is built from a commit which is itself tagged (i.e. when
building a release), it will only show the tag:

	$ bin/subctl version
	subctl version: v0.3.0
	built from git ref: v0.3.0

Fixes: #355

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ branches:
   - /^release-\d+.\d+.*$/
  
 git:
-  depth: 1
+  depth: false
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


skitt:
This is a quick fix for #355; looking at the build process, since we’re tag-based, it should be possible to get rid of the whole `VERSION` thing entirely...

PR #363, opened 2020-05-06
Minimise the operator image

skitt:
The container image only needs the operator binary, which is a static
binary (the operator SDK builds it that way).


skitt:
I need to test this with OCP.

build/Dockerfile
```diff
@@ -1,21 +1,10 @@
-FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
+FROM scratch
 
-RUN rpm -e pinentry gnupg2 gpgme librepo libdnf microdnf pth \
-           shared-mime-info glib2 json-glib gobject-introspection libpeas \
-           librhsm pkgconfig libmodulemd libmount expat libsolv libyaml \
-           libsmartcols json-c libgcrypt libxml2 libblkid libassuan libuuid \
-           libgpg-error
-
-ENV OPERATOR=/usr/local/bin/submariner-operator \
```
mangelajo, line 9: We don't need those env vars anymore?

skitt: No, there were used by the startup scripts, which we no longer need. (In fact I need to delete them...)

skitt:
I’ve tested this on OCP, with the `quay.io/submariner/submariner-operator:mintest2`, and it works:

Screenshot from 2020-05-07 15-40-55

PR #371, opened 2020-05-08
Fetch whole depth, fixes E2E GH actions (GIT_REF failure)

mangelajo:
See: https://github.com/submariner-io/submariner-operator/pull/370/checks?check_run_id=655958357#step:4:1123

PR #372, opened 2020-05-08
Revert "Use the closest git reference instead of the commit hash"

mangelajo:
This reverts commit 98f69a1b7cfc620d4770155b02eae202e73d0c9f.

It breaks the build of subctl in GH actions, which we use to push binaries to the release, I tried this:

https://github.com/submariner-io/submariner-operator/pull/371

But it doesn't work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ branches:
   - /^release-\d+.\d+.*$/
  
 git:
-  depth: false
+  depth: 1
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


dfarrell07:
I was wondering about that too, looking at the GH Actions logs.

mangelajo:
> I was wondering about that too, looking at the GH Actions logs.

I think it's the way GH actions fetches from git.

mangelajo:
Ok this seems to fix it: https://github.com/submariner-io/submariner-operator/pull/371

PR #374, opened 2020-05-12
Move away from Printf

Jaanki:
Replace Printf usage with status Queues in subctl

Colour code for WARNING:

Screenshot from 2020-05-12 20-56-17

### DIFF
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -308,14 +308,14 @@ func ValidateGlobalnetConfiguration(subctlData *datafile.SubctlData, netconfig C
 
 	if subctlData.GlobalnetCidrRange == "" {
 		if globalnetCIDR != "" {
-			status.QueueSuccessMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cidr")
+			status.QueueWarningMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cidr")
 			globalnetCIDR = ""
 		} else if globalnetClusterSize != 0 {
-			status.QueueSuccessMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cluster-size")
+			status.QueueWarningMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cluster-size")
 			globalnetClusterSize = 0
 		}
 	}
-	status.End(true)
+	status.End(cli.Success)
 	return nil, globalnetCIDR, globalnetClusterSize
 }
 
@@ -348,7 +348,7 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
 		if isCIDRPreConfigured(clusterID, globalNetworks) {
 			// globalCidr already configured on this cluster
 			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-			status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+			status.QueueWarningMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
 		} else {
 			// no globalCidr configured on this cluster
 			globalnetCIDR, err := AllocateGlobalCIDR(globalNetworks, subctlData)
@@ -362,7 +362,7 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
 		if isCIDRPreConfigured(clusterID, globalNetworks) {
 			// globalCidr pre-configured on this cluster
 			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-			status.QueueSuccessMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
+			status.QueueWarningMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
 		} else {
 			// globalCidr as specified by the user
 			err := CheckOverlappingCidr(globalNetworks, global, netconfig)
@@ -372,7 +372,7 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
 			status.QueueSuccessMessage(fmt.Sprintf("GlobalCIDR is: %s", globalnetCIDR))
 		}
 	}
-	status.End(true)
+	status.End(cli.Success)
 	return globalnetCIDR, nil
 }
 
--- a/pkg/internal/cli/status.go
+++ b/pkg/internal/cli/status.go
@@ -25,6 +25,14 @@ import (
 	"github.com/submariner-io/submariner-operator/pkg/internal/log"
 )
 
+type Result int
+
+const (
+	Success Result = iota
+	Failure
+	Warning
+)
+
 // Status is used to track ongoing status in a CLI, with a nice loading spinner
 // when attached to a terminal
 type Status struct {
@@ -34,9 +42,11 @@ type Status struct {
 	// for controlling coloring etc
 	successFormat string
 	failureFormat string
+	warningFormat string
 	// message queues
 	successQueue []string
 	failureQueue []string
+	warningQueue []string
 }
 
 func NewStatus() *Status {
@@ -55,17 +65,20 @@ func StatusForLogger(l log.Logger) *Status {
 		logger:        l,
 		successFormat: " ✓ %s\n",
 		failureFormat: " ✗ %s\n",
+		warningFormat: " ⚠ %s\n",
 		successQueue:  []string{},
 		failureQueue:  []string{},
+		warningQueue:  []string{},
 	}
 	// if we're using the CLI logger, check for if it has a spinner setup
 	// and wire the status to that
 	if v, ok := l.(*Logger); ok {
 		if v2, ok := v.writer.(*Spinner); ok {
 			s.spinner = v2
-			// use colored success / failure messages
+			// use colored success / failure / warning messages
 			s.successFormat = " \x1b[32m✓\x1b[0m %s\n"
 			s.failureFormat = " \x1b[31m✗\x1b[0m %s\n"
+			s.warningFormat = " \x1b[33m⚠\x1b[0m %s\n"
 		}
 	}
 	return s
@@ -74,7 +87,7 @@ func StatusForLogger(l log.Logger) *Status {
 // Start starts a new phase of the status, if attached to a terminal
 // there will be a loading spinner with this status
 func (s *Status) Start(status string) {
-	s.End(true)
+	s.End(Success)
 	// set new status
 	s.status = status
 	if s.spinner != nil {
@@ -87,7 +100,7 @@ func (s *Status) Start(status string) {
 
 // End completes the current status, ending any previous spinning and
 // marking the status as success or failure
-func (s *Status) End(success bool) {
+func (s *Status) End(output Result) {
 	if s.status == "" {
 		return
 	}
@@ -96,10 +109,13 @@ func (s *Status) End(success bool) {
 		s.spinner.Stop()
 		fmt.Fprint(s.spinner.writer, "\r")
 	}
-	if success {
+	switch output {
+	case Success:
 		s.logger.V(0).Infof(s.successFormat, s.status)
-	} else {
+	case Failure:
 		s.logger.V(0).Infof(s.failureFormat, s.status)
+	case Warning:
+		s.logger.V(0).Infof(s.warningFormat, s.status)
 	}
 
 	for _, message := range s.successQueue {
@@ -108,10 +124,14 @@ func (s *Status) End(success bool) {
 	for _, message := range s.failureQueue {
 		s.logger.V(0).Infof(s.failureFormat, message)
 	}
+	for _, message := range s.warningQueue {
+		s.logger.V(0).Infof(s.warningFormat, message)
+	}
 
 	s.status = ""
 	s.successQueue = []string{}
 	s.failureQueue = []string{}
+	s.warningQueue = []string{}
 }
 
 // QueueSuccessMessage queues up a message, which will be displayed once
@@ -124,4 +144,18 @@ func (s *Status) QueueSuccessMessage(message string) {
 // the status ends (using the failure format)
 func (s *Status) QueueFailureMessage(message string) {
 	s.failureQueue = append(s.failureQueue, message)
+}
+
+// QueuewarningMessage queues up a message, which will be displayed once
+// the status ends (using the warning format)
+func (s *Status) QueueWarningMessage(message string) {
+	s.warningQueue = append(s.warningQueue, message)
+}
+
+func CheckForError(err error) Result {
+	if err == nil {
+		return Success
+	} else {
+		return Failure
+	}
 }
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -83,7 +83,7 @@ var deployBroker = &cobra.Command{
 		status := cli.NewStatus()
 		status.Start("Deploying broker")
 		err = broker.Ensure(config)
-		status.End(err == nil)
+		status.End(cli.CheckForError(err))
 		exitOnError("Error deploying the broker", err)
 
 		status.Start(fmt.Sprintf("Creating %s file", brokerDetailsFilename))
@@ -92,7 +92,7 @@ var deployBroker = &cobra.Command{
 		if ipsecSubmFile == "" {
 			if _, err := datafile.NewFromFile(brokerDetailsFilename); err == nil {
 				ipsecSubmFile = brokerDetailsFilename
-				status.QueueSuccessMessage(fmt.Sprintf("Reusing IPsec PSK from existing %s", brokerDetailsFilename))
+				status.QueueWarningMessage(fmt.Sprintf("Reusing IPsec PSK from existing %s", brokerDetailsFilename))
 			} else {
 				status.QueueSuccessMessage(fmt.Sprintf("A new IPsec PSK will be generated for %s", brokerDetailsFilename))
 			}
@@ -117,7 +117,7 @@ var deployBroker = &cobra.Command{
 		}
 
 		err = subctlData.WriteToFile(brokerDetailsFilename)
-		status.End(err == nil)
+		status.End(cli.CheckForError(err))
 		exitOnError("Error writing the broker information", err)
 
 		if enableDataplane {
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -201,7 +201,7 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	}
 
 	var err error
-	fmt.Printf("* Discovering network details\n")
+	status.Start("Discovering network details")
 	networkDetails := getNetworkDetails(config)
 
 	serviceCIDR, err = getServiceCIDR(serviceCIDR, networkDetails)
@@ -230,13 +230,13 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 
 	status.Start("Deploying the Submariner operator")
 	err = submarinerop.Ensure(status, config, OperatorNamespace, operatorImage)
-	status.End(err == nil)
+	status.End(cli.CheckForError(err))
 	exitOnError("Error deploying the operator", err)
 
 	if subctlData.ServiceDiscovery {
 		status.Start("Deploying multi cluster service discovery")
 		err = lighthouse.Ensure(status, config, "", "", false, kubeConfig, kubeContext)
-		status.End(err == nil)
+		status.End(cli.CheckForError(err))
 		exitOnError("Error deploying multi cluster service discovery", err)
 	}
 
@@ -244,12 +244,18 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	brokerAdminClientset, err := kubernetes.NewForConfig(subctlData.GetBrokerAdministratorConfig())
 	exitOnError("Error retrieving broker admin config", err)
 	clienttoken, err = broker.CreateSAForCluster(brokerAdminClientset, clusterID)
-	status.End(err == nil)
+	status.End(cli.CheckForError(err))
 	exitOnError("Error creating SA for cluster", err)
 
 	status.Start("Deploying Submariner")
 	err = submarinercr.Ensure(config, OperatorNamespace, populateSubmarinerSpec(subctlData))
-	status.End(err == nil)
+	if err == nil {
+		status.QueueSuccessMessage("Submariner is up and running")
+		status.End(cli.Success)
+	} else {
+		status.QueueFailureMessage("Submariner deployment failed")
+		status.End(cli.Failure)
+	}
 	exitOnError("Error deploying Submariner", err)
 }
 
@@ -259,7 +265,7 @@ func getNetworkDetails(config *rest.Config) *network.ClusterNetwork {
 	exitOnError("Unable to set the Kubernetes cluster connection up", err)
 	networkDetails, err := network.Discover(dynClient, clientSet)
 	if err != nil {
-		fmt.Printf("Error trying to discover network details: %s\n", err)
+		status.QueueWarningMessage(fmt.Sprintf("Error trying to discover network details: %s", err))
 	} else if networkDetails != nil {
 		networkDetails.Show()
 	}
@@ -269,8 +275,8 @@ func getNetworkDetails(config *rest.Config) *network.ClusterNetwork {
 func getPodCIDR(clusterCIDR string, nd *network.ClusterNetwork) (string, error) {
 	if clusterCIDR != "" {
 		if nd != nil && len(nd.PodCIDRs) > 0 && nd.PodCIDRs[0] != clusterCIDR {
-			fmt.Printf("WARNING: your provided cluster CIDR for the pods (%s) does not match discovered (%s)\n",
-				clusterCIDR, nd.PodCIDRs[0])
+			status.QueueWarningMessage(fmt.Sprintf("Your provided cluster CIDR for the pods (%s) does not match discovered (%s)\n",
+				clusterCIDR, nd.PodCIDRs[0]))
 		}
 		return clusterCIDR, nil
 	} else if nd != nil && len(nd.PodCIDRs) > 0 {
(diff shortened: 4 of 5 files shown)


pkg/internal/cli/status.go
```diff
@@ -56,17 +66,20 @@ func StatusForLogger(l log.Logger) *Status {
 		logger:        l,
 		successFormat: " ✓ %s\n",
 		failureFormat: " ✗ %s\n",
+		warningFormat: " # %s\n",
```
mangelajo, line 69:
```suggestion
		warningFormat: " ⚠ %s\n",
```

pkg/internal/cli/status.go
```diff
@@ -75,7 +88,7 @@ func StatusForLogger(l log.Logger) *Status {
 // Start starts a new phase of the status, if attached to a terminal
 // there will be a loading spinner with this status
 func (s *Status) Start(status string) {
-	s.End(true)
+	s.End(0)
```
mangelajo, line 91:
```suggestion
	s.End(success)
```

pkg/subctl/cmd/deploybroker.go
```diff
@@ -83,7 +83,7 @@ var deployBroker = &cobra.Command{
 		status := cli.NewStatus()
 		status.Start("Deploying broker")
 		err = broker.Ensure(config)
-		status.End(err == nil)
+		status.End(0)
```
mangelajo, line 86: This is not equivalent, may be you can have an auxiliary function that takes err and return status ?

pkg/subctl/cmd/deploybroker.go
```diff
@@ -117,7 +117,7 @@ var deployBroker = &cobra.Command{
 		}
 
 		err = subctlData.WriteToFile(brokerDetailsFilename)
-		status.End(err == nil)
+		status.End(0)
```
mangelajo, line 120: not equivalent

pkg/subctl/cmd/join.go
```diff
@@ -206,17 +206,17 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 
 	status.Start("Deploying the Submariner operator")
 	err := submarinerop.Ensure(status, config, OperatorNamespace, operatorImage)
-	status.End(err == nil)
+	status.End(0)
```
mangelajo, line 209: ditto

pkg/subctl/cmd/join.go
```diff
@@ -206,17 +206,17 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 	exitOnError("Error deploying the operator", err)
 
 	if subctlData.ServiceDiscovery {
 		status.Start("Deploying multi cluster service discovery")
 		err = lighthouse.Ensure(status, config, "", "", false, kubeConfig, kubeContext)
-		status.End(err == nil)
+		status.End(0)
```
mangelajo, line 215: ditto

pkg/subctl/cmd/join.go
```diff
@@ -206,17 +206,17 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 		exitOnError("Error deploying multi cluster service discovery", err)
 	}
 
-	fmt.Printf("* Discovering network details\n")
+	status.Start("* Discovering network details\n")
```
mangelajo, line 219:
```suggestion
	status.Start("Discovering network details\n")
```

pkg/subctl/cmd/join.go
```diff
@@ -231,10 +231,10 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	if subctlData.GlobalnetCidrRange == "" {
 		// Globalnet not enabled
 		err = checkOverlappingServiceCidr(globalNetworks)
-		status.End(err == nil)
+		status.End(0)
```
mangelajo, line 234: ditto

pkg/subctl/cmd/join.go
```diff
@@ -231,10 +231,10 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 		exitOnError("Error validating overlapping ServiceCIDRs", err)
 		err = checkOverlappingClusterCidr(globalNetworks)
-		status.End(err == nil)
+		status.End(0)
```
mangelajo, line 237: ditto

pkg/subctl/cmd/join.go
```diff
@@ -245,7 +245,7 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 		} else {
 			// no globalCidr configured on this cluster
 			globalnetCIDR, err = globalnet.AllocateGlobalCIDR(globalNetworks, subctlData)
-			status.End(err == nil)
+			status.End(0)
```
mangelajo, line 248: ditto

pkg/subctl/cmd/join.go
```diff
@@ -267,12 +267,12 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	brokerAdminClientset, err := kubernetes.NewForConfig(subctlData.GetBrokerAdministratorConfig())
 	exitOnError("Error retrieving broker admin config", err)
 	clienttoken, err = broker.CreateSAForCluster(brokerAdminClientset, clusterID)
-	status.End(err == nil)
+	status.End(0)
```
mangelajo, line 270: ditto

pkg/subctl/cmd/join.go
```diff
@@ -267,12 +267,12 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
 	exitOnError("Error creating SA for cluster", err)
 
 	status.Start("Deploying Submariner")
 	err = submarinercr.Ensure(config, OperatorNamespace, populateSubmarinerSpec(subctlData))
-	status.End(err == nil)
+	status.End(0)
```
mangelajo, line 275: ditto

pkg/subctl/cmd/join.go
```diff
@@ -335,7 +335,7 @@ func getNetworkDetails(config *rest.Config) *network.ClusterNetwork {
 	exitOnError("Unable to set the Kubernetes cluster connection up", err)
 	networkDetails, err := network.Discover(dynClient, clientSet)
 	if err != nil {
-		fmt.Printf("Error trying to discover network details: %s\n", err)
+		exitOnError("Error trying to discover network details: %s\n", err)
```
mangelajo, line 338: This shouldn't be a catastrophic failure, this will be asked to the admin with a prompt.

pkg/internal/cli/status.go
mangelajo: Use the symbols you defined :)

pkg/internal/cli/status.go
mangelajo: The UTF symbol for warning is cool :) , those sites are great for copy&paste: https://www.utf8icons.com/character/9888/warning-sign

pkg/subctl/cmd/join.go
skitt, line 219:
```suggestion
	status.Start("Discovering network details")
```

pkg/subctl/cmd/join.go
```diff
@@ -345,8 +345,8 @@ func getNetworkDetails(config *rest.Config) *network.ClusterNetwork {
 func getPodCIDR(clusterCIDR string, nd *network.ClusterNetwork) (string, error) {
 	if clusterCIDR != "" {
 		if nd != nil && len(nd.PodCIDRs) > 0 && nd.PodCIDRs[0] != clusterCIDR {
-			fmt.Printf("WARNING: your provided cluster CIDR for the pods (%s) does not match discovered (%s)\n",
-				clusterCIDR, nd.PodCIDRs[0])
+			status.QueueWarningMessage(fmt.Sprintf("WARNING: your provided cluster CIDR for the pods (%s) does not match discovered (%s)\n",
```
skitt, line 348:
```suggestion
			status.QueueWarningMessage(fmt.Sprintf("your provided cluster CIDR for the pods (%s) does not match discovered (%s)\n",
```

pkg/subctl/cmd/join.go
```diff
@@ -359,8 +359,8 @@ func getPodCIDR(clusterCIDR string, nd *network.ClusterNetwork) (string, error)
 func getServiceCIDR(serviceCIDR string, nd *network.ClusterNetwork) (string, error) {
 	if serviceCIDR != "" {
 		if nd != nil && len(nd.ServiceCIDRs) > 0 && nd.ServiceCIDRs[0] != serviceCIDR {
-			fmt.Printf("WARNING: your provided service CIDR (%s) does not match discovered (%s)\n",
-				serviceCIDR, nd.ServiceCIDRs[0])
+			status.QueueWarningMessage(fmt.Sprintf("WARNING: your provided service CIDR (%s) does not match discovered (%s)\n",
```
skitt, line 362:
```suggestion
			status.QueueWarningMessage(fmt.Sprintf("your provided service CIDR (%s) does not match discovered (%s)\n",
```

mangelajo:
@Jaanki can you handle comments and rebase/fix conflicts on top of master? CI is fixed, but you will need to rebase.

pkg/internal/cli/status.go
Jaanki: DOne

pkg/subctl/cmd/join.go
Jaanki: DOne

pkg/subctl/cmd/join.go
Jaanki: Done

pkg/subctl/cmd/join.go
Jaanki: Done

pkg/subctl/cmd/join.go
Jaanki: Isn't getting network details basic requirement? I think it should fail otherwise Submariner might not work

pkg/discovery/globalnet/globalnet.go
```diff
@@ -284,38 +284,40 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
...
+			status.QueueWarningMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cidr")
 			globalnetCIDR = ""
 		} else if globalnetClusterSize != 0 {
-			status.QueueSuccessMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cluster-size")
+			status.QueueWarningMessage("globalnet is not enabled on Broker. Ignoring specified globalnet-cluster-size")
 			globalnetClusterSize = 0
 		}
 	}
-	status.End(true)
+	status.End(cli.CheckStatus(err))
```
Jaanki, line 320: @mangelajo How about like this?

pkg/subctl/cmd/deploybroker.go
Jaanki: err == nil is true which is success bool. and here also 0 is success.

skitt: `err == nil` is `true` only if `err` is `nil`; you’d need something like `err == nil ? success : failure` (but as a function, since Go doesn’t have `? :`).

pkg/subctl/cmd/join.go
skitt: If we can’t retrieve the network details automatically, `subctl join` will ask the administrator for the relevant info (see `getServiceCIDR` etc.). If we exit here, the program won’t get a chance to do so.

pkg/discovery/globalnet/globalnet.go
Jaanki: @skitt like this?

pkg/internal/cli/status.go
```diff
@@ -125,3 +145,17 @@ func (s *Status) QueueSuccessMessage(message string) {
...
 	s.failureQueue = append(s.failureQueue, message)
 }
+
+// QueuewarningMessage queues up a message, which will be displayed once
+// the status ends (using the warning format)
+func (s *Status) QueueWarningMessage(message string) {
+	s.warningQueue = append(s.warningQueue, message)
+}
+
+func CheckStatus(err error) result {
```
Jaanki, line 155: @mangelajo @skitt and its implementation

pkg/discovery/globalnet/globalnet.go
skitt: Nearly ;-). Your `CheckStatus` is fine, but our comments only concerned the replacement of `status.End(err == nil)`. Here you’re replacing `status.End(true)`; the replacement for that should be `status.End(success)` directly.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -284,38 +284,40 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
 
 func ValidateGlobalnetConfiguration(subctlData *datafile.SubctlData, netconfig Config) (error, string, uint) {
 	status.Start("Validating Globalnet configurations")
+	var err error
+	var clusterSize uint
```
skitt, line 288:
```suggestion
```

The variables don’t need to be defined globally.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -284,38 +284,40 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
...
 	globalnetClusterSize := netconfig.GlobalnetClusterSize
 	globalnetCIDR := netconfig.GlobalnetCIDR
 	if subctlData.GlobalnetCidrRange != "" && globalnetClusterSize != 0 && globalnetClusterSize != subctlData.GlobalnetClusterSize {
-		clusterSize, err := GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
+		clusterSize, err = GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
```
skitt, line 292:
```suggestion
		clusterSize, err := GetValidClusterSize(subctlData.GlobalnetCidrRange, globalnetClusterSize)
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -284,38 +284,40 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
...
 		if err != nil || clusterSize == 0 {
 			return fmt.Errorf("Invalid globalnet-cluster-size %s", err), "", 0
 		}
 		subctlData.GlobalnetClusterSize = clusterSize
 	}
 
 	if globalnetCIDR != "" && globalnetClusterSize != 0 {
-		err := errors.New("Both globalnet-cluster-size and globalnet-cidr can't be specified. Specify either one.\n")
+		err = errors.New("Both globalnet-cluster-size and globalnet-cidr can't be specified. Specify either one.\n")
```
skitt, line 300:
```suggestion
		err := errors.New("Both globalnet-cluster-size and globalnet-cidr can't be specified. Specify either one.\n")
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -284,38 +284,40 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
...
 		return fmt.Errorf("%s", err), "", 0
 	}
 
 	if globalnetCIDR != "" {
-		_, _, err := net.ParseCIDR(globalnetCIDR)
+		_, _, err = net.ParseCIDR(globalnetCIDR)
```
skitt, line 305:
```suggestion
		_, _, err := net.ParseCIDR(globalnetCIDR)
```

pkg/discovery/globalnet/globalnet.go
skitt, line 320:
```suggestion
	status.End(status.Success)
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -362,17 +365,17 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
...
-			err := CheckOverlappingCidr(globalNetworks, global, netconfig)
+			err = CheckOverlappingCidr(globalNetworks, global, netconfig)
 			if err != nil {
 				return "", fmt.Errorf("Error validating overlapping GlobalCIDRs %s %s", globalnetCIDR, err)
 			}
 			status.QueueSuccessMessage(fmt.Sprintf("GlobalCIDR is: %s", globalnetCIDR))
 		}
 	}
-	status.End(true)
+	status.End(cli.CheckStatus(err))
```
skitt, line 378:
```suggestion
	status.End(status.Success)
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -341,17 +343,18 @@ func GetGlobalNetworks(subctlData *datafile.SubctlData) (map[string]*GlobalNetwo
 
 func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[string]*GlobalNetwork, netconfig Config) (string, error) {
 	status.Start("Assigning Globalnet IPs")
+	var err error
```
skitt, line 346:
```suggestion
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -341,17 +343,18 @@ func GetGlobalNetworks(subctlData *datafile.SubctlData) (map[string]*GlobalNetwo
...
 		// Globalnet enabled, GlobalCIDR not specified by the user
 		if isCIDRPreConfigured(clusterID, globalNetworks) {
 			// globalCidr already configured on this cluster
 			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-			status.QueueSuccessMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
+			status.QueueWarningMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
 		} else {
 			// no globalCidr configured on this cluster
-			globalnetCIDR, err := AllocateGlobalCIDR(globalNetworks, subctlData)
+			globalnetCIDR, err = AllocateGlobalCIDR(globalNetworks, subctlData)
```
skitt, line 357:
```suggestion
			globalnetCIDR, err := AllocateGlobalCIDR(globalNetworks, subctlData)
```

pkg/discovery/globalnet/globalnet.go
```diff
@@ -362,17 +365,17 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
 		if isCIDRPreConfigured(clusterID, globalNetworks) {
 			// globalCidr pre-configured on this cluster
 			globalnetCIDR = globalNetworks[clusterID].GlobalCIDRs[0]
-			status.QueueSuccessMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
+			status.QueueWarningMessage(fmt.Sprintf("Pre-configured GlobalCIDR %s detected. Not changing it.", globalnetCIDR))
 		} else {
 			// globalCidr as specified by the user
-			err := CheckOverlappingCidr(globalNetworks, global, netconfig)
+			err = CheckOverlappingCidr(globalNetworks, global, netconfig)
```
skitt, line 371:
```suggestion
			err := CheckOverlappingCidr(globalNetworks, global, netconfig)
```

pkg/internal/cli/status.go
```diff
@@ -74,7 +87,7 @@ func StatusForLogger(l log.Logger) *Status {
 // Start starts a new phase of the status, if attached to a terminal
 // there will be a loading spinner with this status
 func (s *Status) Start(status string) {
-	s.End(true)
+	s.End(success)
```
skitt, line 90:
```suggestion
	s.End(Success)
```

pkg/internal/cli/status.go
```diff
@@ -96,10 +109,13 @@ func (s *Status) End(success bool) {
 		s.spinner.Stop()
 		fmt.Fprint(s.spinner.writer, "\r")
 	}
-	if success {
+	switch output {
+	case success:
```
skitt, line 113:
```suggestion
	case Success:
```

pkg/internal/cli/status.go
```diff
@@ -96,10 +109,13 @@ func (s *Status) End(success bool) {
...
 		s.logger.V(0).Infof(s.successFormat, s.status)
-	} else {
+	case failure:
```
skitt, line 115:
```suggestion
	case Failure:
```

pkg/internal/cli/status.go
```diff
@@ -96,10 +109,13 @@ func (s *Status) End(success bool) {
...
 		s.logger.V(0).Infof(s.failureFormat, s.status)
+	case warning:
```
skitt, line 117:
```suggestion
	case Warning:
```

pkg/internal/cli/status.go
```diff
@@ -125,3 +145,17 @@ func (s *Status) QueueSuccessMessage(message string) {
...
+
+// QueuewarningMessage queues up a message, which will be displayed once
+// the status ends (using the warning format)
+func (s *Status) QueueWarningMessage(message string) {
+	s.warningQueue = append(s.warningQueue, message)
+}
+
+func CheckStatus(err error) result {
+	if err == nil {
+		return success
```
skitt, line 157:
```suggestion
		return Success
```

pkg/internal/cli/status.go
```diff
@@ -125,3 +145,17 @@ func (s *Status) QueueSuccessMessage(message string) {
...
+	} else {
+		return failure
```
skitt, line 159:
```suggestion
		return Failure
```

pkg/internal/cli/status.go
```diff
@@ -25,6 +25,14 @@ import (
 	"github.com/submariner-io/submariner-operator/pkg/internal/log"
 )
 
+type result int
+
+const (
+	success result = iota
+	failure
+	warning
```
skitt, line 33:
```suggestion
	Success result = iota
	Failure
	Warning
```

pkg/internal/cli/status.go
Jaanki: DOne

pkg/subctl/cmd/join.go
Jaanki: Using warning message instead.

pkg/discovery/globalnet/globalnet.go
Jaanki: How would status.End(0) (original implementation) different from status.End(success)? Just trying to understand. After all, its the same value in form of iota.

skitt: From the computer’s perspective, the result is the same, as you say. However from a developer’s perspective it isn’t: if you read `status.End(0)`, you can’t know what it means without looking at the implementation of `status.End()`, which I’d rather avoid; if you read `status.End(status.Success)`, it’s somewhat easier to understand. Using symbolic names also allows us in the future to switch to a specific type, instead of generic `int`s, and thus constrain the allowable values. (As it is, one could write `success.End(4)` and the compiler wouldn’t complain.)

pkg/discovery/globalnet/globalnet.go
Jaanki: I created those for CheckStatus(err) which is not required now.

pkg/discovery/globalnet/globalnet.go
Jaanki: DOne

pkg/discovery/globalnet/globalnet.go
Jaanki: Done

pkg/discovery/globalnet/globalnet.go
Jaanki: It would be cli.Success. DOne

pkg/discovery/globalnet/globalnet.go
Jaanki: DOne

pkg/discovery/globalnet/globalnet.go
Jaanki: DOne

pkg/discovery/globalnet/globalnet.go
Jaanki: DOne

pkg/discovery/globalnet/globalnet.go
Jaanki: Done

pkg/internal/cli/status.go
Jaanki: Done

pkg/internal/cli/status.go
Jaanki: Done

pkg/internal/cli/status.go
Jaanki: Done

pkg/internal/cli/status.go
Jaanki: Done

pkg/internal/cli/status.go
Jaanki: Removed this. Not needed now IMO

pkg/internal/cli/status.go
Jaanki: Done

pkg/discovery/globalnet/globalnet.go
Jaanki: Done

pkg/subctl/cmd/deploybroker.go
```diff
@@ -83,7 +83,7 @@ var deployBroker = &cobra.Command{
 		status := cli.NewStatus()
 		status.Start("Deploying broker")
 		err = broker.Ensure(config)
-		status.End(err == nil)
+		status.End(cli.Success)
```
skitt, line 86: Sorry, I didn’t make this clear. Whenever the original code has a condition on `err`, you need to preserve that, with the conversion function you had in the previous iteration. So `status.end(true)` becomes `status.end(cli.Success)`, `status.end(false)` becomes `status.end(cli.Failure)`, and `status.end(err == nil)` becomes `status.end(cli.CheckStatus(err))`. The idea is to use `err` as an indicator of success: if, when we call `status.End()`, `err` is `nil`, that indicates that the process succeeded, otherwise it failed.

Jaanki: Isn't 'err == nil' true ? Also status.end doesn't do anything functionally except for notifying status to print message according to status.QueueSuccessMessage or status.QueueFailureMessage. Even if I am calling CheckStatus(err) what would it return? true or false?? Which again translates to success or failure.

skitt: `err == nil` is true if `err` is `nil`, false otherwise. The idea here is that if `broker.Ensure(config)` succeeds, it will return no error, and `status.End(true)` is called to indicate success; if it fails, it will return an error, and `status.End(false)` is called to indicate failure.

`CheckStatus(err)` should return `cli.Success` is `err` is `nil`, `cli.Failure` otherwise.

skitt: Put another way,

    status.End(err == nil)

is equivalent to

    if (err == nil) {
        status.End(true)
    } else {
        status.End(false)
    }

Jaanki: Got it

pkg/subctl/cmd/deploybroker.go
Jaanki: Done

pkg/subctl/cmd/deploybroker.go
Jaanki: Got it. THanks :)

PR #375, opened 2020-05-13
Organize embedded yamls

mangelajo:
Move submariner (non-operator) specific yamls into it's own folder,
remove namespace.yaml which is not used, and it's actually json.


mangelajo:
Ok, rebased on top of CI fix by @mkolesnik hopefully back on track ':D

PR #380, opened 2020-05-14
Add ServiceExport CRD

mangelajo:
Adds the ServiceExport CRD for lighthouse U/S alignment, only
on the clusters where the operator is installed, not the broker.

Related-Issue: https://github.com/submariner-io/lighthouse/issues/78

### DIFF
--- a/deploy/lighthouse/cluster_role.yaml
+++ b/deploy/lighthouse/cluster_role.yaml
@@ -17,7 +17,7 @@ rules:
 - apiGroups:
   - lighthouse.submariner.io
   resources:
-  - multiclusterservices
+  - "*"
   verbs:
   - create
   - get
--- a/deploy/lighthouse/crds/serviceexport_crd.yaml
+++ b/deploy/lighthouse/crds/serviceexport_crd.yaml
@@ -1,0 +1,12 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: serviceexports.lighthouse.submariner.io
+spec:
+  group: lighthouse.submariner.io
+  version: v2alpha1
+  names:
+    kind: ServiceExport
+    plural: serviceexports
+    singular: serviceexport
+  scope: Namespaced
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -38,7 +38,7 @@ func Ensure(config *rest.Config) error {
 		return fmt.Errorf("error setting up the engine requirements: %s", err)
 	}
 
-	err = lighthouse.Ensure(config)
+	_, err = lighthouse.Ensure(config, lighthouse.BrokerCluster)
 	if err != nil {
 		return fmt.Errorf("error setting up the lighthouse requirements: %s", err)
 	}
--- a/pkg/lighthouse/crds.go
+++ b/pkg/lighthouse/crds.go
@@ -3,37 +3,42 @@ package lighthouse
 import (
 	"fmt"
 
-	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/common/embeddedyamls"
-	"github.com/submariner-io/submariner-operator/pkg/utils"
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/rest"
+
+	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/common/embeddedyamls"
+	"github.com/submariner-io/submariner-operator/pkg/utils"
+)
+
+const (
+	BrokerCluster = true
+	DataCluster   = false
 )
 
 // Ensure ensures that the required resources are deployed on the target system
-// The resources handled here are the lighthouse CRDs: MultiClusterService
-func Ensure(config *rest.Config) error {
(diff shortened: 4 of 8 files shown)


pkg/broker/ensure.go
```diff
@@ -38,7 +38,7 @@ func Ensure(config *rest.Config) error {
 		return fmt.Errorf("error setting up the engine requirements: %s", err)
 	}
 
-	err = lighthouse.Ensure(config)
+	_, err = lighthouse.EnsureMcsCRD(config)
```
skitt, line 41: That’s a shame, I thought the whole point of the `Ensure` pattern was that it could be blindly used (largely)...

mangelajo: yeah, but... we make a difference between what we "ensure" on the broker, and what we ensure on the operator... I needed a difference for that... better ideas are welcome.

mangelajo: We could deploy all everywhere though, and not care. Which is what we would have in the case of the olm, right?

vthapar: We can use isBroker flag to differentiate betwen data and broker. We may not need ServiceExport CR on broker, but no harm in installing CRD everywhere.

mangelajo: It's what will happen when we people setups the broker through olm...

pkg/subctl/operator/lighthouse/crds/ensure.go
```diff
@@ -17,51 +17,37 @@ limitations under the License.
...
-		return false, err
-	}
-
-	crd, err := getServiceDiscoveryCRD()
-	if err != nil {
-		return false, err
-	}
-
-	serviceDiscoveryResult, err := utils.CreateOrUpdateCRD(clientSet, crd)
+	serviceDiscoveryResult, err := EnsureServiceDiscoveryCRD(restConfig)
```
mangelajo, line 33:
```suggestion
	serviceDiscoveryResult, err := ensureServiceDiscoveryCRD(restConfig)
```

pkg/subctl/operator/lighthouse/crds/ensure.go
```diff
@@ -17,51 +17,37 @@ limitations under the License.
...
-	return (serviceDiscoveryResult || mcsCRDResult), err
+	return (serviceDiscoveryResult || installed), err
 }
 
-func getServiceDiscoveryCRD() (*apiextensionsv1beta1.CustomResourceDefinition, error) {
-	crd := &apiextensionsv1beta1.CustomResourceDefinition{}
-
-	if err := embeddedyamls.GetObject(embeddedyamls.Crds_submariner_io_servicediscoveries_crd_yaml, crd); err != nil {
-		return nil, err
+func EnsureServiceDiscoveryCRD(config *rest.Config) (bool, error) {
```
mangelajo, line 46:
```suggestion
func ensureServiceDiscoveryCRD(config *rest.Config) (bool, error) {
```

deploy/lighthouse/crds/serviceexport_crd.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
skitt, line 1:
```suggestion
apiVersion: apiextensions.k8s.io/v1
```

mangelajo:
This will need a **SQUASH**

pkg/lighthouse/crds.go
```diff
@@ -3,37 +3,42 @@ package lighthouse
...
+	"github.com/submariner-io/submariner-operator/pkg/utils"
 )
 
-// Ensure ensures that the required resources are deployed on the target system
+const (
+	BrokerCluster = true
+	DataCluster   = false
+)
+
+// EnsureMcsCRD ensures that the required resources are deployed on the target system
```
skitt, line 18:
```suggestion
// Ensure ensures that the required resources are deployed on the target system
```

pkg/lighthouse/crds.go
```diff
@@ -3,37 +3,42 @@ package lighthouse
...
 // The resources handled here are the lighthouse CRDs: MultiClusterService
```
skitt, line 19:
```suggestion
// The resources handled here are the lighthouse CRDs: MultiClusterService and ServiceExport
```

mangelajo:
> Thanks for reverting the Ensure changes!
Thank you for noticing that!!!

And thanks for the review in general.

deploy/lighthouse/cluster_role.yaml
```diff
@@ -17,7 +17,7 @@ rules:
 - apiGroups:
   - lighthouse.submariner.io
   resources:
-  - multiclusterservices
+  - "*"
```
vthapar, line 20: Should change this in broker/rbac.go also. It will be useful when we add serviceImport and endpointslice as those will exist on broker too.

deploy/lighthouse/crds/serviceexport_crd.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apiextensions.k8s.io/v1
```
vthapar, line 1: Need to use v1beta1. Refer: https://github.com/submariner-io/submariner-charts/pull/35

mangelajo: Ack thanks

deploy/lighthouse/cluster_role.yaml
mangelajo: ack.

PR #387, opened 2020-05-18
Ensure that files/directories don’t hide Dapper targets

skitt:
Targets offered by the in-Dapper Makefile are hidden by files or
directories of the same name, for example “deploy”. This patch marks
all such files/directories, apart from those starting with “M” (to
avoid problems with the Makefiles themselves), as phony, with an
explicit build target.

This allows

	make deploy

to work even though there’s a deploy directory.

### DIFF
--- a/Makefile.dapper
+++ b/Makefile.dapper
@@ -10,10 +10,16 @@
 	@./.dapper.tmp -v
 	@mv .dapper.tmp .dapper
 
+invoke_dapper = ./.dapper -m bind make $1 $(MAKEFLAGS)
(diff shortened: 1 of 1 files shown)


mkolesnik:
@skitt don't we need the same in other repos (shipyard, subm, etc)?

skitt:
> @skitt don't we need the same in other repos (shipyard, subm, etc)?

@mkolesnik yes, but this is incomplete; I’ll fix it here and then update the other repos with the complete version.

PR #388, opened 2020-05-19
Highlight warning message format

Jaanki:
Warning sign doesn't pop out against black
background and green fonts. Highlight the
sign by adding "WARNING" text next to it.

Screenshot from 2020-05-19 10-42-41

### DIFF
--- a/pkg/internal/cli/status.go
+++ b/pkg/internal/cli/status.go
@@ -65,7 +65,7 @@ func StatusForLogger(l log.Logger) *Status {
 		logger:        l,
 		successFormat: " ✓ %s\n",
 		failureFormat: " ✗ %s\n",
-		warningFormat: " ⚠ %s\n",
+		warningFormat: " ⚠ WARNING %s\n",
 		successQueue:  []string{},
 		failureQueue:  []string{},
 		warningQueue:  []string{},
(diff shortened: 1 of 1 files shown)


mangelajo:
Approved, so we can merge, but left another idea for improvement if you think it can be done.

skitt:
> A slightly more complicated idea, but do you think we could print the whole line in color, and then terminate with : \x1b[0m so we stop coloring :)

That’s easy enough to do, the format string describes the whole line. If we do this, I think we should also adjust the error format accordingly.

However, red tints tend to be poorly readable, so it seems preferable to me to limit the colour change to the “WARNING” and “ERROR” labels, and not the message itself. (Having an explicit label is good for people with colour perception problems.)

pkg/internal/cli/status.go
```diff
@@ -65,7 +65,7 @@ func StatusForLogger(l log.Logger) *Status {
 		logger:        l,
 		successFormat: " ✓ %s\n",
 		failureFormat: " ✗ %s\n",
```
skitt, line 67: Could we do this for error messages too?

skitt: In a separate PR...

Jaanki:
I agree with changing the clolour only till "WARNING" The whole line would be too bright/catchy.
Also on failure, we are anyways using exitOnError which solves the purpose of getting user attention so changing colour for failure is reduntant.

PR #404, opened 2020-05-22
Fix globalCIDR auto-allocation error

Jaanki:
There is an error where globalCIDR is not configured on the cluster
when not specified by the user. This patch solves that issue.


mangelajo:
CI is broken, let's rebase once the ServiceExport patch merges.

pkg/discovery/globalnet/globalnet.go
```diff
@@ -351,7 +352,7 @@ func AssignGlobalnetIPs(subctlData *datafile.SubctlData, globalNetworks map[stri
 			status.QueueWarningMessage(fmt.Sprintf("Cluster already has GlobalCIDR allocated: %s", globalNetworks[clusterID].GlobalCIDRs[0]))
 		} else {
 			// no globalCidr configured on this cluster
-			globalnetCIDR, err := AllocateGlobalCIDR(globalNetworks, subctlData)
+			globalnetCIDR, err = AllocateGlobalCIDR(globalNetworks, subctlData)
```
sridhargaddam, line 355: lol... for some time, I was wondering what is the change in the PR :)

sridhargaddam:
> Can you add a test to make sure this doesnt happen in the future?

Discussed with @mkolesnik over slack, we will be adding a test-case in a separate PR as @Jaanki is on PTO.

PR #405, opened 2020-05-22
Remove verify_subm_operator_container , fix CI

mangelajo:
The submariner operator now it's a scratch binary that does not contain any
commands, we cannot perform those verifications through exec. Also those
are details that don't really matter as long as the operator works, which
will be verified by the E2E tests passing.


mangelajo:
This is @skitt patch I didn't find, merged that one, which also fixes the picking up of the right image.

PR #407, opened 2020-05-22
Deploy the ServiceExport in the lighthouse case

mangelajo:
Otherwise the dns query won't work because lighthouse
now requires a ServiceExport.

Fixes-Issue: https://github.com/submariner-io/shipyard/issues/182

### DIFF
--- a/scripts/kind-e2e/e2e.sh
+++ b/scripts/kind-e2e/e2e.sh
@@ -49,8 +49,9 @@ run_subm_clusters verify_subm_deployed
 echo "Running subctl a second time to verify if running subctl a second time works fine"
 with_context cluster3 subctl_install_subm
 
-with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
 with_context cluster3 deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+[[ "${lighthouse}" = "true" ]] && with_context cluster3 kubectl apply -f "${DAPPER_SOURCE}/scripts/kind-e2e/nginx-demo-export.yaml"
+with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
 
 with_context cluster2 connectivity_tests
 
--- a/scripts/kind-e2e/nginx-demo-export.yaml
+++ b/scripts/kind-e2e/nginx-demo-export.yaml
@@ -1,0 +1,5 @@
+---
+apiVersion: lighthouse.submariner.io/v2alpha1
+kind: ServiceExport
+metadata:
+  name: nginx-demo


scripts/kind-e2e/e2e.sh
```diff
@@ -51,6 +51,7 @@ with_context cluster3 subctl_install_subm
 
 with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
 with_context cluster3 deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+[[ "${lighthouse}" = "true" ]] && with_context cluster3 deploy_resource "${DAPPER_SOURCE}/scripts/kind-e2e/nginx-demo-export.yaml"
```
vthapar, line 54: You may have to add a sleep before testing discovery in case it tries and gives up before cr can be distributed to clusters.

mangelajo: I wonder how that wasn't failing before... :? , but you are right

scripts/kind-e2e/e2e.sh
```diff
@@ -51,6 +51,7 @@ with_context cluster3 subctl_install_subm
 
 with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
 with_context cluster3 deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+[[ "${lighthouse}" = "true" ]] && with_context cluster3 kubectl apply -f "${DAPPER_SOURCE}/scripts/kind-e2e/nginx-demo-export.yaml"
```
mangelajo, line 54: used kubectl apply -f because deploy_resource will check for rollout status... and we don't have that on service export.

scripts/kind-e2e/e2e.sh
```diff
@@ -49,8 +49,9 @@ run_subm_clusters verify_subm_deployed
 echo "Running subctl a second time to verify if running subctl a second time works fine"
 with_context cluster3 subctl_install_subm
 
-with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
 with_context cluster3 deploy_resource "${RESOURCES_DIR}/nginx-demo.yaml"
+[[ "${lighthouse}" = "true" ]] && with_context cluster3 kubectl apply -f "${DAPPER_SOURCE}/scripts/kind-e2e/nginx-demo-export.yaml"
+with_context cluster2 deploy_resource "${RESOURCES_DIR}/netshoot.yaml"
```
mangelajo, line 54: This line is moved after the nginx part, to leave some extra headroom to the service export to happen while we wait for netshoot to rollout.

PR #410, opened 2020-05-24
Refactor connectivity_tests to service_discovery_tests

mkolesnik:
These are more correctly encapsulated now inside the function.
Also added a few retries to the test itself since it might fail due to
timing issues.


mkolesnik:
This has been removed since, closing the PR

PR #412, opened 2020-05-25
Pull the build scripts into the Makefile

skitt:
... and add a clean target, and fix the Windows builds to include the
extension.


.github/workflows/validations.yaml
```diff
@@ -18,4 +18,4 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - run: make generate-embeddedyamls ${{ matrix.command }}
+      - run: make pkg/subctl/operator/common/embeddedyamls/yamls.go ${{ matrix.command }}
```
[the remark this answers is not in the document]
skitt: Done, I’ve restored the `generate-embeddedyamls` target (as a phony).

Makefile
```diff
@@ -23,14 +29,34 @@ e2e: deploy
...
-build-subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
+clean:
+	rm -f $(BINARIES)
 
-bin/subctl: build-subctl
+build: build-operator binaries
+
+binaries: $(BINARIES)
+
+build-operator: vendor/modules.txt
```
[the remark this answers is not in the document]
skitt: Done.

Makefile
```diff
@@ -8,12 +8,18 @@ include $(SHIPYARD_DIR)/Makefile.inc
 
 VERSION := $(shell . scripts/lib/version; echo $$VERSION)
```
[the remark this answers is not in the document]
skitt: Done. (I’ll leave cleanup of the other users for a later PR.)

Makefile
```diff
@@ -23,14 +29,38 @@ e2e: deploy
 $(TARGETS): vendor/modules.txt
 	./scripts/$@
 
-build-subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
+clean:
+	rm -f $(BINARIES)
 
-bin/subctl: build-subctl
+build: operator-image binaries
```
mkolesnik, line 35: Yeah so I missed this before, but I think this is a major change.

Now everything relying on `build` would always run the cross compile..

In actuality it's not really needed for the e2e (just the one subctl of the current os/arch) but adds a few minutes to each job.

I'd rather we not have build-cross as default, but as a distinct target as it used to be, or make the e2e chain not rely on it all but just the subctl

skitt: Fixed, it’s back to `build-cross`.

Makefile
```diff
@@ -23,14 +30,40 @@ e2e: deploy
...
+
+bin/subctl-%: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go") vendor/modules.txt
+	version=$(word 2,$(subst -, ,$@)); \
+	GOOS=$(word 3,$(subst -, ,$@)); \
+	GOARCH=$(word 1,$(subst ., ,$(word 4,$(subst -, ,$@)))); \
+	GOEXE=$(word 2,$(subst ., ,$(word 4,$(subst -, ,$@)))); \
+	[ "$$GOOS" != "darwin" ] || upxflag="--noupx"; \
+	export GOARCH GOEXE GOOS; \
+	$(SCRIPTS_DIR)/compile.sh \
+		--ldflags "-X github.com/submariner-io/submariner-operator/pkg/version.Version=$${version} -X github.com/submariner-io/submariner-operator/pkg/version.GitRef=$(GIT_REF)" \
```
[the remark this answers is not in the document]
skitt: The idea is to explicitly use the version extracted from the target name, without forcing it to be equal to `$(VERSION)`. Think of this as a function which is only supposed to act on its arguments, not on a global variable ;-).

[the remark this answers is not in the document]
mkolesnik: Ok sure, BTW we need to make sure VERSION has no dashes or this'll break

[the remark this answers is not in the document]
skitt: We already rely on that, the version is embedded in the current binary and tarball names.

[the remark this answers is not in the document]
mkolesnik: Yeah but so far we didnt try to parse the tile name, now with 
```
 	version=$(word 2,$(subst -, ,$@)); \ 

	GOOS=$(word 3,$(subst -, ,$@)); \
	GOARCH=$(word 1,$(subst ., ,$(word 4,$(subst -, ,$@)))); \
	GOEXE=$(word 2,$(subst ., ,$(word 4,$(subst -, ,$@)))); \ 
```
You're really relying on the filename, and hence if for example version was `foo-bar` these will break as the calculations will take `foo` as the 2nd word and `bar` as the 3rd (and skew everything else, of course).

Actually now that I think of it, it'll  break right now since our calculation is a tag or `dev` buuut if you tag say `v0.4.0-rc1` (which we did and do) then these will break.

Or am I missing something?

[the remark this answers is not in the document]
skitt: Oh I read slashes, not dashes, duh. You’re right, I need to find a better way to do this...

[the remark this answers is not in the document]
skitt: I’ve updated the patch to rely on `$(VERSION)` and count backwards when splitting.

Makefile
```diff
@@ -30,14 +39,44 @@ e2e: deploy
...
+dist/subctl-%.tar.xz: bin/subctl-%
+	mkdir -p dist
+	tar -cJf $@ --transform "s/^bin/subctl-$(VERSION)/" $<
+
+# Versions may include hyphens so it's easier to use $(VERSION) than to extract them from the target
+bin/subctl-%: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go") vendor/modules.txt
+	mkdir -p bin
+	GOOS=$(word $(guile (- $(words $(subst -, ,$@)) 1)),$(subst -, ,$@)); \
+	GOARCH=$(word 1,$(subst ., ,$(lastword $(subst -, ,$@)))); \
+	GOEXE=$(suffix $@); \
```
mkolesnik, line 66: Turns out GOEXE is not settable, and is determined and reported by Go.
Hence, there's nothing to gain by setting it (funnily, it wasn't even used in the original `build-cross` script)

Makefile
```diff
@@ -30,14 +39,44 @@ e2e: deploy
...
+	export GOARCH GOEXE GOOS; \
+	$(SCRIPTS_DIR)/compile.sh \
+		--ldflags "-X github.com/submariner-io/submariner-operator/pkg/version.Version=$(VERSION) -X github.com/submariner-io/submariner-operator/pkg/version.GitRef=$(GIT_REF)" \
+		--noupx $@ ./pkg/subctl/main.go
+
+ci: generate-embeddedyamls validate test build
 
-bin/subctl: build-subctl
+generate-embeddedyamls: pkg/subctl/operator/common/embeddedyamls/yamls.go
```
mkolesnik, line 74: Could this be a Makefile variable instead? Then you could use it as a dependency of `bin/subctl-%: `

skitt: This is only here to allow other pieces of the builds to continue working as-is. It will be removed when `test` and `validate` are pulled into the Makefile too...

Makefile
```diff
@@ -30,14 +39,44 @@ e2e: deploy
...
+
+dist/subctl-%.tar.xz: bin/subctl-%
+	mkdir -p dist
+	tar -cJf $@ --transform "s/^bin/subctl-$(VERSION)/" $<
+
+# Versions may include hyphens so it's easier to use $(VERSION) than to extract them from the target
+bin/subctl-%: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go") vendor/modules.txt
+	mkdir -p bin
+	GOOS=$(word $(guile (- $(words $(subst -, ,$@)) 1)),$(subst -, ,$@)); \
+	GOARCH=$(word 1,$(subst ., ,$(lastword $(subst -, ,$@)))); \
```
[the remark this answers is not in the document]
skitt: The problem I have with the pattern-specific variable approach here is that adding or removing an architecture involves changes in a bunch of places. I’ve replaced the Make-wrenching syntax by shell parameter expansions, hopefully that will be more maintainable.

Makefile
skitt: Yes, I’ve removed that now.

Makefile
```diff
@@ -30,14 +39,46 @@ e2e: deploy
...
 
-build-subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
+clean:
+	rm -f $(BINARIES) $(CROSS_BINARIES) $(CROSS_TARBALLS)
+
+build: operator-image binaries
+
+build-cross: $(CROSS_TARBALLS)
+
+binaries: $(BINARIES)
```
mkolesnik, line 49: I'd skip this and just have `build` rely on `bin/subctl` unless you think someone would use `make binaries` :)

skitt: Indeed, done!

Makefile
```diff
@@ -30,14 +39,44 @@ e2e: deploy
 $(TARGETS): vendor/modules.txt
 	./scripts/$@
 
-build-subctl: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go")
+clean:
+	rm -f $(BINARIES) $(CROSS_BINARIES) $(CROSS_TARBALLS)
+
+build: operator-image $(BINARIES)
+
+build-cross: $(CROSS_TARBALLS)
```
mangelajo, line 47: hmmm, and where's the logic for this now?

mangelajo: oh, depends on the tarballs.. :)

tpantelis: In the `dist/subctl-%.tar.xz` target below.

mangelajo:
If I print $GOOS and $GOARCH, this is what I get: 

bin/subctl
v0.4

skitt:
> If I print $GOOS and $GOARCH, this is what I get:
> 
> bin/subctl
> v0.4

Yes, it’s stripping everything after the last . instead of removing .exe if any... So the patch works as long as the version is “dev”, but not with a numeric version. Fixed in my latest push.

Makefile
```diff
@@ -30,14 +39,44 @@ e2e: deploy
...
+bin/subctl-%: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg/subctl/ -name "*.go") vendor/modules.txt
+	mkdir -p bin
+	target=$@; \
+	target=$${target%.exe}; \
+	components=($$(echo $${target//-/ })); \
+	GOOS=$${components[-2]}; \
+	GOARCH=$${components[-1]}; \
+	export GOARCH GOOS; \
+	$(SCRIPTS_DIR)/compile.sh \
+		--ldflags "-X github.com/submariner-io/submariner-operator/pkg/version.Version=$(VERSION) -X github.com/submariner-io/submariner-operator/pkg/version.GitRef=$(GIT_REF)" \
```
mkolesnik, line 69: Tangentially related, we can stop using GIT_REF since this was merged https://github.com/submariner-io/shipyard/pull/219

skitt: Good point, https://github.com/submariner-io/submariner-operator/pull/507 fixes that (and if it’s merged before this, I’ll update this to suit, or vice versa as appropriate).

mangelajo:
Sorry I tried to merge during the github outage after testing locally, let me see if I can resolve the conflict.

PR #418, opened 2020-05-28
Stops packing subctl with upx and generates an tar.xz

mangelajo:
Under specific conditions (OSX, some types of mounts..) upx unpacking
of the executable generates a segmentation fault.

A decission was made to move into a tar packaged file, this will
also allow us to include more files in the future if necessary.

Extracting the files can be performed as:

```bash
$ tar xvf bin/subctl-dev-linux-amd64.tar.xz
subctl-dev-linux-amd64/subctl-dev-linux-amd64
```

Fixes-Issue: #335

### DIFF
--- a/scripts/build-subctl
+++ b/scripts/build-subctl
@@ -6,17 +6,21 @@ source ${SCRIPTS_DIR}/lib/version
 
 cd ${DAPPER_SOURCE}
 
+PACK_SUBCTL="${PACK_SUBCTL:-false}"
+
 echo Building subctl version $VERSION for ${GOOS:=$(go env GOOS)}/${GOARCH:=$(go env GOARCH)}
 GIT_REF="$(git describe --tags)"
 ldflags="-X github.com/submariner-io/submariner-operator/pkg/version.Version=${VERSION} "
 ldflags="${ldflags} -X github.com/submariner-io/submariner-operator/pkg/version.GitRef=${GIT_REF}"
 
-# OSX does not seem to like upx compressed from linux, so disabled
-[ "$GOOS" != "darwin" ] || upxflag="--noupx"
-
 output=bin/subctl-${VERSION}-${GOOS}-${GOARCH}${GOEXE}
-${SCRIPTS_DIR}/compile.sh --ldflags "$ldflags" $upxflag $output ./pkg/subctl/main.go "$@"
+
+${SCRIPTS_DIR}/compile.sh --ldflags "$ldflags" --noupx $output ./pkg/subctl/main.go "$@"
 
 if [ "${GOARCH}" = "$(GOARCH= go env GOARCH)" ] && [ "${GOOS}" = "$(GOOS= go env GOOS)" ]; then
     ln -sf ${output#bin/} bin/subctl
+fi
+
+if [ "${PACK_SUBCTL}" = "true" ]; then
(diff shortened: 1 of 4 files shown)


scripts/build-subctl
```diff
@@ -6,17 +6,22 @@ source ${SCRIPTS_DIR}/lib/version
...
-output=bin/subctl-${VERSION}-${GOOS}-${GOARCH}${GOEXE}
-${SCRIPTS_DIR}/compile.sh --ldflags "$ldflags" $upxflag $output ./pkg/subctl/main.go "$@"
+${SCRIPTS_DIR}/compile.sh --ldflags "$ldflags" --noupx $output ./pkg/subctl/main.go "$@"
 
 if [ "${GOARCH}" = "$(GOARCH= go env GOARCH)" ] && [ "${GOOS}" = "$(GOOS= go env GOOS)" ]; then
     ln -sf ${output#bin/} bin/subctl
 fi
+
+if [ "${PACK_SUBCTL}" = "true" ]; then
+  tar -cf - --transform "s/^bin/${binary_name}/" $output | xz -9 -c - > ${output}.tar.xz
```
skitt, line 26: It would be better to drop the version and `${GOEXE}` from the tarball’s name:

```suggestion
  tar -cJf subctl-${GOOS}-${GOARCH}.tar.xz --transform "s/^bin/subctl-${VERSION}/" ${output}
```

I’m not sure about keeping the version inside the tarball, but that matches what’s commonly done.

mangelajo: Hmm, I've always seen the version/etc in tarballs. we need to keep the os/arch at least otherwise they would all have the same name :-)

mangelajo: I did some random sampling trying to figure out if there was some de-facto standard, ... but there is not 50%/50%   ... 

**versions:** 

- https://github.com/operator-framework/operator-sdk/releases/download/v0.18.0/operator-sdk-v0.18.0-x86_64-linux-gnu
- https://github.com/kubernetes-sigs/kubefed/releases/download/v0.3.0/kubefedctl-0.3.0-linux-amd64.tgz
- https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_linux_amd64.tar.gz

**no versions:** 

- https://github.com/che-incubator/chectl/releases/download/20200527140404/chectl-linux-x64.tar.gz
- https://storage.googleapis.com/kubernetes-release/release/v1.18.0/bin/linux/amd64/kubectl
- https://github.com/kubernetes-sigs/kind/releases/download/v0.8.1/kind-linux-amd64


@skitt why do you think it's better to drop it?

I'm a bit paranoid sometimes, and I tend to store different versions of softwares I use in a folder, that saved me multiple times when a vendor stops releasing the old one, and the new one is buggy and stuff like that... having the version number in the package makes it easier.

not kidding:
image

skitt: OK, let’s go with

```suggestion
  tar -cJf subctl-${VERSION}-${GOOS}-${GOARCH}.tar.xz --transform "s/^bin/subctl-${VERSION}/" ${output}
```

mangelajo: thanks for considering it

mangelajo: I will download/check, probably we can remove the binary_name env var too then.

PR #423, opened 2020-05-29
Fix broken e2e and update dependencies

vthapar:
1. Update shipyard dependency.
    2. Use value returned by E2E to return pass/fail status of `verify`
    3. Refactor `verify` code
    4. Use newer lighthouse dependency for e2e fixes


pkg/subctl/cmd/verify.go
```diff
@@ -62,14 +62,20 @@ var verifyCmd = &cobra.Command{
 	Args: func(cmd *cobra.Command, args []string) error {
 		return checkValidateArguments(args)
 	},
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		configureTestingFramework(args)
 		if verifyConnectivity || verifyAll {
-			verifyconnectivity.RunE2E(&testing.T{})
+			if !verifyconnectivity.RunE2E(&testing.T{}) {
+				return fmt.Errorf("Connectivity E2E failed")
```
mangelajo, line 69: use exitOnError, you can probably refactor exit on error into two functions:
```
// exitOnError will print your error nicely and exit in case of error
func exitOnError(message string, err error) {
	if err != nil {
		exitWithErrorMsg(fmt.Sprintf("%s: %s", message, err)
	}
}


func exitWithErrorMsg(message string) {
		fmt.Fprintln(os.Stderr, message)
		fmt.Fprintln(os.Stderr, "")
		PrintSubctlVersion(os.Stderr)
		fmt.Fprintln(os.Stderr, "")
		os.Exit(1)
}
```

mangelajo, line 69: exitWithErrorMsg("Discovery E2E failed")

vthapar: Done

mangelajo:
Code looks good but we need to see why is it failing yet in some cases.

scripts/kind-e2e/e2e.sh
```diff
@@ -37,7 +37,7 @@ with_context cluster3 subctl_install_subm
 
 verify="--connectivity"
 if [[ "$lighthouse" == "true" ]]; then
-    verify="--all"
+    verify="--connectivity=false --service-discovery"
```
mkolesnik, line 40: So if deployed with LH it'll only check the LH things?

mangelajo: Yeah, the idea is that the other combination already checked the connectivity stuff, why repeat?

mangelajo: Why --connectivity=false now? why's that flag necessary?

pkg/subctl/cmd/verify.go
```diff
@@ -89,7 +93,7 @@ func configureTestingFramework(args []string) {
 	// the cluster IDs that will be registered in the Cluster CRDs by submariner
 	framework.TestContext.ClusterIDs = []string{"ClusterA", "ClusterB"}
 
-	config.GinkgoConfig.FocusString = "dataplane"
+	config.GinkgoConfig.FocusString = "dataplane*"
```
mangelajo, line 96: We need to modify the FocusString based on connectivity / discovery

vthapar: > We need to modify the FocusString based on connectivity / discovery

This change doesn't work, I tried already. I am modifying it for discovery.

scripts/kind-e2e/e2e.sh
mangelajo: Oh, I see that's "true" by default

Can we change to use the first argument as the type of connectivity?, or make them sub-commands?

pkg/subctl/verify/connectivity/run.go
```diff
@@ -24,6 +24,6 @@ import (
 	_ "github.com/submariner-io/submariner/test/e2e/framework"
```
mangelajo, line 24: let's remove those two files, connectivity/run and discovery/run. we just need to call shipyard with the right focus regex :)

scripts/kind-e2e/e2e.sh
mkolesnik: > Yeah, the idea is that the other combination already checked the connectivity stuff, why repeat?

Sure in CI, but locally of course it'll not work like that and you'll only get LH without basic connectivity tests. If that's good or bad IDK, but it'll happen

vthapar: @mkolesnik which is fine coz LH should test only discovery, not connectivity. If someone wants to test LH with basic connectivity locally they can run with `-all`.

vthapar: > Oh, I see that's "true" by default
> 
> Can we change to use the first argument as the type of connectivity?, or make them sub-commands?

subcommands sounds like good idea, will add it in next patch.

pkg/subctl/verify/connectivity/run.go
vthapar: Agree. testing changes locally to make sure it works, will push next.

scripts/kind-e2e/e2e.sh
mkolesnik: > @mkolesnik which is fine coz LH should test only discovery, not connectivity. If someone wants to test LH with basic connectivity locally they can run with `-all`.

How would you run it? `make e2e <what goes here?>`

vthapar: @mkolesnik Not make, these options are added to `verify` Locally one would normally do `make deploy` and then run `verify`

mkolesnik: OK but this is part of `make e2e` call and you can run that one locally not only in CI..

vthapar: Done. `connectivity` is true by default, so we just set `service-discovery`

scripts/kind-e2e/e2e.sh
```diff
@@ -37,7 +37,7 @@ with_context cluster3 subctl_install_subm
 
 verify="--connectivity"
 if [[ "$lighthouse" == "true" ]]; then
-    verify="--all"
+    verify="--service-discovery"
```
mkolesnik, line 40: This will still run only SD tests, or am I missing something?

Maybe you meant
```suggestion
    verify+=" --service-discovery"
```

vthapar: > This will still run only SD tests, or am I missing something?
> 
> Maybe you meant

conenctivity flag is set to true by default. This is setting service-discovery flag to true. with both true, it is basically all.

mkolesnik: OK just with line 38 it's a bit confusing, maybe that line should be removed

go.sum
```diff
@@ -1171,6 +1193,10 @@ gopkg.in/yaml.v2 v2.2.4 h1:/eiJrUcujPVeJ3xlSWaiNi3uSVmDGBK1pDHUHAnao1I=
 gopkg.in/yaml.v2 v2.2.4/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gopkg.in/yaml.v2 v2.2.8 h1:obN1ZagJSUGI0Ek/LBmuj4SNLPfIny3KsKFopxRdj10=
 gopkg.in/yaml.v2 v2.2.8/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+<<<<<<< HEAD
```
mangelajo, line 1196: Opps ;)

vthapar: done. surprisingly make passes with these entries.

scripts/kind-e2e/e2e.sh
mangelajo: I'd use mike's suggestion, it's at least more clear :) although we will be changing it soon

mangelajo:
Although we are seeing CI issues, they are unrelated to the content of this patch, this patch will make them surface more because we are testing more.

Issue #346, opened 2020-04-28
Remove Openshift CoreDNS operator


aswinsuryan:
When lighthouse moves to forward-plugin implementation, Openshift core-Dns operator shall be removed.

PR #427, opened 2020-06-01
Add Openshift support for forward plugin

aswinsuryan:
Uses cluster DNS operator to configure openshift-dns
coremap.

Fixes: #346


pkg/controller/helpers/helpers.go
```diff
@@ -195,3 +196,47 @@ func ReconcileService(owner metav1.Object, service *corev1.Service, reqLogger lo
 
 	return service, errorutil.WithMessagef(err, "error creating or updating Service %s/%s", service.Namespace, service.Name)
 }
+
+func ReconcileClusterDNSOperator(owner metav1.Object, dnsOperator *operatorv1.DNS, reqLogger logr.Logger,
+	client client.Client, scheme *runtime.Scheme) (*operatorv1.DNS, error) {
+	var err error
+
+	// Set the owner and controller
+	if err = controllerutil.SetControllerReference(owner, dnsOperator, scheme); err != nil {
```
mangelajo, line 205: not sure if we can  / should change the owner for an object we didn't create?

aswinsuryan: It did not throw any error, not sure if it can have multiple owners, I removed it will add it later if when we add support for operator status.

mangelajo:
looks good @aswinsuryan thanks for looking at my previous comments and acting quickly

PR #428, opened 2020-06-01
Fix bleeding edge release

mkolesnik:
Fix the quotes which broken the job, also the script should build the
tar.xz into bin directory.


manosnoam:
Why did you make it "xz" and not the common "gz" extension ?

skitt:
> Why did you make it "xz" and not the common "gz" extension ?

`xz` provides better compression than `gzip` and is supported everywhere.

PR #435, opened 2020-06-02
Update the verify command behaviour

mangelajo:
The gateway-failover (redundancy) tests are added.

The new format is:
```
subctl verify k1 k2 [--enable-disruptive] [--only service-discovery,network-policy,...]
```

subctl will prompt for disruptive tests if --enable-disruptive is not enabled and any
disruptive test is expected to be executed, so the user is warned before execution
and excluded if the user does not want to perform those verifications.

### DIFF
--- a/pkg/subctl/cmd/verify.go
+++ b/pkg/subctl/cmd/verify.go
@@ -18,8 +18,11 @@ package cmd
 
 import (
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


mangelajo:
> The e2e seems to be failing.

It’s fixed now, I rebased the patch but last run passed :-)

PR #440, opened 2020-06-03
Mount the host file system in the engine

skitt:
... so that the engine container can use the host iptables.

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -369,6 +369,9 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 					Image:           getImagePath(cr, engineImage),
 					Command:         []string{"submariner.sh"},
 					SecurityContext: &security_context_all_caps_privilaged,
+					VolumeMounts: []corev1.VolumeMount{
+						{Name: "host-slash", MountPath: "/host", ReadOnly: true},
+					},
 					Env: []corev1.EnvVar{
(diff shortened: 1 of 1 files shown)


sridhargaddam:
> ... so that the engine container can use the host iptables.

Just curious: With strongswan we did not have this requirement. Any idea what kind of iptable rules we are programming with Libreswan?

PS: Please ignore the failure in globanet job, its fixed via - https://github.com/submariner-io/shipyard/pull/201

> Signed-off-by: Stephen Kitt [skitt@redhat.com](mailto:skitt@redhat.com)

skitt:
> > ... so that the engine container can use the host iptables.
> 
> Just curious: With strongswan we did not have this requirement. Any idea what kind of iptable rules we are programming with Libreswan?

I don’t think we’re programming anything, but Libreswan checks for `iptables` when it starts, which produces a warning if the configured `iptables` don’t match the in-kernel tables. (Libreswan can use `iptables` for nflog devices, I think, but that’s it in terms of writing anything AFAICT; it uses it a lot in its diagnostic output, _e.g._ `ipsec barf`.)

sridhargaddam:
> > > ... so that the engine container can use the host iptables.
> > 
> > 
> > Just curious: With strongswan we did not have this requirement. Any idea what kind of iptable rules we are programming with Libreswan?
> 
> I don’t think we’re programming anything, but Libreswan checks for `iptables` when it starts, which produces a warning if the configured `iptables` don’t match the in-kernel tables. (Libreswan can use `iptables` for nflog devices, I think, but that’s it in terms of writing anything AFAICT; it uses it a lot in its diagnostic output, _e.g._ `ipsec barf`.)

Okay, thanks.

Issue #444, opened 2020-06-08
Operator: Fix terminationGracePeriodSeconds in Submariner Pods


sridhargaddam:
In K8s the default termination grace period for a Pod is 30
secs. However, for Submariner Pods, we were configuring this as
0 secs, because of which, cleanup was not happening in a
consistent manner and we are seeing failures in CI jobs.

PR #445, opened 2020-06-08
Fix terminationGracePeriodSeconds in Submariner Pods

sridhargaddam:
In K8s the default termination grace period for a Pod is 30
secs. However, for Submariner Pods, we were configuring this as
0 secs, because of which, cleanup was not happening in a
consistent manner and we are seeing failures in CI jobs.

This PR modifies the terminationGracePeriodSeconds to 10 secs.

Along with this change, for the Globalnet Pod, it also uses the proper
namespace of submariner in the excluded namespaces.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/444

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -346,7 +346,7 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 		RunAsNonRoot:             &runAsNonRoot}
 
 	// Create Pod
-	terminationGracePeriodSeconds := int64(0)
+	terminationGracePeriodSeconds := int64(10)
 	podTemplate := corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: labels,
@@ -438,7 +438,7 @@ func newRouteAgentDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 		RunAsNonRoot:             &runAsNonRoot,
 	}
 
-	terminationGracePeriodSeconds := int64(0)
+	terminationGracePeriodSeconds := int64(10)
 
 	routeAgentDaemonSet := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -510,7 +510,7 @@ func newGlobalnetDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 		RunAsNonRoot:             &runAsNonRoot,
 	}
 
-	terminationGracePeriodSeconds := int64(0)
+	terminationGracePeriodSeconds := int64(10)
 
 	globalnetDaemonSet := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


sridhargaddam:
The failing test is a known failure (i.e., Pods with HostNetworking will not get globalIp annotation) and is already fixed in shipyard.

mangelajo:
Can we generalize the patch and do the same to the other pods?,

I submariner-engine and routeagent also do cleanups at exit. In some cases that is going to be neglected.

pkg/controller/submariner/submariner_controller.go
```diff
@@ -537,7 +537,7 @@ func newGlobalnetDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 							Env: []corev1.EnvVar{
 								{Name: "SUBMARINER_NAMESPACE", Value: cr.Spec.Namespace},
 								{Name: "SUBMARINER_CLUSTERID", Value: cr.Spec.ClusterID},
-								{Name: "SUBMARINER_EXCLUDENS", Value: "submariner,kube-system,operators"},
+								{Name: "SUBMARINER_EXCLUDENS", Value: "submariner-operator,kube-system,operators"},
```
dfarrell07, line 540: If this is wrong, there are at least two other places it seems it should be changed.

```
submariner-operator/pkg/controller/submariner/submariner_controller.go:534						
submariner-operator/pkg/controller/servicediscovery/servicediscovery_controller.go:207							
submariner-charts/submariner/templates/globalnet.yaml:33
```

sridhargaddam: Yes @dfarrell07 I noticed that but I'm not sure if Service Discovery/LH controller also needs some grace period. Basically, gracePeriod is required if a Pod wants to do cleanup before its terminated. 

If the other components also require cleanup, IMO we can propose a separate PR for that. WDYS?

sridhargaddam:
The failing test is a known failure (i.e., Pods with HostNetworking will not get globalIp annotation) and is already fixed in shipyard.

pkg/controller/submariner/submariner_controller.go
dfarrell07: And if this is wrong, can we explain why more in the commit message? Might also be better in a separate commit from the timing change.

dfarrell07: I mean the SUBMARINER_EXCLUDENS change has parallels elsewhere, not the grace period. Can we explain why that's wrong in the commit message (and/or maybe use two commits)?

sridhargaddam: Ack. Originally this PR was focused only on Globalnet Pod and the scope has increased based on the review comments. I added a brief note on the EXCLUDENS change, but I agree this can be a separate commit as well. Let me update it. BTW the EXCLUDENS is not a fix, its more like an optimization. I'll elaborate more in the commit message.

dfarrell07: Okay, thanks @sridhargaddam!

sridhargaddam: thanks for reviewing @dfarrell07 :)

PR #466, opened 2020-06-17
subctl: add export service

### LINKED ISSUE #465: subctl: add create service-export
Add a subcommand to allow easy creation of `ServiceExport`. This will be helpful for test automation and first time users as they won't have to write yaml.

vthapar:
Fixes #465

### DIFF
--- a/pkg/subctl/cmd/export.go
+++ b/pkg/subctl/cmd/export.go
@@ -1,0 +1,102 @@
+/*
+© 2020 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/submariner-io/lighthouse/pkg/apis/lighthouse.submariner.io/v2alpha1"
+	lighthouseClientset "github.com/submariner-io/lighthouse/pkg/client/clientset/versioned"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	exportCmd = &cobra.Command{
+		Use:   "export",
+		Short: "Exports a resource to other clusters",
+		Long:  "This command exports a resource so it is accessible to other clusters",
+	}
+	exportServiceCmd = &cobra.Command{
+		Use:   "service",
+		Short: "Exports a Service to other clusters",
+		Long:  "This command creates a ServiceExport resource with a given name which causes the Service of the same name to be accessible to other clusters",
+		Run:   exportService,
+	}
+	serviceNamespace string
+)
+
+func init() {
+	addKubeconfigFlag(exportCmd)
+	addServiceExportFlags(exportServiceCmd)
+	exportCmd.AddCommand(exportServiceCmd)
+	rootCmd.AddCommand(exportCmd)
+}
+
+func addServiceExportFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service to be exported")
+
+}
+func exportService(cmd *cobra.Command, args []string) {
+	err := validateArguments(args)
+	exitOnError("Insufficient arguments", err)
+
+	clientConfig := getClientConfig(kubeConfig, kubeContext)
+	restConfig, err := clientConfig.ClientConfig()
+
+	exitOnError("Error connecting to the target cluster", err)
+
+	_, clientSet, err := getClients(restConfig)
+	exitOnError("Error connecting to the target cluster", err)
+
+	lhClientSet, err := lighthouseClientset.NewForConfig(restConfig)
+	exitOnError("Error connecting to the target cluster", err)
+
+	if serviceNamespace == "" {
+		if serviceNamespace, _, err = clientConfig.Namespace(); err != nil {
+			serviceNamespace = "default"
+		}
+	}
+	svcName := args[0]
+	_, err = clientSet.CoreV1().Services(serviceNamespace).Get(svcName, metav1.GetOptions{})
+	exitOnError(fmt.Sprintf("Unable to find the Service %q in namespace %q", svcName, serviceNamespace), err)
+
+	newServiceExport := v2alpha1.ServiceExport{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      svcName,
+			Namespace: serviceNamespace,
+		},
+	}
+	_, err = lhClientSet.LighthouseV2alpha1().ServiceExports(serviceNamespace).Create(&newServiceExport)
+	if k8serrors.IsAlreadyExists(err) {
+		fmt.Fprintln(os.Stdout, "Service already exported")
+		return
+	}
+	exitOnError("Failed to export Service", err)
+	fmt.Fprintln(os.Stdout, "Service exported successfully")
+
+}
+
+func validateArguments(args []string) error {
+	if len(args) == 0 || args[0] == "" {
+		return errors.New("Name of the Service to be exported must be specified")
+	}
+	return nil
+}
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -126,6 +126,10 @@ func getClients(config *rest.Config) (dynamic.Interface, kubernetes.Interface, e
 }
 
 func getRestConfig(kubeConfigPath string, kubeContext string) (*rest.Config, error) {
+	return getClientConfig(kubeConfigPath, kubeContext).ClientConfig()
+}
+
+func getClientConfig(kubeConfigPath string, kubeContext string) clientcmd.ClientConfig {
 	rules := clientcmd.NewDefaultClientConfigLoadingRules()
 	rules.ExplicitPath = kubeConfigPath
 	rules.DefaultClientConfig = &clientcmd.DefaultClientConfig
@@ -133,7 +137,7 @@ func getRestConfig(kubeConfigPath string, kubeContext string) (*rest.Config, err
 	if kubeContext != "" {
 		overrides.CurrentContext = kubeContext
 	}
-	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides).ClientConfig()
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
 }
 
 func handleNodeLabels(config *rest.Config) error {


pkg/subctl/cmd/create.go
```diff
@@ -0,0 +1,89 @@
...
+	exitOnError("Insufficient arguments", err)
+	config, err := getRestConfig(kubeConfig, kubeContext)
+	exitOnError("Error connecting to the target cluster", err)
+	_, clientSet, err := getClients(config)
+	exitOnError("Error connecting to the target cluster", err)
+	lhClientSet, err := createLighthouseClient(config)
+	exitOnError("Error connecting to the target cluster", err)
+	svcName := args[0]
+	_, err = clientSet.CoreV1().Services(serviceNamespace).Get(svcName, metav1.GetOptions{})
+	exitOnError(fmt.Sprintf("Unable to find service %s in namespace %s", svcName, serviceNamespace), err)
```
skitt, line 67: Users might want to export services before they’re created, shouldn’t we allow that? Technically, the order doesn’t matter, does it?

vthapar:
> It would be nice to tell the user that the service was successfully exported (or marked for export, if we allow exporting services which don’t exist yet).
I wanted to but wasn't sure what should I use for that, `fmt.Printf()`?

pkg/subctl/cmd/create.go
vthapar: > Users might want to export services before they’re created, shouldn’t we allow that? Technically, the order doesn’t matter, does it?

Currently we don't, adding this support in 0.5.0. Once we add, will remove this check.

pkg/subctl/cmd/create.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func createLighthouseClient(restConfig *rest.Config) (*lighthouseClientset.Clientset, error) {
+	clientSet, err := lighthouseClientset.NewForConfig(restConfig)
+	return clientSet, err
+}
+
+func validateArguments(args []string) error {
+	if len(args) == 0 || args[0] == "" {
+		return errors.New("Name of Service to be Exported not specified")
```
tpantelis, line 86:
```suggestion
		return errors.New("The name of the Service to export must be specified")
```

pkg/subctl/cmd/create.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+	}
+	_, err = lhClientSet.LighthouseV2alpha1().ServiceExports(serviceNamespace).Create(&newServiceExport)
+	exitOnError("Failed to create ServiceExport", err)
+
+}
+
+func createLighthouseClient(restConfig *rest.Config) (*lighthouseClientset.Clientset, error) {
+	clientSet, err := lighthouseClientset.NewForConfig(restConfig)
+	return clientSet, err
```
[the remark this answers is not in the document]
vthapar: Removed this method and calling `lighthouseClientset.NewforConfig()` directly

pkg/subctl/cmd/create.go
```diff
@@ -0,0 +1,89 @@
...
+
+var (
+	createCmd = &cobra.Command{
+		Use:   "create",
+		Short: "Create a submariner resource",
+		Long:  `This command provides an easy CLI to create specific submariner resources`,
+	}
+	createServiceExportCmd = &cobra.Command{
+		Use:   "service-export",
+		Short: "Creates ServiceExport resource",
```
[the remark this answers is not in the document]
vthapar: done

vthapar:
> I think it would be better to add features based on the functions as perceived by end-users, not as implemented (currently) by Submariner. The user doesn’t care that exporting a service involves creating a ServiceExport resource; they care about exporting a service, that’s all.
> 
> May I suggest instead a “service” command, supporting exporting, listing, and “un-exporting” services?

This is not about implementation though, `ServiceExport` is a resource user has to explicitly create to export a service as per the spec. User will know export status from ServiceExport, not from Service.

It were an implementation if 'Export' were a verb available on Service and we were providing an alternate means to do it by creating ServiceExport. Once this is available in Kubernetes users will be required to create ServiceExports. What we are doing here is tooling support to make it easy for user than having to create a yaml.

skitt:
> > I think it would be better to add features based on the functions as perceived by end-users, not as implemented (currently) by Submariner. The user doesn’t care that exporting a service involves creating a ServiceExport resource; they care about exporting a service, that’s all.
> > May I suggest instead a “service” command, supporting exporting, listing, and “un-exporting” services?
> 
> This is not about implementation though, `ServiceExport` is a resource user has to explicitly create to export a service as per the spec. User will know export status from ServiceExport, not from Service.
> 
> It were an implementation if 'Export' were a verb available on Service and we were providing an alternate means to do it by creating ServiceExport. Once this is available in Kubernetes users will be required to create ServiceExports. What we are doing here is tooling support to make it easy for user than having to create a yaml.
> 
> EDIT: Another reason is also if we go with `subctl service`, what should `list` do? List services, exports? Other way to go about it would be something like `subctl export service`. Then we can later have option to add more options like exporting all services in a namespace with `subctl export namespace`

My point about implementation is that the user shouldn’t have to know about `ServiceExport` etc., IMO; they should be able to say “export this service” in some way or other, and have that happen.

The point you made in the Lighthouse sync call about `subctl service` being too generic is an important aspect. `subctl exported-service` would be too long. I like your `subctl export service` suggestion, it follows the action paradigm used in `kubectl`.

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,90 @@
...
+	_, err = clientSet.CoreV1().Services(serviceNamespace).Get(svcName, metav1.GetOptions{})
+	exitOnError(fmt.Sprintf("Unable to find the Service %q in namespace %q", svcName, serviceNamespace), err)
+
+	newServiceExport := v2alpha1.ServiceExport{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      svcName,
+			Namespace: serviceNamespace,
+		},
+	}
+	_, err = lhClientSet.LighthouseV2alpha1().ServiceExports(serviceNamespace).Create(&newServiceExport)
```
mangelajo, line 79: just nit: can we check if the error is that "it already exists" and print nice message (not error): fmt.Fprintln(os.Stdout, "Service already exported")

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,90 @@
...
+	exitOnError("Failed to export Service", err)
+	fmt.Fprint(os.Stdout, "Service exported successfully")
```
mangelajo, line 81:
```suggestion
	fmt.Fprintln(os.Stdout, "Service exported successfully")
```

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,90 @@
...
+
+func init() {
+	addKubeconfigFlag(exportCmd)
+	addServiceExportFlags(exportServiceCmd)
+	exportCmd.AddCommand(exportServiceCmd)
+	rootCmd.AddCommand(exportCmd)
+}
+
+func addServiceExportFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&serviceNamespace, "namespace", "default", "Namespace in which service to be exported belongs")
```
mangelajo, line 53: I would expect that we could grab the current namespace in the kubeconfig. So it behaves like kubectl in this, otherwise it would be confusing.

stackoverflow to the rescue of us: https://stackoverflow.com/questions/55314152/how-to-get-namespace-from-current-context-set-in-kube-config ':D


```
contexts:
- context:
    cluster: default
    namespace: pr-brancher-webhook
    user: default
  name: default
```

and then...


```
current-context: kind-cluster3
```

vthapar: done.

pkg/subctl/cmd/export.go
vthapar: Done

pkg/subctl/cmd/export.go
vthapar: Done

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+	lighthouseClientset "github.com/submariner-io/lighthouse/pkg/client/clientset/versioned"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	exportCmd = &cobra.Command{
+		Use:   "export",
+		Short: "Exports a service",
+		Long:  "This command simplifies exporting service(s) to other clusters",
```
tpantelis, line 35: It seems this should be generalized as we may want to export other things.

```suggestion
		Short: "Exports a resource to other clusters",
		Long:  "This command exports a resource so it is accessible to other clusters",
```

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	exportServiceCmd = &cobra.Command{
+		Use:   "service",
+		Short: "Creates a ServiceExport resource",
+		Long:  "This command creates a ServiceExport resource with a given name which causes the Service of the same name to be accessible to other clusters",
```
tpantelis, line 40:
```suggestion
		Short: "Exports a Service to other clusters",
		Long:  "This command creates a ServiceExport resource with the given name of a corresponding Service which causes the Service to be accessible to other clusters",
```

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+
+func init() {
+	addKubeconfigFlag(exportCmd)
+	addServiceExportFlags(exportServiceCmd)
+	exportCmd.AddCommand(exportServiceCmd)
+	rootCmd.AddCommand(exportCmd)
+}
+
+func addServiceExportFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace in which service to be exported belongs")
```
tpantelis, line 54:
```suggestion
	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service to be exported")
```

pkg/subctl/cmd/export.go
vthapar: I considered this but decided not to for now. I don't think we have anything other than services to be exported. We'll add something like `namespace` but even in that case it will be restricted to only services in that namespace. In case we do support something else, we can revisit help text. Thoughts?

tpantelis: So  `subctl export --help`  shows:

_This command simplifies exporting service(s) to other clusters_

_Usage:
  subctl export [command]_

_Available Commands_:
  _service     Creates a ServiceExport resource_

_Flags_: …

I know we’re only using `export` for services (right now) and it’s not a big deal but I still think technically the help should be generic and not reflect that of the/a sub-command.  I mean if we’re going to tie export to its sub-command then we might as well do away with the intermediate and go back to `export-service` :)

pkg/subctl/cmd/export.go
tpantelis: Sorry one more tweak:

```suggestion
		Short: "Exports a Service to other clusters",
		Long:  "This command creates a ServiceExport resource with a given name which causes the Service of the same name to be accessible to other clusters",
```

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+
+func init() {
+	addKubeconfigFlag(exportCmd)
+	addServiceExportFlags(exportServiceCmd)
+	exportCmd.AddCommand(exportServiceCmd)
+	rootCmd.AddCommand(exportCmd)
+}
+
+func addServiceExportFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service ot be exported")
```
tpantelis, line 54:
```suggestion
	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service to be exported")
```

Sorry just noticed this typo 😞

vthapar: Done

pkg/subctl/cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+
+func init() {
+	addKubeconfigFlag(exportCmd)
+	addServiceExportFlags(exportServiceCmd)
+	exportCmd.AddCommand(exportServiceCmd)
+	rootCmd.AddCommand(exportCmd)
+}
+
+func addServiceExportFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service o obe exported")
```
tpantelis, line 54: oops

```suggestion
	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "Namespace of the service to be exported")
```

vthapar: done

manosnoam:
@tpantelis 
Did we add unit-test for this new subctl feature ?
Can you show usage example (subctl command) ?

tpantelis:
> @tpantelis
> Did we add unit-test for this new subctl feature ?
> Can you show usage example (subctl command) ?

There's no unit testing. Try `subctl export service --help`.

aswinsuryan:
> > @tpantelis
> > Did we add unit-test for this new subctl feature ?
> > Can you show usage example (subctl command) ?
> 
> There's no unit testing. Try `subctl export service --help`.
@manosnoam 
You can refer https://github.com/submariner-io/submariner-website/pull/202

Issue #299, opened 2020-04-06
Improve the current auto-allocation mechanism of globalCIDR in subctl


sridhargaddam:
Currently, subctl tries to automatically allocate a globalCIDR chunk from the un-allocated globalnet-cidr-range to a cluster during join operation. Inorder to do this, it depends on the submariner clusterCRD on the Broker. However, for the clusterCRD to be populated by the respective cluster on the Broker usually takes sometime. Because of this, we are unable to execute "subctl join ...." commands in parallel. 

This issue is to enhance the current implementation to use an appropriate mechanism like a  ConfigMap that stores the [clusterName: globalCIDR] on the Broker cluster and use this info instead of depending on clusterCRD.

sridhargaddam:
@Jaanki since you will be going on vacation, shall I look into this issue?

Jaanki:
Sure @sridhargaddam. I have half of the logic implemented at PR #475.

manosnoam:
In addition to having auto-generated CIDRs, please also alert the user if trying to use same globalnet-cidr for two clusters. (i.e. subctl join command should fail for such case.)

Jaanki:
@Noam, the current implementation does that.

sridhargaddam:
> Sure @sridhargaddam. I have half of the logic implemented at PR #475.

Done, I pushed a PR which addresses this issue along with some other use-cases.
Added you as co-author.
https://github.com/submariner-io/submariner-operator/pull/501

sridhargaddam:
> In addition to having auto-generated CIDRs, please also alert the user if trying to use same globalnet-cidr for two clusters. (i.e. subctl join command should fail for such case.)

Yes @manosnoam, it's handled. Please let us know if you find any issues.

manosnoam:
@sridhargaddam Now that it is closed, can I remove the static "wait" of 30 seconds - before joining another cluster (when using globalnet) ?

@nyechiel, @mangelajo 
Since this is kind of feature to be tested by QE, I re-open it, and assign to myself.
(as we discussed, it's important to move finished features to test, before closing)

sridhargaddam:
> @sridhargaddam Now that it is closed, can I remove the static "wait" of 30 seconds - before joining another cluster (when using globalnet) ?
> 

Yes @manosnoam . You can remove any sleep as well as remove the globalnet-cidr as part of subctl join command. The CIDRs will be auto-allocated.

> @nyechiel, @mangelajo
[...]

PR #475, opened 2020-06-17
[WIP] Improve GlobalCIDR auto-allocation mechanism

Jaanki:
Idea is to create a configmap while deploying broker
and then populating it with the auto-assigned GlobalCIDR
while joining the cluster.

Closes: #299


sridhargaddam:
This is handled via the following PR - https://github.com/submariner-io/submariner-operator/pull/501

PR #479, opened 2020-06-19
Bump to 0.4.0 rc3

mangelajo:


mangelajo:
ok the lighthouse failure was not related to lighthouse, a dataplane issue

skitt:
We need Admiral and Shipyard releases too... See #480 for some more on this.

PR #484, opened 2020-06-22
Fix the release versioning

mkolesnik:
Since this file gets updated in CI, the version is calculated as `dev`
instead of the real tag.


mangelajo:
I'm merging this as it's not altering the codebase itself and it's a fix for CI

PR #486, opened 2020-06-23
Fix make flags handling

mkolesnik:
Instead of dabbing around with MAKEFLAGS in ways that could potentially
break, just pass it down to the docker container to be used by the
sub-make running there.


skitt:
What was going wrong with `make` on the operator? One of the reasons behind the explicit `MAKEFLAGS` handling is so that they show up in the logs, perhaps we could log them here too...

mangelajo:
> What was going wrong with `make` on the operator? One of the reasons behind the explicit `MAKEFLAGS` handling is so that they show up in the logs, perhaps we could log them here too...

That using= ... was not being passed down, and globalnet / lighthouse was not being tested.

Makefile.dapper
```diff
@@ -10,11 +10,7 @@
...
 	@mv .dapper.tmp .dapper
 
-ifeq (,$(MAKEFLAGS))
 invoke_dapper = +./.dapper -m bind make -- $1
-else
-invoke_dapper = +./.dapper -m bind make -$(MAKEFLAGS) -- $1
-endif
 
 %: .dapper
 	$(call invoke_dapper,$@)
```
skitt, line 16:
```suggestion
	@echo Invoking Dapper, MAKEFLAGS=$(MAKEFLAGS)
	$(call invoke_dapper,$@)
```

PR #488, opened 2020-06-23
Add globalnet flag for lighthouse

vthapar:

### DIFF
--- a/pkg/apis/submariner/v1alpha1/servicediscovery_types.go
+++ b/pkg/apis/submariner/v1alpha1/servicediscovery_types.go
@@ -35,6 +35,7 @@ type ServiceDiscoverySpec struct {
 	Debug                    bool   `json:"debug"`
 	ClusterID                string `json:"clusterID"`
 	Namespace                string `json:"namespace"`
+	GlobalnetEnabled         bool   `json:"globalnetEnabled,omitempty"`
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
--- a/deploy/crds/submariner.io_servicediscoveries_crd.yaml
+++ b/deploy/crds/submariner.io_servicediscoveries_crd.yaml
@@ -43,6 +43,8 @@ spec:
               type: string
             debug:
               type: boolean
+            globalnetEnabled:
+              type: boolean
             namespace:
               type: string
             repository:
--- a/pkg/apis/submariner/v1alpha1/zz_generated.openapi.go
+++ b/pkg/apis/submariner/v1alpha1/zz_generated.openapi.go
@@ -125,6 +125,12 @@ func schema_pkg_apis_submariner_v1alpha1_ServiceDiscoverySpec(ref common.Referen
 							Format: "",
 						},
 					},
+					"globalnetEnabled": {
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"boolean"},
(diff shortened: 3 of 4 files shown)


pkg/apis/submariner/v1alpha1/servicediscovery_types.go
```diff
@@ -35,6 +35,7 @@ type ServiceDiscoverySpec struct {
 	Debug                    bool   `json:"debug"`
 	ClusterID                string `json:"clusterID"`
 	Namespace                string `json:"namespace"`
+	GlobalnetEnabled         bool   `json:"globalnetEnabled,omitempty"`
```
mangelajo, line 38: I think omitempty needs to be a pointer to bool.

mangelajo:
> I think it would be fine, but we need to check what happens with the CRD API version, do we need a bump for this?, and if so.. we would need the translator webhook for CRDs.
> 
> I suspect that being additive and optional field makes that change fine, specially in Alpha. check my comment about "omitempty"

Also, I wonder, if we start relying on this flag, then when the operator is updated, for example 0.4.0 to 0.5.0, we'd need to reprocess the "Submariner CR" into the "ServiceDiscovery CR" updating that field.

Can we try something like this manually? 
1) install with subctl release 0.4.1 ...
2) push this operator image somewhere, and compile the new subctl
3) run subctl join again .. ? 

I suspect it would work.

pkg/apis/submariner/v1alpha1/servicediscovery_types.go
vthapar: > I think omitempty needs to be a pointer to bool.

We're not using pointer for other bool `omitempty`

vthapar:
> > I think it would be fine, but we need to check what happens with the CRD API version, do we need a bump for this?, and if so.. we would need the translator webhook for CRDs.
> > I suspect that being additive and optional field makes that change fine, specially in Alpha. check my comment about "omitempty"
> 
> Also, I wonder, if we start relying on this flag, then when the operator is updated, for example 0.4.0 to 0.5.0, we'd need to reprocess the "Submariner CR" into the "ServiceDiscovery CR" updating that field.
> 
> Can we try something like this manually?
> 
> 1. install with subctl release 0.4.1 ...
> 2. push this operator image somewhere, and compile the new subctl
> 3. run subctl join again .. but specifying the image that you pushed on 2.
> 
> I suspect it would work.

It didn't work coz we don't really have reconcile logic yet. But this is something that can be handled easily when we do have an upgrade story in place. As an extra precaution, I'll use configMap in Lighthouse code if the flag is missing/nil and then we can remove that check once we have upgrade/migration sorted out.

mangelajo:
> > > I think it would be fine, but we need to check what happens with the CRD API version, do we need a bump for this?, and if so.. we would need the translator webhook for CRDs.
> > > I suspect that being additive and optional field makes that change fine, specially in Alpha. check my comment about "omitempty"
> > 
> > 
> > Also, I wonder, if we start relying on this flag, then when the operator is updated, for example 0.4.0 to 0.5.0, we'd need to reprocess the "Submariner CR" into the "ServiceDiscovery CR" updating that field.
> > Can we try something like this manually?
> > 
> > 1. install with subctl release 0.4.1 ...
> > 2. push this operator image somewhere, and compile the new subctl
> > 3. run subctl join again .. but specifying the image that you pushed on 2.
> > 
> > I suspect it would work.
> 
> It didn't work coz we don't really have reconcile logic yet. But this is something that can be handled easily when we do have an upgrade story in place. As an extra precaution, I'll use configMap in Lighthouse code if the flag is missing/nil and then we can remove that check once we have upgrade/migration sorted out.

Ack I added this for a follow up: https://github.com/submariner-io/submariner-operator/issues/510

PR #501, opened 2020-06-25
Migrate globalnet configuration to ConfigMap on Broker

sridhargaddam:
This PR implements the following changes.

1. Provides a mechanism to know the globalnet status as part of ConfigMap
  (submariner-globalnet-info) on the Broker Cluster.

2. Moves the globalnet configuration parameters like GlobalnetStatus,
   globalnet-cidr-range, and globalnet-cluster-size from the broker-info.subm
   file to configMap on the broker.

3. Currently, subctl tries to automatically allocate a globalCIDR chunk from
   the un-allocated globalnet-cidr-range to a cluster during join operation.
   Inorder to do this, it was depending on the submariner clusterCRD on the
   Broker. However, for the clusterCRD to be populated by the respective cluster
   on the Broker, it usually takes sometime. Because of this, we were unable to
   execute "subctl join ...." commands in parallel. This issue is now resolved
   as we now store the globalCIDR allocations as ConfigMap on Broker which is
   immediately updated as soon as allocation is made for a new cluster.

4. Overlapping CIDR validation for Service/Cluster CIDRs is removed from
   subctl and the same is handled in route-agent. 

5. Validations for various globalnet params are preserved, however the config
   is now stored/read from configMap instead of broker-info.subm file.

Note: For brownfield deployments, it could create issues if users try to use
the updated subctl with older clusters deployed with Globalnet. This will be
handled via https://github.com/submariner-io/submariner-operator/issues/504

Fixes Issue: https://github.com/submariner-io/submariner-operator/issues/299
Fixes Issue: https://github.com/submariner-io/submariner-operator/issues/505

Co-authored-by: Janki Chhatbar <jchhatba@redhat.com>


sridhargaddam:
Appreciate your review comments. Thanks.

pkg/subctl/cmd/join.go
```diff
@@ -182,22 +182,34 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+		globalnetCIDR, err = globalnet.AssignGlobalnetIPs(globalnetInfo, netconfig)
 		exitOnError("Error assigning Globalnet IPs", err)
+
+		if globalnetInfo.GlobalCidrInfo[clusterID] == nil ||
+			(globalnetInfo.GlobalCidrInfo[clusterID] != nil && globalnetInfo.GlobalCidrInfo[clusterID].GlobalCIDRs[0] != globalnetCIDR) {
+			var newClusterInfo broker.ClusterInfo
+			newClusterInfo.ClusterId = clusterID
+			newClusterInfo.GlobalCidr = []string{globalnetCIDR}
+
+			err = broker.UpdateGlobalnetConfigMap(brokerAdminClientset, brokerNamespace, newClusterInfo)
```
tpantelis, line 210: It seems we should encapsulate the entire workflow of `globalnet.GetGlobalNetworks` -> `globalnet.AssignGlobalnetIPs` -> "update ConfigMap" within `retry.RetryOnConflict` otherwise 2 concurrent joins could allocate the same CIDR. Unless I'm missing something ...

pkg/subctl/cmd/join.go
```diff
@@ -182,22 +182,34 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	brokerAdminConfig, err := subctlData.GetBrokerAdministratorConfig()
+	exitOnError("Error retrieving broker admin config", err)
+	brokerAdminClientset, err := kubernetes.NewForConfig(brokerAdminConfig)
+	exitOnError("Error retrieving broker admin connection", err)
+	brokerNamespace := string(subctlData.ClientToken.Data["namespace"])
+
 	status.Start("Discovering multi cluster details")
-	globalNetworks, err := globalnet.GetGlobalNetworks(subctlData)
-	exitOnError("Error getting Global network details", err)
+	globalnetInfo, err := globalnet.GetGlobalNetworks(brokerAdminClientset, brokerNamespace)
```
vthapar, line 192: This entire logic can be moved to `globalnet` and have single function call from `join.go` 
This will keep join code neat and move complexity to globalnet. Will be easier to add retry logic in globalnet than clutter in here.

pkg/broker/rbac.go
```diff
@@ -57,9 +57,9 @@ func NewBrokerAdminRole() *rbacv1.Role {
 				Resources: []string{"clusters", "endpoints"},
 			},
 			rbacv1.PolicyRule{
-				Verbs:     []string{"create", "get", "list", "delete"},
+				Verbs:     []string{"create", "get", "list", "patch", "update", "delete"},
```
vthapar, line 60: Do we need patch? Isn't update enough?

pkg/subctl/cmd/join.go
sridhargaddam: Let me move it to an API

pkg/subctl/cmd/join.go
sridhargaddam: Done

pkg/discovery/globalnet/globalnet.go
```diff
@@ -282,112 +236,122 @@ func isCIDRPreConfigured(clusterID string, globalNetworks map[string]*GlobalNetw
...
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error reading globalnet clusterInfo: %s", err)
 	}
-	// exitOnError("Error trying to discover multi-cluster network details", err)
-	if globalNetworks != nil {
-		ShowNetworks(globalNetworks)
+
+	var globalNetworks = make(map[string]*GlobalNetwork)
+	if len(clusterInfo) > 0 {
```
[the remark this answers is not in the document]
sridhargaddam: Aah, I see. I thought you are suggesting me to change len(clusterInfo) to clusterInfo  ;-) 
Will update, thanks.

mangelajo:
lol, it was merged while I was reviewing.. I'll leave a comment next time, actually my bad ':D

mangelajo:
I'll be working to move this detection into the operator itself, so both operator alone, and helm+operator could benefit in the future.

sridhargaddam:
> lol, it was merged while I was reviewing.. I'll leave a comment next time, actually my bad ':D

Yeah, I was reaching out to @skitt to take a look. I'll take a look at your updated PR, thanks @mangelajo

PR #503, opened 2020-06-26
Disable the Openshift log message for discovery

mangelajo:
When this is ran on non-openshift that log message appears on console,
when it's actually irrelevant. All discovery mechanism are tried
one after another until we find one that works.

```
[majopela@bluehat submariner-operator]$ bin/subctl show networks
I0626 11:06:21.463397    3263 openshift4.go:47] Attempted network discovery for OpenShift4, no clusternetworks CRD
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs:   [100.93.0.0/16]
        Cluster CIDRs:   [10.243.0.0/16]
```

### DIFF
--- a/pkg/discovery/network/openshift4.go
+++ b/pkg/discovery/network/openshift4.go
@@ -24,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog"
+
+	"github.com/submariner-io/submariner-operator/pkg/log"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/discovery/network/openshift4.go
```diff
@@ -43,8 +42,8 @@ func discoverOpenShift4Network(dynClient dynamic.Interface) (*ClusterNetwork, er
 	crClient := dynClient.Resource(openshift4clusterNetworkGVR)
 
 	cr, err := crClient.Get("default", metav1.GetOptions{})
+
 	if err != nil {
-		klog.Info("Attempted network discovery for OpenShift4, no clusternetworks CRD")
```
tpantelis, line 47: Perhaps we should at least log this to debug (including `err`) rather than silence it completely.

maayanf24: In any case, it should be symmetric to all of the discovery mechanism

mangelajo: Good comments.

tpantelis: Actually we just need to check for `IsNotFound` error. This is addressed as part of https://github.com/submariner-io/submariner-operator/pull/581 which should obviate the need for any changes in this PR.

mangelajo:
This would be better handled by: https://github.com/submariner-io/submariner-operator/pull/581

PR #512, opened 2020-06-30
Update ref to use latest shipyard master

sridhargaddam:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/spf13/cobra v0.0.7
 	github.com/spf13/pflag v1.0.5
 	github.com/submariner-io/lighthouse v0.4.1
-	github.com/submariner-io/shipyard v0.4.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/spf13/cobra v0.0.7
 	github.com/spf13/pflag v1.0.5
 	github.com/submariner-io/lighthouse v0.4.1
-	github.com/submariner-io/shipyard v0.4.1
+	github.com/submariner-io/shipyard v0.4.2-0.20200630133127-f7b9cae0e41d
```
mangelajo, line 23: The final merged commit seems to be: 78b626b2e2a1e4254cfd568bef89673d7dac0a5 we need to update this ref before merging (I guess this is why it's in draft , so thanks for that hint).

mangelajo:
Created an updated version here, https://github.com/submariner-io/submariner-operator/pull/528 I was waiting for dependabot, but it seems to be lazy.. or not liking our -pre0 thing

PR #514, opened 2020-07-01
Reset terminationGracePeriodSeconds for SM Engine and RouteAgent Pods

sridhargaddam:
In one of the earlier PRs, we modified the terminationGracePeriodSeconds
to 10 secs. However, its seen that this is causing more CI failures
during e2e redundancy tests. Ideally, once the Pods are terminated,
it should cleanup itself ASAP but it is seen that SM Pods are sometimes
taking time to exit and during this Period since there is no active SM
Pod running, this is triggering some failures.

Until we figure out the exact reason why the Pods are taking time for
cleanup, this PR resets the terminationGracePeriodSeconds.

Related to: https://github.com/submariner-io/submariner-operator/issues/444
Related to: https://github.com/submariner-io/submariner-operator/pull/445

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -354,7 +354,7 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 		RunAsNonRoot:             &runAsNonRoot}
 
 	// Create Pod
-	terminationGracePeriodSeconds := int64(10)
+	terminationGracePeriodSeconds := int64(1)
 	podTemplate := corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: labels,
@@ -446,7 +446,7 @@ func newRouteAgentDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 		RunAsNonRoot:             &runAsNonRoot,
 	}
 
-	terminationGracePeriodSeconds := int64(10)
+	terminationGracePeriodSeconds := int64(1)
 
 	routeAgentDaemonSet := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -518,7 +518,7 @@ func newGlobalnetDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 		RunAsNonRoot:             &runAsNonRoot,
 	}
 
-	terminationGracePeriodSeconds := int64(10)
+	terminationGracePeriodSeconds := int64(2)
 
 	globalnetDaemonSet := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
--- a/scripts/kind-e2e/lib_operator_verify_subm.sh
+++ b/scripts/kind-e2e/lib_operator_verify_subm.sh
@@ -334,7 +334,7 @@ function verify_subm_routeagent_pod() {
     [[ $(jq -r ".spec.volumes[] | select(.name==\"host-slash\").hostPath.path" $json_file) = '/' ]]
     validate_equals '.status.phase' 'Running'
     validate_equals '.metadata.namespace' $subm_ns
-    validate_equals '.spec.terminationGracePeriodSeconds' '10'
+    validate_equals '.spec.terminationGracePeriodSeconds' '1'
   done
 }
 


sridhargaddam:
> I think it's better to look for the root cause, if we hide it we will have less reasons to look for it

Agree to some extent, but the side-effect of this problem is seen on Globalnet Jobs.
Initially while debugging Globalnet failures, I realized that Globalnet Pod was not cleaning up the rules properly so pushed the PR to tune the terminationGracePeriodSeconds on Globalnet Pods alone. But based on the review comments, I had to modify the values of other Pods as you know.
https://github.com/submariner-io/submariner-operator/pull/445#issuecomment-640433487

So IMO the change we did for Engine/RouteAgent Pods has to be properly tested and then modified. Until then, my idea was to revert it so that we can stabilize the Globalnet jobs which are failing randomly.
I reported the following issue to track it further - https://github.com/submariner-io/submariner/issues/694

mangelajo:
May be we can give it 1 second grace period at least?, and keep the issue open.

The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.

tpantelis:
> May be we can give it 1 second grace period at least?, and keep the issue open.
> 
> The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.

yeah setting to 0 may trigger different failures wrt to Gateway deletion on shutdown.

Where specifically are the redundancy tests intermittently failing? Perhaps there's a timeout we can bump in those tests to offset the extra 10 sec?

mangelajo:
> > May be we can give it 1 second grace period at least?, and keep the issue open.
> > The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.
> 
> yeah setting to 0 may trigger different failures wrt to Gateway deletion on shutdown.
> 
> Where specifically are the redundancy tests intermittently failing? Perhaps there's a timeout we can bump in those tests to offset the extra 10 sec?

There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.

tpantelis:
> 
> There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.

Right but, either way, eventually the pods get killed so I'm wondering why the extra 10 sec causes failures.

sridhargaddam:
> > There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.
> 
> Right but, either way, eventually the pods get killed so I'm wondering why the extra 10 sec causes failures.

Basically, it's like this. Most of the failures seen in Globalnet jobs are either in redundancy restarting of the GatewayPod or subsequent test. When e2e redundancy test terminates the GatewayPod, its taking time to exit. As part of this restart scenario, the local endpoint is first deleted and once the GatewayPod is restarted a new endpoint is published (with the node being the same).

GlobalnetPod has a listener to Endpoints and as soon as it sees Endpoint delete notification, it flushes various iptable chains and which are only re-created once the new Endpoint is created. So any delay in these will be affecting the e2e tests as our e2e tests are bit time sensitive and not designed for Globalnet use-cases (i.e., we do not wait for the globalIP annotation on the Pod before we start validating the connectivity).

sridhargaddam:
> > > May be we can give it 1 second grace period at least?, and keep the issue open.
> > > The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.
> > 
> > 
> > yeah setting to 0 may trigger different failures wrt to Gateway deletion on shutdown.
> > Where specifically are the redundancy tests intermittently failing? Perhaps there's a timeout we can bump in those tests to offset the extra 10 sec?
> 
> There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.

Agree, can we do it as part of the new issue that is reported - https://github.com/submariner-io/submariner/issues/694
and if you really want, I can keep 1 sec (although I would prefer to revert the time back to 0 secs until the issue is resolved). 
Please let me know, thanks.

mangelajo:
> > > > May be we can give it 1 second grace period at least?, and keep the issue open.
> > > > The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.
> > > 
> > > 
> > > yeah setting to 0 may trigger different failures wrt to Gateway deletion on shutdown.
> > > Where specifically are the redundancy tests intermittently failing? Perhaps there's a timeout we can bump in those tests to offset the extra 10 sec?
> > 
> > 
> > There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.
> 
> Agree, can we do it as part of the new issue that is reported - [submariner-io/submariner#694](https://github.com/submariner-io/submariner/issues/694)
> and if you really want, I can keep 1 sec (although I would prefer to revert the time back to 0 secs until the issue is resolved).
> Please let me know, thanks.

I'm good with 1 sec for now If we don't lose sight of the submariner-io/submariner#694 

This will make our E2E and user experience better overall.

pkg/controller/submariner/submariner_controller.go
```diff
@@ -349,7 +349,7 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 		RunAsNonRoot:             &runAsNonRoot}
 
 	// Create Pod
-	terminationGracePeriodSeconds := int64(10)
+	terminationGracePeriodSeconds := int64(0)
```
mangelajo, line 352:
```suggestion
	terminationGracePeriodSeconds := int64(1)
```

pkg/controller/submariner/submariner_controller.go
```diff
@@ -441,7 +441,7 @@ func newRouteAgentDaemonSet(cr *submopv1a1.Submariner) *appsv1.DaemonSet {
 		RunAsNonRoot:             &runAsNonRoot,
 	}
 
-	terminationGracePeriodSeconds := int64(10)
+	terminationGracePeriodSeconds := int64(0)
```
mangelajo, line 444:
```suggestion
	terminationGracePeriodSeconds := int64(1)
```

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -334,7 +334,7 @@ function verify_subm_routeagent_pod() {
     [[ $(jq -r ".spec.volumes[] | select(.name==\"host-slash\").hostPath.path" $json_file) = '/' ]]
     validate_equals '.status.phase' 'Running'
     validate_equals '.metadata.namespace' $subm_ns
-    validate_equals '.spec.terminationGracePeriodSeconds' '10'
+    validate_equals '.spec.terminationGracePeriodSeconds' '0'
```
mangelajo, line 337:
```suggestion
    validate_equals '.spec.terminationGracePeriodSeconds' '1'
```

pkg/controller/submariner/submariner_controller.go
sridhargaddam: Done

pkg/controller/submariner/submariner_controller.go
sridhargaddam: Done

scripts/kind-e2e/lib_operator_verify_subm.sh
sridhargaddam: Done

sridhargaddam:
> > > > > May be we can give it 1 second grace period at least?, and keep the issue open.
> > > > > The gateway agent does cleanup at exit. Ideally we should make sure they are able to do part of that job.
> > > > 
> > > > 
> > > > yeah setting to 0 may trigger different failures wrt to Gateway deletion on shutdown.
> > > > Where specifically are the redundancy tests intermittently failing? Perhaps there's a timeout we can bump in those tests to offset the extra 10 sec?
> > > 
> > > 
> > > There's a low-level issue in the gateway and routeagent pods, they don't terminate on sigterm... and keep hanging. I think we actually need to figure that out and fix it instead.
> > 
> > 
> > Agree, can we do it as part of the new issue that is reported - [submariner-io/submariner#694](https://github.com/submariner-io/submariner/issues/694)
> > and if you really want, I can keep 1 sec (although I would prefer to revert the time back to 0 secs until the issue is resolved).
> > Please let me know, thanks.
> 
> I'm good with 1 sec for now If we don't lose sight of the [submariner-io/submariner#694](https://github.com/submariner-io/submariner/issues/694)
> 
> This will make our E2E and user experience better overall.

Thanks @mangelajo

PR #517, opened 2020-07-01
Always build a dev version of the operator image

mkolesnik:
We don't want the local image we build to be affected by the existence
of any tags that can be created while the CI is running (e.g. the devel
tag).


skitt:
I’ve made the same change to #412 so we won’t lose this fix ;-).

skitt:
I think this is obsolete now that #412 is merged, but please check that the current master does what you expect in this regard.

mangelajo:
> I think this is obsolete now that #412 is merged, but please check that the current master does what you expect in this regard.

Can we close this then @mkolesnik @skitt ?

PR #518, opened 2020-07-01
Use shared targets from Shipyard

mkolesnik:
Validation, tests and other targets were migrated to Shipyard, use them instead of cloned scripts.


mkolesnik:
> Heh, I have the same PR waiting for #412, but it looks like that one is never going to get merged ;-).

Actually I was careful not to touch anything you had in #412 so this PR is complimentary :)

skitt:
> > Heh, I have the same PR waiting for #412, but it looks like that one is never going to get merged ;-).
> 
> Actually I was careful not to touch anything you had in #412 so this PR is complimentary :)

Yeah, I know, what I meant was that I had a branch locally with the same changes, waiting for #412 to be merged before I pushed it as a PR.

PR #523, opened 2020-07-01
Add subctl show versions cmd

maayanf24:
Add subctl show versions


mangelajo:
@maayanf24 also, we need to provide a more meaningful commit title, and add Resolve-Issue: #471 as part of the description instead.

pkg/controller/servicediscovery/servicediscovery_controller.go
```diff
@@ -35,8 +35,8 @@ var log = logf.Log.WithName("controller_servicediscovery")
 
 const (
 	componentName                 = "submariner-lighthouse"
-	deploymentName                = "submariner-lighthouse-agent"
-	lighthouseCoreDNSName         = "submariner-lighthouse-coredns"
+	DeploymentName                = "submariner-lighthouse-agent"
```
mangelajo, line 38: Please un-do all those changes which are not necessary now for the patch.

To grab older versions of the files you can use

git checkout HEAD~1 pkg/controller/servicediscovery/servicediscovery_controller.go
...
...
etc...
then when you do `git status` you will see those files are modified, commit them...
and the changes will be removed from your patch.

Note HEAD~1 assumes that your PR has only one patch, so it will grab the state of the file before your patch.

pkg/controller/submariner/submariner_controller.go
```diff
@@ -47,6 +47,11 @@ import (
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 )
 
+const (
+	RouteAgentName = "submariner-routeagent"
```
mangelajo, line 51: Ditto.

pkg/subctl/cmd/show_versions.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+
+	operatorFullImageStr := operatorConfig.Spec.Template.Spec.Containers[0].Image
+	img := strings.Split(operatorFullImageStr, ":")
+	var image string
+	var version string
+	if len(img) < 2 {
+		image = ""
+		version = img[0]
```
tpantelis, line 66:
```suggestion
		image = img[0]
```

PR #530, opened 2020-07-03
Allow the operator to retrieve Gateway objects

skitt:
Currently the Gateway types aren’t registered; this fixes that. It
turns out we also need to fill in the GatewayList meta-data if we want
to add it as-is to the status, so this also changes the status to
include the Gateway slice instead.


mangelajo:
Cool, no more errors about gateways: https://github.com/submariner-io/submariner-operator/pull/530/checks?check_run_id=834272393#step:5:4277 :-)

Issue #541, opened 2020-07-07
subctl join has a separate --operator-image that ignores --repository and --version


mangelajo:
This can lead into confusion, unless we have good reasons for keeping separate handling of those images
I propose removing it and keeping just --repository and --version for all images for simplicity.

For example this caused the issue on CI where the 2nd run of subctl over 'cluster2' was not providing this
parameter and overwriting the submariner-operator image.

PR #542, opened 2020-07-07
Remove --operator-image parameter

mangelajo:
Resolves-Issue: #541


pkg/subctl/cmd/join.go
```diff
@@ -354,3 +355,14 @@ func populateSubmarinerSpec(subctlData *datafile.SubctlData) submariner.Submarin
 	}
 	return submarinerSpec
 }
+
+func operatorImage() string {
+
```
mkolesnik, line 360:
```suggestion
```

PR #546, opened 2020-07-07
Fix image handling for operator

mangelajo:
Since https://github.com/submariner-io/submariner-operator/pull/542 default
repository handling is not done for submariner operator. This
moves the default repository as the parameter default and makes
sure that populateSubmarinerSpec will not tamper any global variable.


mangelajo:
@tpantelis we need this one to unblock noam. I broke the operator image construction on a previous patch. The default won't work.

mangelajo:
@manosnoam this will un-break your CI

PR #553, opened 2020-07-08
Don't modify error on Update

vthapar:
If Update returns errors.IsConflict, it needs to be retried. Don't
modify error returned by Update so RetryOnconflict logic can handle it
correctly. Any other errors will be logged/caught by callers.


pkg/utils/createorupdate.go
```diff
@@ -46,7 +46,7 @@ func CreateOrUpdateClusterRole(clientSet clientset.Interface, clusterRole *rbacv
 			// Potentially retried
 			_, err = clientSet.RbacV1().ClusterRoles().Update(clusterRole)
 			if err != nil {
-				return fmt.Errorf("failed to update pre-existing cluster role %s : %v", clusterRole.Name, err)
+				return err
```
sridhargaddam, line 49: you can remove the if check all-together and simply use "return err"

vthapar: > you can remove the if check all-together and simply use "return err"

Will do. I was considering checking for conflict error, but late realized we're logging error in caller.

Issue #557, opened 2020-07-09
submariner-operator does not configure CIDRs in openshift


mangelajo:
The root cause is that we need a ClusterRole assigned to the submariner-operator. This ClusterRole shoud permit
list/get access to ClusterNetwork openshift objects. We also need permissions to list/get pods and ConfigMaps (for flannel/weave config detection).

The submariner-route-agent is in crash loop (because it can't find the IP addresses).

**submariner-operator logs:**

{"level":"error","ts":1594291790.7272599,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"submariner-controller","request":"submariner-operator/submariner","error":"Submariner.submariner.io \"submariner\" is invalid: [status.clusterCIDR: Required value, status.serviceCIDR: Required value]","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\tsubmariner-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:218\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:192\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:171\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\tsubmariner-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\tsubmariner-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\tsubmariner-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
{"level":"info","ts":1594291792.9526892,"logger":"controller_submariner","msg":"Reconciling Submariner","Request.Namespace":"submariner-operator","Request.Name":"submariner"}
[...]

PR #564, opened 2020-07-10
Fix the operator cluster permissions for discovery

mangelajo:
Discovery of the network looks at ClusterNetwork, and if that
fails it will look for specific configurations and generic
heuristics.

Fixes-Issue: #557

### DIFF
--- a/deploy/cluster_role_binding.yaml
+++ b/deploy/cluster_role_binding.yaml
@@ -1,11 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: submariner-operator:lighthouse
+  name: submariner-operator
 subjects:
     - kind: ServiceAccount
       name: submariner-operator
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: submariner-operator:lighthouse
+  name: submariner-operator
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -1,10 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: submariner-operator:lighthouse
+  name: submariner-operator
 rules:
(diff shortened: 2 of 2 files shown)


mangelajo:
I have manually verified this change and works on openshift: 
```
"level":"info","ts":1594383102.9205534,"logger":"controller_submariner","msg":"Cluster network discovered"}
{"level":"info","ts":1594383102.9205759,"logger":"controller_submariner","msg":"Discovered K8s network details","plugin":"OpenShift","clusterCIDRs":["10.128.0.0/14"],"serviceCIDRs":["172.30.0.0/16"]}
{"level":"info","ts":1594383102.9205904,"logger":"controller_submariner","msg":"Using detected CIDR","type":"Cluster","CIDR":"10.128.0.0/14"}
{"level":"info","ts":1594383102.9205978,"logger":"controller_submariner","msg":"Using detected CIDR","type":"Service","CIDR":"172.30.0.0/16"}
```

deploy/cluster_role_binding.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: submariner-operator:lighthouse
+  name: submariner-operator
```
sridhargaddam, line 4: Should we not change the corresponding references in 
pkg/subctl/operator/common/embeddedyamls/yamls.go ?

vthapar: Those are already correct. These are legacy bits which I don't think we're using anymore.

sridhargaddam: Ok, I'm not very familiar with this code. So I'll let @tpantelis to review this.

mangelajo: Yes, this is only changing the name, since now those are not operator lighthouse specific permissions, just general operator ones.

This file controls the cluster permissions that submariner-operator has in general for accessing the cluster, before we only needed that for the dns configmap, etc... but now we need more permissions, and it's not lighthouse specific anymore.

PR #566, opened 2020-07-11
Detect globalnet in subctl verify

tpantelis:
Read the operator Submariner resource from one of the clusters and
check the GlobalCIDR field.

### DIFF
--- a/pkg/subctl/cmd/verify.go
+++ b/pkg/subctl/cmd/verify.go
@@ -31,9 +31,11 @@ import (
 	_ "github.com/submariner-io/lighthouse/test/e2e/framework"
 	"github.com/submariner-io/shipyard/test/e2e"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
+	submarinerclientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/submarinercr"
 	_ "github.com/submariner-io/submariner/test/e2e/dataplane"
-	_ "github.com/submariner-io/submariner/test/e2e/framework"
 	_ "github.com/submariner-io/submariner/test/e2e/redundancy"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var (
@@ -52,6 +54,8 @@ func init() {
 	verifyCmd.Flags().BoolVar(&enableDisruptive, "enable-disruptive", false, "enable disruptive verifications like gateway-failover")
 	addVerifyFlags(verifyCmd)
 	rootCmd.AddCommand(verifyCmd)
+
+	framework.AddBeforeSuite(detectGlobalnet)
 }
 
 func addVerifyFlags(cmd *cobra.Command) {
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/verify.go
```diff
@@ -264,3 +268,18 @@ func getVerifyPatterns(csv string, includeDisruptive bool) ([]string, []string,
 	}
 	return outputPatterns, outputVerifications, nil
 }
+
+func detectGlobalnet() {
+	submarinerClient, err := submarinerclientset.NewForConfig(framework.RestConfigs[framework.ClusterA])
+	if err != nil {
+		exitWithErrorMsg(fmt.Sprintf("Error creating submariner client: %v", err))
+	}
```
mangelajo, line 276:
```suggestion
	exitOnError("Error creating submariner client", err)
```

pkg/subctl/cmd/verify.go
```diff
@@ -264,3 +268,18 @@ func getVerifyPatterns(csv string, includeDisruptive bool) ([]string, []string,
...
+
+	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName,
+		v1.GetOptions{})
+	if err != nil {
+		exitWithErrorMsg(fmt.Sprintf("Error obtaining Submariner resource: %v", err))
+	}
```
mangelajo, line 282:
```suggestion
	exitOnError("Error obtaining Submariner resource", err)
```

pkg/subctl/cmd/verify.go
tpantelis: Done

pkg/subctl/cmd/verify.go
tpantelis: Done

pkg/subctl/cmd/verify.go
```diff
@@ -264,3 +268,14 @@ func getVerifyPatterns(csv string, includeDisruptive bool) ([]string, []string,
 	}
 	return outputPatterns, outputVerifications, nil
 }
+
+func detectGlobalnet() {
+	submarinerClient, err := submarinerclientset.NewForConfig(framework.RestConfigs[framework.ClusterA])
+	exitWithErrorMsg(fmt.Sprintf("Error creating submariner client: %v", err))
```
mangelajo, line 274:
```suggestion
	exitOnError("Error creating submariner client", err)
```

pkg/subctl/cmd/verify.go
```diff
@@ -264,3 +268,14 @@ func getVerifyPatterns(csv string, includeDisruptive bool) ([]string, []string,
...
+
+	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName,
+		v1.GetOptions{})
+	exitWithErrorMsg(fmt.Sprintf("Error obtaining Submariner resource: %v", err))
```
mangelajo, line 278:
```suggestion
	exitOnError("Error obtaining Submariner resource", err))
```

mangelajo:
Thanks tom

tpantelis:
> This means `subctl verify` can only be run against operator deployments. Which should be fine, but need to add a check and warning/error msg somewhere if someone runs it against a non-operator deployment. Or at least capture it in documentation. Maybe raise an issue that we can close once helm deployments move to operator underneath.

`subctl` is part of the operator project and the most important part of it, `join`, relies on the operator so I think it's reasonable for `verify` to do so. It'll fail if the `Submariner` object isn't found. I can improve the error message to make it clear that the operator must be deployed and helm deployment is not supported.

tpantelis:
> I can improve the error message to make it clear that the operator must be deployed and helm deployment is not supported.

I updated the error message via https://github.com/submariner-io/submariner-operator/pull/566/commits/3e375a67f93a8e3bde685070f8c40cf0f0b402bd

Issue #468, opened 2020-06-17
subctl show cmd: add subctl show gateways


maayanf24:
This is sub-issue of #359

See enhancement: https://github.com/submariner-io/enhancements/blob/master/submariner-operator/issue-359-subctl-info-status.md

PR #570, opened 2020-07-13
Add subctl cmd show gateways

maayanf24:
Resolve issue #468

### DIFF
--- a/pkg/subctl/cmd/show_gateways.go
+++ b/pkg/subctl/cmd/show_gateways.go
@@ -11,10 +11,9 @@ import (
 )
 
 type gatewayStatus struct {
-	node        string
-	haStatus    submv1.HAStatus
-	connections string
-	summary     string
+	node     string
+	haStatus submv1.HAStatus
+	summary  string
 }
 
 var showGatewaysCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/show_gateways.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+
+		var summary string
+		if totalConnections > 0 && totalConnections == countConnected {
+			summary = "All connections OK"
+		} else {
+			summary = gateway.Status.StatusFailure
+		}
+		connectionString := fmt.Sprintf("%d/%d", countConnected, totalConnections)
```
tpantelis, line 75: `Status.StatusFailure`  (currently) is populated when there is a failure obtaining the connections in the driver. I think the connection string and the summary somewhat overlap plus if there's partial connectivity then the summary would be empty. I propose we simplify and combine the connection string with the summary:

```suggestion
                if gateway.Status.StatusFailure != "" {
                    summary = gateway.Status.StatusFailure
                } else if totalConnections == 0 {
                    summary = "There are no connections"
                } else if totalConnections == countConnected {
			summary = fmt.Sprintf("All %d connections are established",  totalConnections)
		} else {
			summary = fmt.Sprintf("%d connections out of %d are established",  totalConnections, countConnected)
		}
	
```

pkg/subctl/cmd/show_gateways.go
```diff
@@ -0,0 +1,99 @@
...
+
+func showGateways(cmd *cobra.Command, args []string) {
+	var status []gatewaysStatus
+	status = getGatewaysStatus(status)
+	printGateways(status)
+}
+
+func printGateways(gateways []gatewaysStatus) {
+	template := "%-20s%-16s%-16s%-32s\n"
+	fmt.Printf(template, "NODE", "HA-STATUS", "CONNSCTIONS", "SUMMARY")
```
tpantelis, line 90:
```suggestion
	fmt.Printf(template, "NODE", "HA STATUS", "CONNECTIONS", "SUMMARY")
```

pkg/subctl/cmd/show_gateways.go
maayanf24: in case there will be only 1 connection, the summary will show: `All 1 connections are established`

tpantelis: yeah good point. How about `All connections (%d) are established`?

maayanf24: sounds good.

maayanf24: shell I keep the `CONNECTIONS` column? 

```
[maafried@localhost submariner-operator]$ ./bin/subctl show gateways
NODE                HA STATUS       CONNECTIONS     SUMMARY                         
cluster2-worker     active          1/1             All connections (1) are established
```

tpantelis: no I think it's redundant.

PR #579, opened 2020-07-13
Enable unparam unused param linting, fix errors

dfarrell07:
Builds on https://github.com/submariner-io/submariner-operator/pull/581

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -47,7 +47,7 @@ linters:
     #- testpackage
     - typecheck
     - unconvert
-    #- unparam
+    - unparam
     - unused
     - varcheck
     - whitespace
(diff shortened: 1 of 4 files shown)


pkg/discovery/network/network.go
```diff
@@ -95,24 +95,24 @@ func Discover(dynClient dynamic.Interface, clientSet kubernetes.Interface, submC
...
 	}
 
 	canalClusterNet, err := discoverCanalFlannelNetwork(clientSet)
 	if err == nil && canalClusterNet != nil {
-		return canalClusterNet, nil
+		return canalClusterNet
 	}
 
-	return nil, nil
+	return nil
```
tpantelis, line 115: Given that the code always returned a nil error, it seems reasonable to remove the return value. OTOH, if an error occurred that prevented the discovery of a technology, eg a k8s API server failure, then I think it should return the error and the caller should bail. So a non-nil error means no determination could be made, otherwise a nil `ClusterNetwork` means it definitively wasn't found. But I think that's outside the scope of this PR - let's just fix the linting errors as simply as we can for now.

tpantelis: I pushed https://github.com/submariner-io/submariner-operator/pull/581 to refactor the error handling which will also fix the linting error.

dfarrell07: Awesome, thanks @tpantelis! I was nervous we'd end up losing this great feedback about the better solution. I'll enable the linter on top of that better-fix.

PR #581, opened 2020-07-13
Refactor/improve network discovery error handling

tpantelis:
In `networkPluginsDiscovery`, if an error occurs that prevents the discovery of a technology, eg a k8s API server failure, then it should return the error. Therefore a non-nil error means no determination could be made and callers should bail, otherwise a nil `ClusterNetwork` means it definitively wasn't found.

Also ensured the OpenShift4, weave, and canal network disovery methods follow the same pattern, specifically return `nil, nil` if the target resource isn't found.

This also addresses the issue in https://github.com/submariner-io/submariner-operator/pull/503 and the linting errors in https://github.com/submariner-io/submariner-operator/pull/579


mangelajo:
Wait wait, let's not merge this yet, there are corner cases that won't work

pkg/discovery/network/openshift4.go
```diff
@@ -43,8 +44,10 @@ func discoverOpenShift4Network(dynClient dynamic.Interface) (*ClusterNetwork, er
 
 	cr, err := crClient.Get("default", metav1.GetOptions{})
 	if err != nil {
-		klog.Info("Attempted network discovery for OpenShift4, no clusternetworks CRD")
-		return nil, nil
+		if apierrors.IsNotFound(err) {
```
mangelajo, line 47: I think that missing CRD could be another type of error we may need to check.

pkg/discovery/network/pods.go
```diff
@@ -55,7 +56,7 @@ func findPod(clientSet kubernetes.Interface, labelSelector string) (*v1.Pod, err
 	})
 
 	if err != nil || len(pods.Items) == 0 {
-		return nil, err
+		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", labelSelector)
```
mangelajo, line 59: if len(pod.Items)==0  we should not return error, in the previous code "err" could still be nil in this case.

mangelajo:
This handles https://github.com/submariner-io/submariner-operator/pull/503 in a better way, please close that other PR when we merge this one.

pkg/discovery/network/openshift4.go
tpantelis: Good point - I'll check on that.

pkg/discovery/network/pods.go
tpantelis: Sorry not sure I follow. I didn't change the logic. If `len(pod.Items)==0 ` then `err` is nil.

pkg/discovery/network/openshift4.go
tpantelis: I verified that it still returns `IsNotFound` error if the resource type doesn't exit. That's also evident b/c all the e2e tests succeeded :smile:

pkg/discovery/network/network.go
```diff
@@ -62,53 +62,57 @@ func (cn *ClusterNetwork) IsComplete() bool {
...
+
+	// If nothing specific was discovered, use the generic discovery
+	return discoverGenericNetwork(clientSet)
 }
 
 func networkPluginsDiscovery(dynClient dynamic.Interface, clientSet kubernetes.Interface) (*ClusterNetwork, error) {
 	osClusterNet, err := discoverOpenShift4Network(dynClient)
-	if err == nil && osClusterNet != nil {
-		return osClusterNet, nil
+	if err != nil || osClusterNet != nil {
```
skitt, line 104: This changes the behaviour significantly: if checking for OCP4 causes an error, we’ll give up instead of checking the remaining networks.

pkg/discovery/network/pods.go
skitt: @mangelajo if `err` is `nil`, `errors.WithMessagef()` returns `nil` too, so the behaviour doesn’t change here (but the code can be confusing).

pkg/discovery/network/network.go
skitt: ... and that’s the whole point of this PR. Move along, nothing to see here!

pkg/discovery/network/pods.go
mangelajo: ahh ok I didn't know that bit @skitt explains thanks.

PR #585, opened 2020-07-14
Enable lll line length linting, fix errors

dfarrell07:


dfarrell07:
It seems like the embedded YAML auto-generation is causing long lines for lines with links.

Issue #601, opened 2020-07-28
subctl should check the K8s version


skitt:
We’ll need 1.16 (for `iptables` changes being worked on) or even 1.17 (for endpoint slices), `subctl` should check that.

PR #602, opened 2020-07-28
Check the K8s version on the cluster

skitt:
For endpoint slices we need 1.17, enforce that before deploying.

Fixes: #601

### DIFF
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -18,12 +18,14 @@ package cmd
 
 import (
 	"encoding/base64"
-	"errors"
 	"fmt"
+	"os"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
@@ -94,6 +96,8 @@ func addJoinFlags(cmd *cobra.Command) {
 
 const (
 	SubmarinerNamespace = "submariner-operator" // We currently expect everything in submariner-operator
+	minK8sMajor         = 1                     // We need K8s 1.17 for endpoint slices
+	minK8sMinor         = 17
 )
 
 var joinCmd = &cobra.Command{
@@ -176,6 +180,18 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, subctlData *datafile.S
 	clientConfig, err := config.ClientConfig()
 	exitOnError("Error connecting to the target cluster", err)
 
+	failedRequirements, err := checkRequirements(clientConfig)
+	// We display failed requirements even if an error occurred
+	if len(failedRequirements) > 0 {
+		fmt.Println("The target cluster fails to meet Submariner's requirements:")
+		for i := range failedRequirements {
+			fmt.Printf("* %s\n", (failedRequirements)[i])
+		}
+		exitOnError("Unable to check all requirements", err)
+		os.Exit(1)
+	}
+	exitOnError("Unable to check requirements", err)
+
 	if !noLabel {
 		err := handleNodeLabels(clientConfig)
 		exitOnError("Unable to set the gateway node up", err)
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -166,12 +169,14 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 		}
 	}
 
+	err := checkRequirements(config)
+	exitOnError("The cluster doesn't meet our requirements", err)
```
tpantelis, line 173: Minor thing but the message here is redundant with the unsupported version message  and wouldn't apply to the other API failure messages. I suggest using `exitWithErrorMsg` and maybe clarifying the API failure messages.

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +226,29 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
 	exitOnError("Error deploying Submariner", err)
 }
 
+func checkRequirements(config *rest.Config) error {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
```
tpantelis, line 232:
```suggestion
		return errors.WithMessage(err, "Error creating API server client")
```

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +226,29 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	}
+	serverVersion, err := clientset.Discovery().ServerVersion()
+	if err != nil {
+		return err
```
tpantelis, line 236:
```suggestion
		return errors.WithMessage(err, "Error obtaining API server version")
```

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +226,29 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	}
+	major, err := strconv.Atoi(serverVersion.Major)
+	if err != nil {
+		return err
```
tpantelis, line 240:
```suggestion
		return errors.WithMessage(err, "Error converting API server major version")
```

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +226,29 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	}
+	minor, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		return err
```
tpantelis, line 244:
```suggestion
		return errors.WithMessage(err, "Error converting API server minor version")
```

pkg/subctl/cmd/join.go
skitt: Done, thanks!

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +226,30 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	if err != nil {
+		return errors.WithMessagef(err, "error parsing API server major version %v", serverVersion.Major)
+	}
+	minor, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		return errors.WithMessagef(err, "error parsing API server minor version %v", serverVersion.Minor)
+	}
+	if major < minK8sMajor || (major == minK8sMajor && minor < minK8sMinor) {
+		exitWithErrorMsg(fmt.Sprintf("Submariner requires Kubernetes %d.%d; your cluster is running %s.%s",
+			minK8sMajor, minK8sMinor, serverVersion.Major, serverVersion.Minor))
```
tpantelis, line 248: Not trying to nitpick but I think we should be consistent with the other code paths (ie L232, L236 etc) and either return the error here or `exitWithErrorMsg` in the other code paths and don't return error. The latter is simpler.

skitt: Right, I misunderstood your initial comment about `exitWithErrorMsg` ;-).

skitt: I think the mistake here is trying to use errors for both technical errors and requirement errors. The requirements check should return something else, and use errors for only technical errors.

tpantelis: yeah that makes sense - use regular output for requirement errors.

pkg/subctl/cmd/join.go
```diff
@@ -221,6 +234,32 @@ func joinSubmarinerCluster(config *rest.Config, subctlData *datafile.SubctlData)
...
+	minor, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		return &failedRequirements, errors.WithMessagef(err, "error parsing API server minor version %v", serverVersion.Minor)
+	}
+	if major < minK8sMajor || (major == minK8sMajor && minor < minK8sMinor) {
+		failedRequirements = append(failedRequirements,
+			fmt.Sprintf("Submariner requires Kubernetes %d.%d; your cluster is running %s.%s",
+				minK8sMajor, minK8sMinor, serverVersion.Major, serverVersion.Minor))
+	}
+	return &failedRequirements, nil
```
tpantelis, line 260: Why not return `[]string` for failed requirements and nil for success, ie

     
```
        clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		return nil, errors.WithMessage(err, "error creating API server client")
	}
	 ...
        var failedRequirements []string
	if major < minK8sMajor || (major == minK8sMajor && minor < minK8sMinor) {
		failedRequirements = append(failedRequirements,
			fmt.Sprintf("Submariner requires Kubernetes %d.%d; your cluster is running %s.%s",
				minK8sMajor, minK8sMinor, serverVersion.Major, serverVersion.Minor))
	}
	return failedRequirements, nil
```

skitt: I wanted to start a pattern where, if additional requirements are added in the future, if an error occurs we can still show the missing requirements we’ve identified so far (which sounds weird from a Java background...).

tpantelis: Sorry I'm not sure I follow, ie I'm not seeing the reason for returning `*[]string` instead of `[]string`.  Either way it's not a big deal.

tpantelis:
I somehow hit the Close button by mistake. It doesn't ask for confirmation.

PR #608, opened 2020-07-30
If OCP4 network discovery fails, log the error

skitt:
Currently if OCP4 network discovery fails, the details of the error
aren’t logged, which complicates debugging related issues on OCP4.

This patch ensures the error is logged and returns the error (which
doesn’t change the caller’s behaviour).


skitt:
#581 does this better.

PR #612, opened 2020-08-02
Bump lighthouse

mkolesnik:
To unbreak the e2e tests since the test code has changed.


mkolesnik:
Not sure if this is the desired way to go with it, but LH had changes in the way it tests E2E (and probably deploys due to the whole service export introduction) which aren't in v0.5.0 tag

Issue #603, opened 2020-07-29
subctl verify should validate if kubeconfigs point to same cluster


sridhargaddam:
While running subctl verify if the user accidentally supplies the same kubeconfig as kubeConfig1 and kubeConfig2, we should return an appropriate error message instead of showing 

...
Ran 16 of 22 Specs in 0.347 seconds
**FAIL! -- 0 Passed | 16 Failed | 0 Pending | 6 Skipped**
[] E2E failed

**Sample command to reproduce the error message:**
subctl verify output/kubeconfigs/kind-config-cluster3 output/kubeconfigs/kind-config-cluster3 --only connectivity

PR #614, opened 2020-08-04
Check that kubeconfig files are not the same file in subctl verify

maayanf24:
Fix issue #603


pkg/subctl/cmd/verify.go
```diff
@@ -164,6 +164,9 @@ func checkValidateArguments(args []string) error {
 	if len(args) != 2 {
 		return fmt.Errorf("Two kubeconfigs must be specified.")
 	}
+	if strings.Compare(args[0], args[1]) == 0 {
```
sridhargaddam, line 167: This can be bypassed if the user gives full-path to one of the kubeconfigs and just the filename (or relative path) to the other. To make it fool-proof, shall we compare the server IP address from the kubeconfigs?

@mkolesnik , @mangelajo @skitt @nyechiel thoughts?

skitt: Or even if two kubeconfigs have the same contents. I think the best way to check this would be to check the `rest.Config` instances we load; but that might end up being much more complex than checking the files.

maayanf24: @skitt - should the `rest.Config` be checked at the testing framework?
Or I can load the `rest.Config` in the `verify` cmd, inside `checkValidateArguments` ?

skitt: I would check the configs in the test framework; otherwise, if the test framework changes in a way that would allow testing inside a single cluster, we’d also have to change `subctl` to match.

mkolesnik: Makes sense to compare the contents as well, but let's not try to get too carried away.
I think 1st check if the filename is the same (simple quick check) and if it isn't then check if the contents are the same.

If it were a security concern it'd make sense to check more thoroughly, but for sake of simplicity I think compare file names and then contents will give us enough protection from user mistakes.

skitt (approved):

mangelajo (approved):
Seems ok for now, we may need to revisit this, and add support for using KUBECONFIG settings and contexts, etc..

Issue #539, opened 2020-07-06
Generate a cluster id if one is not provided


manosnoam:
Currently we expect users to modify their kubeconfig context, before running subctl join with their chosen clusterId, otherwise the default openshift cluster context will stay "admin" on all clusters' kubeconfigs.

@mangelajo, @skitt
Can you set clusterId with a constant identifier of current cluster context, with some API call (instead of generating a random UUID that beeing hard to trace later) ?

This is related to:
https://github.com/submariner-io/submariner-operator/pull/8

manosnoam:
@skitt thanks for adding a fix.
I will close this issue after verifying the fix.

skitt:
Re-opening, Noam needs to verify this.

manosnoam:
This is working great now on latest devel. 
No need to specify `--clusterid` for the join command:
'oc describe Submariner' shows for example: `Cluster ID:                 default-cl1`

skitt:
This was implemented in 0.6.0, please don’t change the milestone. I’m not sure what to do about the project though ;-).

PR #616, opened 2020-08-04
Use the cluster name as cluster id by default

skitt:
If the user doesn't specify a cluster id on the command line, use the
cluster name from the kubeconfig by default. In both cases, the name
is checked for validity and the user is involved if necessary.

The kubeconfig cluster name is guaranteed to be unique (otherwise we
wouldn't be able to connect).

Fixes: #539


pkg/subctl/cmd/root.go
```diff
@@ -100,6 +101,14 @@ func getClients(config *rest.Config) (dynamic.Interface, kubernetes.Interface, e
 	return dynClient, clientSet, nil
 }
 
+func getClusterName(rawConfig clientcmdapi.Config) *string {
+	context, ok := rawConfig.Contexts[rawConfig.CurrentContext]
```
mangelajo, line 105: I wonder if we may check with the broker that we are not repeating the cluster ID (separate PR).

mangelajo: I was thinking... let's add this to the operator... but ... it's not possible. We don't have access to the context name at that time.

Issue #604, opened 2020-07-29
Missing info in subctl export service help section


sridhargaddam:
Currently, when `"subctl export service -h"` command is executed, it's not showing the mandatory serviceName as one of the parameters. This needs to be fixed.

PR #619, opened 2020-08-05
Add serviceName to subctl export service help

maayanf24:
Fix issue #604

### DIFF
--- a/pkg/subctl/cmd/export.go
+++ b/pkg/subctl/cmd/export.go
@@ -35,7 +35,7 @@ var (
 		Long:  "This command exports a resource so it is accessible to other clusters",
 	}
 	exportServiceCmd = &cobra.Command{
-		Use:   "service",
+		Use:   "service <serviceName>",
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/export.go
```diff
@@ -35,7 +35,7 @@ var (
 		Long:  "This command exports a resource so it is accessible to other clusters",
 	}
 	exportServiceCmd = &cobra.Command{
-		Use:   "service",
+		Use:   "service <serviceName>",
 		Short: "Exports a Service to other clusters",
 		Long:  "This command creates a ServiceExport resource with a given name which causes the Service of the same name to be accessible to other clusters",
```
skitt, line 40:
```suggestion
		Long:  "This command creates a ServiceExport resource with the given name which causes the Service of the same name to be accessible to other clusters",
```

Now that we specify the name, we can refer to it.

tpantelis:
Need to add `Signed-Off-By` to the commits (and might as well rebase).

PR #621, opened 2020-08-06
Print 'no resource found' in show cmds

maayanf24:
Fix issue https://github.com/submariner-io/enhancements/issues/5

### DIFF
--- a/pkg/subctl/cmd/show_connections.go
+++ b/pkg/subctl/cmd/show_connections.go
@@ -83,8 +83,12 @@ func showConnectionsFromConfig(config *rest.Config) {
 }
 
 func printConnections(connections []connectionStatus) {
-	template := "%-20s%-16s%-16s%-24s%-40s%-16s\n"
+	if len(connections) == 0 {
+		fmt.Println("No resources found.")
+		return
+	}
 
+	template := "%-20s%-16s%-16s%-24s%-40s%-16s\n"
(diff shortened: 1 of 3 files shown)


pkg/subctl/cmd/show_connections.go
```diff
@@ -73,9 +73,13 @@ func showConnections(cmd *cobra.Command, args []string) {
 
 func printConnections(connections []connectionStatus) {
 	template := "%-20s%-16s%-16s%-24s%-40s%-16s\n"
-
 	fmt.Printf(template, "GATEWAY", "CLUSTER", "REMOTE IP", "CABLE DRIVER", "SUBNETS", "STATUS")
 
+	if len(connections) == 0 {
+		fmt.Println("no resource found")
```
skitt, line 79: To match `kubectl`,

```suggestion
		fmt.Println("No resources found.")
```

(and likewise for the other two occurrences).

maayanf24: One more question - should the header line be printed in or not in the case of no resources?

skitt: > One more question - should the header line be printed in or not in the case of no resources?

`kubectl` doesn’t print the header, so no (see `kubectl get pods -n default` on a newly-deployed cluster for example).

skitt:
@maayanf24 the conflicts need to be resolved manually.

PR #622, opened 2020-08-10
Increase field size in subctl show connections and show endpoints cmds

maayanf24:


maayanf24:
> Thanks @maayanf24 for fixing the alignment issue.
> 
> Can you also take a look at "subctl show endpoints" which has a similar problem.

@sridhargaddam Thanks, I will add this as well

tpantelis:
Still conflicts - I think you'll need to manually squash.

Issue #516, opened 2020-07-01
Operator restarting the Gateway Pods


sridhargaddam:
While looking at CI logs, it was seen that the operator was restarting the GatewayPod sometimes as part of reconcileEngineDaemonSet. Ideally we would want the Pods to be restarted only if there are any config changes and not otherwise. This PR is to ensure that code is handling this case properly.

```

func (r *ReconcileSubmariner) reconcileEngineDaemonSet(instance *submopv1a1.Submariner, reqLogger logr.Logger) (*appsv1.DaemonSet, error) {
	daemonSet, err := helpers.ReconcileDaemonSet(instance, newEngineDaemonSet(instance), reqLogger, r.client, r.scheme)
	if err == nil {
		err = r.deletePreExistingEngineDeployment(instance.Namespace, reqLogger)
	}
	return daemonSet, err
}
func (r *ReconcileSubmariner) reconcileRouteagentDaemonSet(instance *submopv1a1.Submariner, reqLogger logr.Logger) (*appsv1.DaemonSet, error) {
	return helpers.ReconcileDaemonSet(instance, newRouteAgentDaemonSet(instance), reqLogger, r.client, r.scheme)
}
func (r *ReconcileSubmariner) reconcileGlobalnetDaemonSet(instance *submopv1a1.Submariner, reqLogger logr.Logger) (*appsv1.DaemonSet, error) {
	return helpers.ReconcileDaemonSet(instance, newGlobalnetDaemonSet(instance), reqLogger, r.client, r.scheme)
}
```

sridhargaddam:
Had a look at the Operator code and also did some testing where I modified different values in "submariner" CR. I noticed that EnginePod was getting restarted only when there is any change to the corresponding params and not otherwise. 

There was a small optimization that could be done, so proposed the following PR - https://github.com/submariner-io/submariner-operator/pull/626/files

mangelajo:
@sridhargaddam can we close this one? Looks like it's handled :)

sridhargaddam:
> @sridhargaddam can we close this one? Looks like it's handled :)

Yes @mangelajo we can close it. Thanks

PR #626, opened 2020-08-11
Avoid checking for EngineDeployment during reconcileEngineDaemonSet

sridhargaddam:
During the initial releases of Submariner, submariner-gateway used to
be a Deployment. However, it was changed to a DaemonSet long ago and
many releases happened after that change. So we can remove the check
for deployment.

Related to: https://github.com/submariner-io/submariner-operator/issues/516


skitt:
You need to also remove the corresponding unit test ;-).

PR #643, opened 2020-08-21
Remove getsubctl.sh, moved to get.submariner.io

mangelajo:
From now on getsubctl.sh will be maintained in
https://github.com/submariner-io/get.submariner.io

This commit will break users of the script with the
old URL.


mangelajo:
> Hi @mangelajo, @tpantelis indeed this breaks using the old link from the Website docs.
> It would have been great if there was an issue to verify, not just this fix - that is already closed now, before it is verified.
> Did you also fix the Website instructions ?

I updated the website and I verified the instructions work.

PR #645, opened 2020-08-24
Retain supercluster in dns configmap

vthapar:
MCS API now uses `clusterset.local` instead of `supercluster.local` as
zone for multicluster service discovery. To support any existing
deployments, support `supercluster` for at least one release to allow
users to switch to `clusterset`

### DIFF
--- a/pkg/controller/servicediscovery/servicediscovery_controller.go
+++ b/pkg/controller/servicediscovery/servicediscovery_controller.go
@@ -236,6 +236,12 @@ errors
 health
 ready
 }
+supercluster.local:53 {
+lighthouse
+errors
+health
+ready
+}
 `
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


skitt (approved):

mangelajo (approved):

pkg/controller/servicediscovery/servicediscovery_controller.go
```diff
@@ -374,6 +383,9 @@ func updateDNSConfigMap(client controllerClient.Client, k8sclientSet *clientset.
 clusterset.local:53 {
 forward . `
 		expectedCorefile = expectedCorefile + lighthouseDnsService.Spec.ClusterIP + "\n" + "}\n"
+		superclusterCorefile := `clusterset.local:53 {
```
aswinsuryan, line 386: 'supercluster'?

vthapar: Argh! yes.

mangelajo (approved):

pkg/controller/servicediscovery/servicediscovery_controller.go
mangelajo: Ooops good catch :)

PR #648, opened 2020-08-25
Enable ServiceDiscovery by default on subctl

mangelajo:
This patch enables service discovery on deployments by default,
now that we implement the upstream API and we have a stable implementation
it's a good time to make this feature enabled by default.

### DIFF
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -49,8 +49,8 @@ func init() {
 	deployBroker.PersistentFlags().StringVar(&ipsecSubmFile, "ipsec-psk-from", "",
 		"Import IPsec PSK from existing submariner broker file, like broker-info.subm")
 
-	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", false,
-		"Enable Multi Cluster Service Discovery")
+	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", true,
+		"Enable Multi-Cluster Service Discovery")
 
 	addKubeconfigFlag(deployBroker)
 	rootCmd.AddCommand(deployBroker)


pkg/subctl/cmd/deploybroker.go
```diff
@@ -49,7 +49,15 @@ func init() {
 	deployBroker.PersistentFlags().StringVar(&ipsecSubmFile, "ipsec-psk-from", "",
 		"Import IPsec PSK from existing submariner broker file, like broker-info.subm")
 
-	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", false,
+	deployBroker.PersistentFlags().Bool("service-discovery", true,
+		"Enable Multi Cluster Service Discovery")
```
sridhargaddam, line 53:
```suggestion
	deployBroker.PersistentFlags().Bool("service-discovery", true,
		"Enable Multi-Cluster Service Discovery")
```

pkg/subctl/cmd/deploybroker.go
```diff
@@ -49,7 +49,15 @@ func init() {
...
+
+	err := deployBroker.Flags().MarkDeprecated("service-discovery",
```
sridhargaddam, line 55: Just a thought: Instead of deprecating the flag "service-discovery", why not just modify the default from false to true?

mangelajo: It's a cmdline flag :-) then you can't disable it, how do you disable it with --service-discovery anymore? :)

pkg/subctl/cmd/deploybroker.go
mangelajo: :+1: wasn't my code/change though.

pkg/subctl/cmd/deploybroker.go
vthapar: You can do `--service-discovery=false`

mangelajo: Not sure why my comment didn't go through.

We can only deprecate it because

subctl --service-discovery being called doesn't make sense anymore (is a bool flag, does not accept true/false afterwards,...)

sridhargaddam (approved):

pkg/subctl/cmd/deploybroker.go
skitt: As Vishal says, `--service-discovery` is still useful to disable service discovery; boolean flags [support a value after an equals symbol](https://github.com/spf13/pflag/#command-line-flag-syntax).

mangelajo: > You can do `--service-discovery=false`

can you? ? let me try

pkg/subctl/cmd/deploybroker.go
```diff
@@ -49,7 +49,15 @@ func init() {
 	deployBroker.PersistentFlags().StringVar(&ipsecSubmFile, "ipsec-psk-from", "",
 		"Import IPsec PSK from existing submariner broker file, like broker-info.subm")
 
-	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", false,
+	deployBroker.PersistentFlags().Bool("service-discovery", true,
+		"Enable Multi-Cluster Service Discovery")
```
skitt, line 53: So we’re dropping support for `--service-discovery=false`?

mangelajo: I'm changing it back to that... another patch in a minute.

skitt (approved):

sridhargaddam (approved):
thanks for letting us know that we can use values for boolean flags @skitt

mangelajo:
Ok this change was agreed to make sense from the user-facing perspective. Although it will deploy lighthouse on configurations we are testing where lighthouse is not needed, hopefully lightweight enough. It was agreed also that we didn't want to complicate shipyard or other CI scripts to be aware of --service-discover=false beyond the operator repository (where we are testing subctl)

PR #650, opened 2020-08-25
Enable gosec linting, fix errors

dfarrell07:

### DIFF
--- a/pkg/subctl/datafile/datafile.go
+++ b/pkg/subctl/datafile/datafile.go
@@ -68,7 +68,7 @@ func (data *SubctlData) WriteToFile(filename string) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/subctl/datafile/datafile.go
```diff
@@ -68,7 +68,7 @@ func (data *SubctlData) WriteToFile(filename string) error {
 		return err
 	}
 
-	if err := ioutil.WriteFile(filename, []byte(dataStr), 0644); err != nil {
```
dfarrell07, line 71: Is this going to cause some undesired side-effect not tested in CI?

skitt: The only difference is that the file won’t be readable by anyone but the user who ran `subctl`, which shouldn’t be an issue in practice.

skitt (approved):

PR #652, opened 2020-08-26
Bump MD link check linter 0.6->1.*

dfarrell07:
Adds support for check-modified-files-only config, which currently isn't
honored and throws a warning.

### DIFF
--- a/.github/workflows/md-links-periodic.yml
+++ b/.github/workflows/md-links-periodic.yml
@@ -12,7 +12,7 @@ jobs:
       uses: actions/checkout@v1
 
     - name: Run markdown-link-check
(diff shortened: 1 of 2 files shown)


dfarrell07:
It's weird that only @mkolesnik was added as a code owner.

skitt (approved):

dfarrell07:
> It's weird that only @mkolesnik was added as a code owner.

It's because of:

```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
```

https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax

Issue #680, opened 2020-09-07
subctl verify - use clustername from client config


vthapar:
`subctl verify` has clusterIDs hard coded as follows:
```	
        // For some tests this is only printing, but in some of them they need those to be
	// the cluster IDs that will be registered in the Cluster CRDs by submariner
	framework.TestContext.ClusterIDs = []string{"ClusterA", "ClusterB"}
```
This makes ClusterIDs inaccurate to use when writing e2e tests for `verify` It should instead read from client config similar to the way `subctl join` does.

PR #681, opened 2020-09-07
Use clusternames from client config in verify

skitt:
Instead of hard-coding cluster names, pick up the names from the
client configuration, in the same way as join.

Fixes: #680

### DIFF
--- a/pkg/subctl/cmd/verify.go
+++ b/pkg/subctl/cmd/verify.go
@@ -152,12 +152,22 @@ func configureTestingFramework(args []string) {
 	framework.TestContext.ReportPrefix = "subctl"
 	framework.TestContext.SubmarinerNamespace = submarinerNamespace
 
-	// For some tests this is only printing, but in some of them they need those to be
-	// the cluster IDs that will be registered in the Cluster CRDs by submariner
-	framework.TestContext.ClusterIDs = []string{"ClusterA", "ClusterB"}
+	// Read the cluster names from the given kubeconfigs
+	framework.TestContext.ClusterIDs = []string{clusterNameFromConfig(args[0]), clusterNameFromConfig(args[1])}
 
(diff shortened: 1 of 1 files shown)


vthapar (approved):

pkg/subctl/cmd/verify.go
```diff
@@ -152,9 +152,12 @@ func configureTestingFramework(args []string) {
...
 
-	// For some tests this is only printing, but in some of them they need those to be
-	// the cluster IDs that will be registered in the Cluster CRDs by submariner
-	framework.TestContext.ClusterIDs = []string{"ClusterA", "ClusterB"}
+	// Read the cluster names from the given kubeconfigs
+	rawConfigA, err := getClientConfig(args[0], "").RawConfig()
+	exitOnError("Error connecting to the first target cluster", err)
+	rawConfigB, err := getClientConfig(args[1], "").RawConfig()
+	exitOnError("Error connecting to the second target cluster", err)
+	framework.TestContext.ClusterIDs = []string{*getClusterName(rawConfigA), *getClusterName(rawConfigB)}
```
tpantelis, line 160: Perhaps we should check nil return from `getClusterName` to be complete? ie:

```
func clusterFromConfig(kubeConfigPath string) string {
    rawConfig, err := getClientConfig(kubeConfigPath, "").RawConfig()
    exitOnError(fmt.Sprintf("Error obtaining the kube config for path %q", kubeConfigPath), err)
    cluster := getClusterName(rawConfig) 
    if cluster == nil {
        exitWithErrorMsg(fmt.Sprintf("Could not obtain the cluster name from kube config: %#v", rawConfig))
    }

    return *cluster
} 
```

skitt: Done, thanks!

tpantelis (approved):

Issue #678, opened 2020-09-07
Migrate to Operator SDK 1.0


skitt:
We need to migrate the operator to use version 1.0 of the SDK. This will bring a number of benefits, not least of which is using a “supported” stable SDK: better kubebuilder integration (so we can generate RBAC from code, in particular), flat bundle generation (for the CSVs)...

SteveMattar:
+1

nyechiel:
Instructions for migrating a legacy Go-based project to use the new Kubebuilder-style: layout.https://master.sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide/

skitt:
Oops, sorry @SteveMattar, I didn’t think removing your comment would also unassign you from the issue!

SteveMattar:
we should open an issue on that :)

nyechiel:
bump

PR #682, opened 2020-09-08
refactor: flatten the deploy directory structure and add the missing lighthouse resources

SteveMattar:
* flatten the deploy directory structure
* rename the globalnet service account
* update the cr and add the missing ones for lighthouse
* add the missing lighthouse api & controller
* modify the validations test according the changes above

Signed-off-by: Steve Mattar smattar@redhat.com


pkg/apis/lighthouse/v2alpha1/serviceimport_types.go
```diff
@@ -0,0 +1,116 @@
+package v2alpha1
```
mangelajo, line 1: We cannot add the lighthouse APIs in our tree.

That lives on lighthouse.

I understand it's just to circunvent the limitation of the operator-sdk, and I understand if that's a step that we do during generation, I believe we shouldn't add them on the git tree.

@skitt  ^ thoughts?

pkg/controller/serviceimport/serviceimport_controller.go
```diff
@@ -0,0 +1,153 @@
+package serviceimport
```
mangelajo, line 1: Neither a controller, etc...

Nothing that is "Submariner" or "ServiceDiscovery" CRDs API should live in this repo.

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -32,18 +32,17 @@ const (
 var files = []string{
 	"crds/submariner.io_submariners_crd.yaml",
 	"crds/submariner.io_servicediscoveries_crd.yaml",
+	"crds/lighthouse.submariner.io_serviceexports_crd.yaml",
+	"crds/lighthouse.submariner.io_serviceimports_crd.yaml",
+	"cluster_role.yaml",
+	"cluster_role_binding.yaml",
+	"globalnet_cluster_role.yaml",
```
mangelajo, line 39: can we prefix those files with submariner_... (the ones which were in the submariner dir)?

submariner directory meant (handled by submariner, not the operator)
lighthouse directory meant (handled by lighthouse, not the operator)

We just have them here to be able to install the CRDs and do the deploy.

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -32,18 +32,17 @@ const (
...
+	"globalnet_cluster_role_binding.yaml",
```
mangelajo, line 40: same for this one :)

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
SteveMattar: I changed the names because I removed the directory.
do you mean the operator should be aware of these definitions?
I made this change so that when generating the csv those definitions would be included as well.

SteveMattar:
depends on #683

pkg/apis/lighthouse/v2alpha1/serviceimport_types.go
skitt: Yup, I’m hoping we can use the vendored dependencies to drive the generation somehow... Or Kubebuilder once we’ve upgraded to Operator SDK v1.

SteveMattar: @mangelajo @skitt should I close this PR for now? I will start working on #678 and then we can have this discussion again

skitt: Yes, I suspect that’s the most sensible option — migrating to the Operator SDK involves re-creating the project anyway, and will resultin a flattened structure. Note too that I’m migrating most CRDs to the operator itself in https://github.com/submariner-io/submariner-operator/pull/685 which will solve a few issues for you ;-).

SteveMattar: sounds good

PR #684, opened 2020-09-08
Install jq in the dapper Dockerfile

mkolesnik:
This project is the only one using the `jq` tool, hence it makes sense
to install directly here instead of in the Shipyard base image.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -4,11 +4,15 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} \
     DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_SHA MAKEFLAGS CLUSTERS_ARGS DEPLOY_ARGS E2E_ARGS RELEASE_ARGS" \
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner-operator DAPPER_DOCKER_SOCKET=true \
-    OPERATOR_SDK_VERSION=0.12.0 GOROOT=/usr/lib/golang
+    GOROOT=/usr/lib/golang \
+    JQ_VERSION=1.6 \
+    OPERATOR_SDK_VERSION=0.12.0
 ENV DAPPER_OUTPUT=${DAPPER_SOURCE}/output PATH=${DAPPER_SOURCE}/bin/:${PATH}
 
 RUN curl -Lo /usr/bin/operator-sdk "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu" && \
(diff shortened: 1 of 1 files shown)


skitt (changes requested):
Any particular reason not to use the packaged version?

mkolesnik:
> Any particular reason not to use the packaged version?

The main reason is that the whole `dnf install & clean up` step just takes a lot more time than simply grabbing this self contained binary.

We could used dnf instead but in my experience it's more annoying.

skitt:
> > Any particular reason not to use the packaged version?
> 
> The main reason is that the whole `dnf install & clean up` step just takes a lot more time than simply grabbing this self contained binary.
> 
> We could used dnf instead but in my experience it's more annoying.

Right, I can understand that. However I don’t think this change is all that useful in practice: it introduces a fair amount of complexity (compared to including `jq` in the `dnf` command in the base image), and adds 3.8MiB to the operator Dapper image, along with that much to download for _every single_ operator build, in order to save 357KiB from the base image...

mkolesnik:
> Right, I can understand that. However I don’t think this change is all that useful in practice: it introduces a fair amount of complexity (compared to including `jq` in the `dnf` command in the base image), and adds 3.8MiB to the operator Dapper image, along with that much to download for _every single_ operator build, in order to save 357KiB from the base image...

I don't agree this "introduces a fair amount of complexity". IMHO it's clearer like this since it's only being used here.
We could instead push it into the verify script if it makes more sense, I don't mind that.

Also, by the same logic we could argue we need to put the operator-sdk download in Shipyard since its "simpler" but that would be just as wrong since it's only used by this project (and for a very specific target, at that).

skitt:
> > Right, I can understand that. However I don’t think this change is all that useful in practice: it introduces a fair amount of complexity (compared to including `jq` in the `dnf` command in the base image), and adds 3.8MiB to the operator Dapper image, along with that much to download for _every single_ operator build, in order to save 357KiB from the base image...
> 
> I don't agree this "introduces a fair amount of complexity". IMHO it's clearer like this since it's only being used here.
> We could instead push it into the verify script if it makes more sense, I don't mind that.

By complexity, I mean this:

* including `jq` in the Shipyard Dapper base image required adding three characters, ` jq`, and added 357KiB to the (uncompressed) image;
* including `jq` in the operator Dapper image requires adding `curl -Lo /usr/bin/jq "https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64" && chmod a+x /usr/bin/jq`, and results in a 3.8MiB download during the build and an additional 3.8MiB in the Dapper image — and that cost is incurred for every single operator build.

You’re right that `jq` is only used in the operator, but I don’t agree that the benefit of only having it in the operator outweigh the cost.

> Also, by the same logic we could argue we need to put the operator-sdk download in Shipyard since its "simpler" but that would be just as wrong since it's only used by this project (and for a very specific target, at that).

I don’t think it would be simpler to download the Operator SDK in Shipyard, the commands involved would be the same whether it’s in Shipyard or the operator.

mangelajo:
I think it's okay to have the jq in shipyard :-) (357k) and potentially useful to other repos.

Issue #640, opened 2020-08-20
subcl show provides wrong cluster name:


mangelajo:
[majopela@bluehat subm-demo]$ subctl show endpoints

Showing information for cluster "**admin**":
CLUSTER ID          ENDPOINT IP     PUBLIC IP       CABLE DRIVER        TYPE            
10.0.41.139     3.134.80.25     libreswan           local           
majopela-cluster-b-08-20-20-159792048410.0.8.235      18.220.231.188  libreswan           remote      

We should either say "Showing information for context ...." or... look up the cluster name in the config and use.... "Showing information for cluster "majopela-cluster-a-08-20-20-1597920484"

PR #687, opened 2020-09-09
Use cluster name instead of context in subctl show cmds

maayanf24:
Fixes issue #640

### DIFF
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -102,11 +102,7 @@ func getClients(config *rest.Config) (dynamic.Interface, kubernetes.Interface, e
 }
 
 func getClusterName(rawConfig clientcmdapi.Config) *string {
-	context, ok := rawConfig.Contexts[rawConfig.CurrentContext]
-	if !ok {
-		return nil
-	}
-	return &context.Cluster
+	return getClusterNameFromContext(rawConfig, rawConfig.CurrentContext)
 }
 
 func getClusterNameFromContext(rawConfig clientcmdapi.Config, overridesContext string) *string {
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -109,6 +109,14 @@ func getClusterName(rawConfig clientcmdapi.Config) *string {
 	return &context.Cluster
 }
 
+func getClusterNameFromContext(rawConfig clientcmdapi.Config, overridesContext string) *string {
+	context, ok := rawConfig.Contexts[overridesContext]
+	if !ok {
+		return nil
+	}
+	return &context.Cluster
+}
+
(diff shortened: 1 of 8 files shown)


pkg/subctl/cmd/show.go
```diff
@@ -74,22 +74,25 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
...
-		raw, err := config.RawConfig()
-		if err != nil {
-			return restConfig{}, err
-		}
-		context = raw.CurrentContext
+		clusterName = *getClusterName(raw)
 	}
-	return restConfig{config: clientConfig, context: context}, nil
+
+	return restConfig{config: clientConfig, context: clusterName}, nil
```
skitt, line 97: This ends up being somewhat confusing — `context` no longer contains the context, but the cluster name. I see that this fixes the bug in all `show` commands without causing any issues, but it would be better to rename the field.

skitt (changes requested):

pkg/subctl/cmd/root.go
```diff
@@ -109,6 +109,14 @@ func getClusterName(rawConfig clientcmdapi.Config) *string {
 	return &context.Cluster
```
tpantelis, line 109:
```suggestion
	return getClusterNameFromContext(rawConfig, rawConfig.CurrentContext)
```

pkg/subctl/cmd/show.go
```diff
@@ -74,22 +74,25 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
...
+	raw, err := config.RawConfig()
+	if err != nil {
+		return restConfig{}, err
+	}
+
+	var clusterName string
+
 	if overrides.CurrentContext != "" {
-		context = overrides.CurrentContext
+		clusterName = *getClusterNameFromContext(raw, overrides.CurrentContext)
```
tpantelis, line 92: This should check for nil return.

pkg/subctl/cmd/show.go
```diff
@@ -74,22 +74,25 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
...
 	} else {
-		raw, err := config.RawConfig()
-		if err != nil {
-			return restConfig{}, err
-		}
-		context = raw.CurrentContext
+		clusterName = *getClusterName(raw)
```
tpantelis, line 94: same here

tpantelis (approved):

skitt:
@maayanf24 the latest changes to master don’t take this into account ;-).

maayanf24:
> @maayanf24 the latest changes to master don’t take this into account ;-).

I will fix this

PR #699, opened 2020-09-18
Clean up the yamls2go generator invocations

skitt:
The go:generate statement only makes sense (for now) in the
embeddedyamls generate.go file, so remove it everywhere else.

To avoid hard-coding the directories used by the generator, require
them as arguments and specify them in generate.go (relative).


mangelajo (approved):

skitt:
Hmm, I wonder why the bot didn’t create a branch here...

tpantelis (approved):

PR #700, opened 2020-09-18
Generate CRDs as needed

skitt:
Instead of storing CRDs in YAML, generate them from the type
definitions using controller-gen. This also allows us to switch to v1
CRDs by relying on controller-gen 0.4.0 or later.


mangelajo (commented):
Looks good, but something happened to lighthouse.

dfarrell07 (dismissed):

mangelajo (approved):
Looks great

tpantelis (approved):

PR #703, opened 2020-09-22
Make devel the default version, use a variable

mangelajo:


skitt (changes requested):
Could you explain, in the commit message, what the point of using a variable is?

mangelajo:
> Could you explain, in the commit message, what the point of using a variable is?

Yes, also, we need to figure out when to override the var. At tag?, if so... what happens for a release like 0.7.0-pre0 which didn't have matching submariner/lighthouse? 

May be we could detect such tag doesn't exist and leave just "devel"?

skitt:
> > Could you explain, in the commit message, what the point of using a variable is?
> 
> Yes, also, we need to figure out when to override the var. At tag?, if so... what happens for a release like 0.7.0-pre0 which didn't have matching submariner/lighthouse?
> 
> May be we could detect such tag doesn't exist and leave just "devel"?

We’d have to tag before building, or rebuild after tagging... And yes, unsynchronised releases would be a pain! Perhaps we could rely on the version of the module we import...

mangelajo:
Ok, dropping this for a bigger plan we made.

PR #704, opened 2020-09-22
Configure custom domains

vthapar:
1. Adds `custom-domains` flag in `subctl`
2. Configure custom-domains parameter in submariner and
service-discovery specs.

Fixes https://github.com/submariner-io/lighthouse/issues/286

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -367,7 +367,9 @@ func (r *ReconcileSubmariner) reconcileServiceDiscovery(submariner *submopv1a1.S
 					ClusterID:                submariner.Spec.ClusterID,
 					Namespace:                submariner.Spec.Namespace,
 					GlobalnetEnabled:         submariner.Spec.GlobalCIDR != "",
-					CustomDomains:            submariner.Spec.CustomDomains,
+				}
+				if len(submariner.Spec.CustomDomains) > 0 {
+					sd.Spec.CustomDomains = submariner.Spec.CustomDomains
 				}
 				return nil
 			})
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -66,6 +66,7 @@ var (
 	cableDriver          string
 	clienttoken          *v1.Secret
 	globalnetClusterSize uint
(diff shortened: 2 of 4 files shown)


pkg/controller/submariner/submariner_controller.go
```diff
@@ -297,7 +297,9 @@ func (r *ReconcileSubmariner) reconcileServiceDiscovery(submariner *submopv1a1.S
 					ClusterID:                submariner.Spec.ClusterID,
 					Namespace:                submariner.Spec.Namespace,
 					GlobalnetEnabled:         submariner.Spec.GlobalCIDR != "",
-					CustomDomains:            submariner.Spec.CustomDomains,
+				}
+				if submariner.Spec.CustomDomains != nil && len(submariner.Spec.CustomDomains) > 0 {
```
skitt, line 301: `len(nil)` is 0, so

```suggestion
				if len(submariner.Spec.CustomDomains) > 0 {
```

(unlike Java, this won’t throw an NPE).

skitt (approved):
Just a minor comment, this can go in as-is as far as I’m concerned.

pkg/subctl/cmd/join.go
```diff
@@ -419,6 +422,10 @@ func populateSubmarinerSpec(subctlData *datafile.SubctlData, netconfig globalnet
 		crClusterCIDR = netconfig.ClusterCIDR
 	}
 
+	if customDomains == nil {
+		customDomains = *subctlData.CustomDomains
```
skitt, line 426: This is breaking e2e!

skitt (changes requested):
... except that e2e is broken.

pkg/subctl/cmd/join.go
skitt: Ah, I suspect you need to handle older data files where `CustomDomains` ends up `nil`.

vthapar: > Ah, I suspect you need to handle older data files where `CustomDomains` ends up `nil`.

Done, hope this fixes it.

skitt (approved):

aswinsuryan (approved):

PR #708, opened 2020-09-24
Delete existing CRs when joining

skitt:
When joining a cluster, we currently try to update existing Submariner
CRs. But this causes problems if the CRD has been updated in a
non-backwards-compatible fashion (which is allowable since it's still
v1alpha1): on conflict, we try to retrieve the current CR, but that
fails because the existing CR doesn't match the CRD and therefore
can't be unmarshalled. The only reason we retrieve the existing CR is
to copy its resource version to the new CR.

With this change, existing CRs are blindly deleted on conflict, which
allows upgrades to proceed.

### DIFF
--- a/pkg/subctl/operator/submarinercr/ensure.go
+++ b/pkg/subctl/operator/submarinercr/ensure.go
@@ -57,7 +57,7 @@ func Ensure(config *rest.Config, namespace string, submarinerSpec submariner.Sub
 		panic(err.Error())
 	}
 
-	_, err = updateOrCreateSubmariner(submarinerClient, namespace, submarinerCR)
+	_, err = createSubmariner(submarinerClient, namespace, submarinerCR)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


mangelajo (approved):
Thanks for handling this @skitt makes sense.

dfarrell07 (approved):

dfarrell07:
```
Summarizing 3 Failures:

[Fail] [discovery] Test Service Discovery Across Clusters when a pod tries to resolve a service which is present locally and in a remote cluster [It] should resolve the local service 
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/shipyard/test/e2e/framework/framework.go:371

[Fail] [discovery] Test Service Discovery Across Clusters when service export is created before the service [It] should resolve the service 
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/shipyard/test/e2e/framework/framework.go:371

[Fail] [discovery] Test Service Discovery Across Clusters when there are no active pods for a service [It] should not resolve the service 
/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/submariner-io/shipyard/test/e2e/framework/framework.go:371
```

Do we know what's going wrong? Collect logs and retrigger?

dfarrell07:
Logs: https://pastebin.com/Xs029a10
Retriggering...

Issue #147, opened 2020-01-29
Support toleration/taint model for deployment too.


mangelajo:
That allows submariner exclusive worker nodes, etc.

PR #710, opened 2020-09-24
Tolerate Taints in Nodes

mangelajo:
The route-agent and submariner-gateway pods are now configured
as infrastructure pods which will ignore any existing taints.

Without this route-agent does not run on nodes reserved to
specific workloads via node taints, precluding such workloads
to talk to remote clusters.

This patch also allows reserving nodes just for the submariner
gateway with a submariner taint, and the submariner-gateway
will run in such nodes as long as the submariner.io/gateway
label is set.

Fixes-Issue: https://github.com/submariner-io/submariner/issues/828
Related-Issue: https://github.com/submariner-io/submariner-operator/issues/147

### DIFF
--- a/pkg/controller/submariner/submariner_controller.go
+++ b/pkg/controller/submariner/submariner_controller.go
@@ -509,6 +509,8 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 			Volumes: []corev1.Volume{
 				{Name: "host-slash", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/"}}},
 			},
+			// The gateway engine must be able to run on any flagged node, regardless of existing taints
+			Tolerations: []corev1.Toleration{{Operator: corev1.TolerationOpExists}},
 		},
 	}
 	if cr.Spec.CeIPSecIKEPort != 0 {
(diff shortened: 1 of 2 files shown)


pkg/controller/submariner/submariner_controller.go
```diff
@@ -509,6 +509,8 @@ func newEnginePodTemplate(cr *submopv1a1.Submariner) corev1.PodTemplateSpec {
 			Volumes: []corev1.Volume{
 				{Name: "host-slash", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: "/"}}},
 			},
+			// The gateway engine must be able to run on any flagged node, regardless of existing taints
```
sridhargaddam, line 512: @mangelajo I *think* this also makes the Engine and Routeagent pods to be scheduled on the Master nodes. Can you please check once.

sridhargaddam (changes requested):

pkg/controller/submariner/submariner_controller.go
mangelajo: Yes correct, it's intended. There is also no other way (that I know) to workaround random taints.

sridhargaddam (approved):

mangelajo:
@sridhargaddam I had to push again to remove a check in the shell script (that verified that we didn't send instances to the controllers) but since we want to do that now it'd fail.

The votes have been cleared

skitt (approved):

sridhargaddam (approved):

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -367,7 +367,6 @@ function verify_subm_engine_container() {
   kubectl exec $subm_engine_pod_name --namespace=$subm_ns -- env | tee $env_file
 
   # Verify SubM Engine pod environment variables
-  grep "HOSTNAME=$cluster-worker" $env_file
```
dfarrell07, line 370: Heh, I'm amazed these old tests of mine still exist and work...maybe it's about time to remove them?

dfarrell07 (approved):

PR #715, opened 2020-09-26
build: update operator-sdk to v1.0

### LINKED ISSUE #678: Migrate to Operator SDK 1.0

SteveMattar:
### New installed tools:
- controller-gen generates RBAC manifests and crds
- kustomize manages declarative configuration for each content using kustomization.yaml

### Kubebuilder-style project:
- move the CRDs from deploy/crds/ to config/crd/bases
- move the CRs from deploy/crds/ to config/samples
- move the Controller deploy/operator.yaml to config/manager/manager.yaml
- move the RBAC manifests from deploy to config/rbac/
- move pkg/apis to apis
- move pkg/controller to controllers
- move build/Dockerfile to the project's root directory
- move cmd/manager/main.go to the project's root directory
- preparations for kustomize-based deployment for prometheus, webhooks and cert-manager
- scorecard configuration to test and validate the bundle
- for each api type the `+k8s:deepcopy-gen:interfaces=` marker was replaced with `+kubebuilder:object:root=true`
- for each api type the `+operator-sdk:csv:customresourcedefinitions` marker was added
- for each controller the struct Reconciler struct has been renamed from Reconcile<Kind> to <Kind>Reconciler
- the Reconciler variables are now exported
- for each controller a new function `SetupWithManager` was added that is being called in main.go
- operator-lib module was added in order to keep the SDK leader.Become
- no need for zz_generated files anymore!
- OPERATOR_NAME and POD_NAME environment variable are no longer used
- envtest framework is not yet implemented

### Add More targets to Makefile:
- `make generate` handles deepcopy and adds the license text
- `make manifests` uses controller-gen and replaces the old commands `operator-sdk generate k8s`, `operator-sdk generate openapi` and `operator-sdk generate crds`

### Dapper:
- install controller-gen and kustomize

### Other changes:
- rename the globalnet ServiceAccount to follow [RFC1123](https://tools.ietf.org/html/rfc1123)
- remove the deprecated `multiclusterservices` CRD
- update the CRs and add the missing ones for lighthouse
- add more entries to yamllint
- adjust the validations test

### TODO:
[ ] main.go zap.FlagSet() is no longer supported
[ ] main.go in opeartor-sdk v1 some utilities were moved to internal
[ ] automation for adding the fields displayName and description to customresourcedefinitions.owned[]
[ ] delete the zz_generated files

#### generate the manifest files
- make generate
- make manifests

#### build the operator docker image
- make docker-build IMG=quay.io/submariner/submariner-operator:0.6.1

#### bundle format
- make bundle VERSION=0.6.1 CHANNELS=alpha DEFAULT_CHANNEL=aplha IMG=quay.io/submariner/submariner-operator:0.6.1
- make bundle-build BUNDLE_IMG=quay.io/submariner/submariner-operator-bundle:0.6.1

#### packagemanifests format
- make packagemanifests VERSION=0.6.1 FROM_VERSION=0.5.0 CHANNEL=alpha IMG=quay.io/submariner/submariner-operator:0.6.1

### Refs:
- [migration guide](https://sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide)
- [kustomize docs](https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html)

**BREAKING CHANGE**: migrating the legacy project to the new layout. closes #678


Makefile
```diff
@@ -93,15 +120,9 @@ generate-clientset:
 	GO111MODULE=on $${GOPATH}/src/k8s.io/code-generator/generate-groups.sh \
 		client,deepcopy \
 		github.com/submariner-io/submariner-operator/pkg/client \
-		github.com/submariner-io/submariner-operator/pkg/apis \
+		github.com/submariner-io/submariner-operator/apis \
```
mangelajo, line 123: Why do we move the apis out of pkg?  is it necessary?

Makefile
```diff
@@ -113,7 +134,82 @@ preload-images:
...
+# We check BUILD_ARGS since that's what the compile script uses
+ifeq (--debug,$(findstring --debug,$(BUILD_ARGS)))
+	go build -o bin/submariner-operator main.go
+else
+	go build -ldflags -s -w -o bin/submariner-operator main.go
+endif
+
+# Run against the configured Kubernetes cluster in ~/.kube/config
+run: generate fmt vet manifests
+	go run ./main.go
```
mangelajo, line 154: Where does this come from, we don't need it in our case have our own E2E testing

Makefile
```diff
@@ -113,7 +134,82 @@ preload-images:
...
+
+# Uninstall CRDs from a cluster
+uninstall: manifests
+	$(KUSTOMIZE) build config/crd | sed 's/$$(VERSION)/$(VERSION)/g' | kubectl delete -f -
+
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: manifests clusters preload-images
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | sed 's/$$(VERSION)/$(VERSION)/g' | kubectl apply -f -
+
```
mangelajo, line 168: All this is unnecesary in our case, our testing is more complex than that, because we do multicluster, all this is handled from shipyard/subctl. etc.

Makefile
```diff
@@ -113,7 +134,82 @@ preload-images:
...
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+# Run go fmt against code
+fmt:
+	go fmt ./...
+
+# Run go vet against code
+vet:
+	go vet ./...
+
```
mangelajo, line 180: If we want those targets they should live in shipyard.

skitt (changes requested):
This is *amazing* work, thanks Steve!

I haven’t reviewed it entirely, but a few thoughts spring to mind:
* our current operator image uses the scratch base image, and only ships the operator binary in the container image — is that still possible?
* could you repackage this using successive commits to show the various changes? in particular, separate commits for steps which only involve moving files around, or generating files, would make the PR much easier to review
* does it still make sense to generate the CRDs in every build, as in #700? I know we’ll need to keep historical CRDs, and presumably bundle CRDs, but I’d still like to generate as much as possible all the time...

Makefile
```diff
@@ -113,7 +134,82 @@ preload-images:
...
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+
+# Build the docker image
+docker-build: test
+ifeq (--debug,$(findstring --debug,$(BUILD_ARGS)))
+	docker build --build-arg DEBUG=true . -t ${IMG}
+else
+	docker build --build-arg DEBUG=false . -t ${IMG}
+endif
+
```
mangelajo, line 192: This belongs to shipyard, we already have tooling in there for building.

Makefile
skitt: That’s what the SDK expects now...

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -26,20 +26,19 @@ import (
...
-	"service_account.yaml",
-	"submariner/globalnet_cluster_role.yaml",
-	"submariner/globalnet_cluster_role_binding.yaml",
-	"cluster_role.yaml",
-	"cluster_role_binding.yaml",
-	"lighthouse/crds/multiclusterservices_crd.yaml",
-	"lighthouse/crds/serviceexport_crd.yaml",
-	"lighthouse/crds/serviceimport_crd.yaml",
-	"lighthouse/cluster_role_binding.yaml",
-	"lighthouse/cluster_role.yaml",
```
mangelajo, line 42: Please. let's remain the 

lighthouse/
and 
submariner/

CRDs untouched, or let's move them to a different directory altogether, or see how to do it.

The convention we had here is:

Everything on the 
   * submariner directory is stuff used by the core submariner services
   * lighthouse directory is stuff used by lighthouse
   * <outside> : stuff and CRDs that the operator need to work & exist.

Currently after latest patches, the items in the "submariner & lighthouse" directories are installed by the operator internally, they don't need to be part of the CSV anympre.

mangelajo (changes requested):
Hey, thanks for working on this Steve. But I see this patch is doing two (or more things at once), moving to operator sdk v1 and also switching our build system out of Dapper & shipyard scripts. 

I think we need minimize those changes and discuss more widely with the team.

Also see some notes at the end about the CRD files

Makefile
SteveMattar: I followed the migration guide and this is how the directories layout should look like. I think we can choose the directories by changing the controller-gen parameters and kustomize configuration

SteveMattar:
> This is _amazing_ work, thanks Steve!
> 
> I haven’t reviewed it entirely, but a few thoughts spring to mind:
> 
>     * our current operator image uses the scratch base image, and only ships the operator binary in the container image — is that still possible?
> 
>     * could you repackage this using successive commits to show the various changes? in particular, separate commits for steps which only involve moving files around, or generating files, would make the PR much easier to review
> 
>     * does it still make sense to generate the CRDs in every build, as in #700? I know we’ll need to keep historical CRDs, and presumably bundle CRDs, but I’d still like to generate as much as possible all the time...

thanks @skitt 
* the docker image is a multi-stage build and it ships the operator binary only based on ubi. We can change it of course - does the build run on shipyard?
* yes, I will separate the commits to make it more clear
* sure, I generate the CRDs using controller-gen by running `make manifests` then I can run `make bundle` or `make packagemanifests` depends on the format I want in order to create the CSV metadata compatible with OLM

Makefile
SteveMattar: oh yes, I can see it in shipyard we can remove `fmt` and `vet`
`make manager` builds the operator binary using `go build` since in operator-sdk v1 there's no `build` command anymore.
`make run` runs the operator on a k8s cluster during development it can be removed as well

Makefile
SteveMattar: @mangelajo thanks I will remove this part

Makefile
SteveMattar: I think we can remove `fmt` and `vet` as we already have what @dfarrell07 prepared as part of shipyard
`make manifests` actually generates the manifests files CRDS, RBAC, etc... should I move it to shipyard as well?

Makefile
SteveMattar: ok let's discuss it I probably missed the idea of shipyard

SteveMattar:
#724

Issue #706, opened 2020-09-24
Intra-cluster latency benchmark


maayanf24:
Add ability to test intra-cluster latency benchmark: 
ClusterA worker node to ClusterA gateway

Update subctl command:
`subctl benchmark latency <kubeconfig1> [<kubeconfig2>] [flags]`

PR #718, opened 2020-09-29
Add intra-cluster test to subctl latency benchmark

maayanf24:
Fixes issue #706

### DIFF
--- a/pkg/subctl/cmd/benchmark.go
+++ b/pkg/subctl/cmd/benchmark.go
@@ -24,7 +24,7 @@ var (
 		},
 		Run: testThroughput,
 	}
-	benchmarkLatenchCmd = &cobra.Command{
+	benchmarkLatencyCmd = &cobra.Command{
 		Use:   "latency <kubeconfig1> [<kubeconfig2>]",
 		Short: "Benchmark latency between two clusters",
 		Long:  "This command runs latency benchmark tests between two clusters",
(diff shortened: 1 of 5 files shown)


pkg/subctl/cmd/benchmark.go
```diff
@@ -18,30 +20,38 @@ var (
...
 	}
 )
 
 func init() {
+	benchmarkLatenchCmd.PersistentFlags().BoolVar(&intraCluster, "intra-cluster-baseline", false,
+		"Run benchmark latency test with intra-cluster pods")
+
+	benchmarkThroughputCmd.PersistentFlags().BoolVar(&intraCluster, "intra-cluster-baseline", false,
+		"Run benchmark latency test with intra-cluster pods")
+
```
skitt, line 44: I’m wondering if we need a separate flag for this, or if we can use the number of kubeconfigs given to choose... What do you think?

maayanf24: Having the flag is more explicit.
Without the flag, if a user forget to add second kubeconfig the result will be another type of test.

skitt: Right, more explicit is better since the failure scenario isn’t obvious.

skitt (approved):

pkg/subctl/cmd/benchmark.go
```diff
@@ -18,30 +20,38 @@ var (
...
 		Args: func(cmd *cobra.Command, args []string) error {
-			return checkThroughputArguments(args)
+			return checkThroughputArguments(args, intraCluster)
 		},
 		Run: testLatency,
 	}
 )
 
 func init() {
+	benchmarkLatenchCmd.PersistentFlags().BoolVar(&intraCluster, "intra-cluster-baseline", false,
```
sridhargaddam, line 39:
```suggestion
	benchmarkLatencyCmd.PersistentFlags().BoolVar(&intraCluster, "intra-cluster-baseline", false,
```

sridhargaddam (commented):
small nit, otherwise LGTM.

sridhargaddam (approved):

skitt (approved):

PR #724, opened 2020-10-02
build: update operator-sdk to v1.0

### LINKED ISSUE #678: Migrate to Operator SDK 1.0

SteveMattar:
### New installed tools:
- controller-gen generates RBAC manifests and crds
- kustomize manages declarative configuration for each content using kustomization.yaml

### Kubebuilder-style project:
- move the CRDs from deploy/crds/ to config/crd/bases
- move the CRs from deploy/crds/ to config/samples
- move the Controller deploy/operator.yaml to config/manager/manager.yaml
- move the RBAC manifests from deploy to config/rbac/
- move pkg/apis to apis
- move pkg/controller to controllers
- move build/Dockerfile to the project's root directory
- move cmd/manager/main.go to the project's root directory
- preparations for kustomize-based deployment for prometheus, webhooks and cert-manager
- scorecard configuration to test and validate the bundle
- replace the `+k8s:deepcopy-gen:interfaces=` marker with `+kubebuilder:object:root=true` for each api type
- add the marker `+operator-sdk:csv:customresourcedefinitions` for each api
- rename each Reconciler from `Reconcile<kind>` to `<kind>Reconciler`
- replace the `Add` function with `SetupWithManager`
- operator-lib module was added in order to keep the SDK leader.Become
- save the bundle and packagemanifests for the current version
- no need for zz_generated files anymore!
- OPERATOR_NAME and POD_NAME environment variable are no longer used
- envtest framework is not yet implemented

### Add More targets to Makefile:
- `make generate` handles deepcopy and adds the license text
- `make manifests` uses controller-gen and replaces the old commands `operator-sdk generate k8s`, `operator-sdk generate openapi` and `operator-sdk generate crds`

### Dapper:
- install controller-gen and kustomize

### Other changes:
- rename the globalnet ServiceAccount to follow [RFC1123](https://tools.ietf.org/html/rfc1123)
- update the CRs
- add more entries to yamllint
- handle the embedded yamls and support the new directory structure when generating crds

### TODO:

- [ ] main.go zap.FlagSet() is no longer supported
- [ ] main.go in opeartor-sdk v1 some utilities were moved to internal
- [x] automation for adding the fields displayName and description to customresourcedefinitions.owned[]

#### build the operator docker image
- make operator-image

#### bundle format
- make bundle VERSION=0.6.1 CHANNELS=alpha DEFAULT_CHANNEL=alpha
- make bundle-build

#### packagemanifests format
- make packagemanifests VERSION=0.6.1 FROM_VERSION=0.5.0 CHANNEL=alpha

### Refs:
- [migration guide](https://sdk.operatorframework.io/docs/building-operators/golang/project_migration_guide)
- [controller-gen](https://book.kubebuilder.io/reference/controller-gen.html)
- [kustomize](https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html)

**BREAKING CHANGE**: migrating the legacy project to the new layout. closes #678


Dockerfile.dapper
```diff
@@ -4,17 +4,21 @@ ARG DAPPER_HOST_ARCH
 ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} \
     DAPPER_ENV="REPO TAG QUAY_USERNAME QUAY_PASSWORD GITHUB_SHA MAKEFLAGS CLUSTERS_ARGS DEPLOY_ARGS E2E_ARGS RELEASE_ARGS" \
     DAPPER_SOURCE=/go/src/github.com/submariner-io/submariner-operator DAPPER_DOCKER_SOCKET=true \
-    OPERATOR_SDK_VERSION=0.12.0 \
-    CONTROLLER_GEN_VERSION=0.3.0 \
+    OPERATOR_SDK_VERSION=1.0.1 \
+    KUSTOMIZE_VERSION=3.5.4 KUSTOMIZE=$(GOBIN)/kustomize \
+    CONTROLLER_GEN_VERSION=0.3.0 CONTROLLER_GEN=$(GOBIN)/controller-gen \
```
mkolesnik, line 9: Defining `KUSTOMIZE` and `CONTROLLER_GEN` here without using them in the Dockerfile strikes me as odd.

Furthermore, we're not doing this "style" for any other command so why are these two special?

Finally, /go/bin is already a part of `$PATH` so you don't actually need to specify it.

SteveMattar: Thanks for the review @mkolesnik I will remove those envs and call the cli commands directly in `Makefile`

SteveMattar:
@mangelajo my e2e test fails because the new version of controller-runtime registers the flag `--kubeconfig` and it conflicts with ours the one in `subctl` command
https://github.com/kubernetes-sigs/controller-runtime/blob/c000ea850121b53dd769f624f3bfa74a4fcf100f/pkg/client/config/config.go#L38
do you have any experience with this? is there any workaround?

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,15 @@
+/*
+
```
skitt, line 2: Please add a copyright header here:

```suggestion
© 2020 Red Hat, Inc. and others.
```

skitt (changes requested):
A few initial comments, I’m looking into why this fails e2e.

Please also rebase this on top of current master, handling the conflicts, and ideally, squash the commits which fix issues in earlier commits with those commits (so the commits are never broken).

Makefile
```diff
@@ -49,16 +84,15 @@ build: $(BINARIES)
 
 build-cross: $(CROSS_TARBALLS)
 
-OSDK_BUILD_ARGS := -ldflags -X=github.com/submariner-io/submariner-operator/pkg/version.Version=$(CALCULATED_VERSION)
-# We check BUILD_ARGS since that's what the compile script uses
-ifneq (--debug,$(findstring --debug,$(BUILD_ARGS)))
-OSDK_BUILD_ARGS += -ldflags -s -ldflags -w
-endif
+operator-image: bin/submariner-operator
+	docker build -t ${IMG} .
```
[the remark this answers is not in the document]
SteveMattar: you mean `make package/.image.submariner-operator` ? I think we can use this target I'll try it locally first

Makefile
```diff
@@ -83,16 +117,13 @@ bin/subctl-%: pkg/subctl/operator/common/embeddedyamls/yamls.go $(shell find pkg
...
 	go generate pkg/subctl/operator/common/embeddedyamls/generate.go
 
 # Operator CRDs
-deploy/crds/submariner.io_servicediscoveries.yaml: ./pkg/apis/submariner/v1alpha1/servicediscovery_types.go vendor/modules.txt
-	controller-gen crd paths="./..." output:crd:artifacts:config=deploy/crds
-
-deploy/crds/submariner.io_submariners.yaml: ./pkg/apis/submariner/v1alpha1/submariner_types.go vendor/modules.txt
+deploy/crds/submariner.io_submariners.yaml deploy/crds/submariner.io_servicediscoveries.yaml: ./apis/submariner/v1alpha1/submariner_types.go ./apis/submariner/v1alpha1/servicediscovery_types.go vendor/modules.txt
```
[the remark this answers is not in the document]
SteveMattar: yes, you are right I will revert this change

hack/boilerplate.go.txt
SteveMattar: `controller-gen` has a `year` argument I can use it to automate the process

.gitignore
```diff
@@ -5,9 +5,33 @@ dist/
...
 deploy/submariner/crds
+config/bundle/kustomization.yaml
 output/
 vendor/
 *.coverprofile
-.idea
 .shflags
 junit.xml
+
+# Binaries for programs and plugins
```
dfarrell07, line 15: Running many of the make targets locally, removing this .gitignore change, I don't see any modified files.

Imho, we should recommend folks configure their editors/IDEs to store their files in some location other than the repo. Generally, I think we should try to avoid getting into the business of ignores rules for dev-specific tools.

Makefile
```diff
@@ -127,11 +161,41 @@ preload-images:
 		import_image quay.io/submariner/$${image}; \
 	done
 
+# Generate kustomization.yaml for bundle
+kustomization: manifests
```
dfarrell07, line 165: Does this need to be included in the release process somehow? I ask because when I run `make kustomization`, I get a modified version of `config/manager/kustomization.yaml`.

```
-  newTag: 0.7.0
+  newTag: v0.7.0-16-g80b6c81-dev
```

dfarrell07 (changes requested):
Thanks for all the work @SteveMattar! Some comments inline.

Also, I don't know how to comment inline on commit messages, but it would be ideal to start body sentences with Caps and end with periods. See: https://submariner.io/contributing/code-review/#commit-message-formatting and https://chris.beams.io/posts/git-commit/.

Makefile
```diff
@@ -127,11 +161,41 @@ preload-images:
...
+	operator-sdk generate kustomize manifests -q && \
+	cd config/manager && kustomize edit set image controller=$(IMG) && cd ../../ && \
+	(cd config/bundle && \
+	cat kustomization.template.yaml > kustomization.yaml && \
+	sed -i -e 's/$${VERSION}/$(VERSION)/g' kustomization.yaml && \
+	kustomize edit add annotation createdAt:"$(shell date "+%Y-%m-%d %T")" -f && \
+	cd ../../ )
+
+# Generate bundle manifests and metadata, then validate generated files.
+bundle: kustomization
```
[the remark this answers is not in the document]
dfarrell07: Maybe we should think about making CI cover this somehow?

SteveMattar:
Thanks for your time @dfarrell07 
I will edit the commit messages

Makefile
```diff
@@ -31,6 +31,39 @@ GOARCH = $(shell go env GOARCH)
...
+# Options for 'bundle-build'
+ifneq ($(origin CHANNELS), undefined)
+BUNDLE_CHANNELS := --channels=$(CHANNELS)
+endif
+ifneq ($(origin DEFAULT_CHANNEL), undefined)
+BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
+endif
+BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
+
+# Options for "packagemanifests".
```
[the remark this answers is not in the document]
SteveMattar: ACK I will change that

Makefile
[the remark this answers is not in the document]
SteveMattar: @dfarrell07 it fails because the calculated version includes some invalid characters for the bundle
you should specify the `VERSION` variable before running `make bundle`. kindly see the PR description for an example.

Makefile
```diff
@@ -127,11 +161,41 @@ preload-images:
...
+	(cd config/bundle && kustomize edit add resource ../../bundle/manifests/submariner.clusterserviceversion.yaml && cd ../../) && \
+	kustomize build config/bundle/ --load_restrictor=LoadRestrictionsNone --output bundle/manifests/submariner.clusterserviceversion.yaml && \
+	operator-sdk bundle validate ./bundle
+
+# Build the bundle image.
+bundle-build:
+	$(SCRIPTS_DIR)/build_image.sh -i submariner-operator-bundle -f bundle.Dockerfile
+
+# Generate package manifests.
+packagemanifests: kustomization
```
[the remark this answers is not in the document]
SteveMattar: same here as done in `make bundle`

.gitignore
SteveMattar: I agree I'll remove this section

Makefile
SteveMattar: do you mean the new release repo? then yes.
`make bundle` and `make packagemanifests` both create a new bundle in different formats
and both use `make kustomization` as a preparation step

.gitignore
SteveMattar: I agree, I'll remove this section

SteveMattar (commented):
Thanks for your time @dfarrell07
I will edit the commit messages

Makefile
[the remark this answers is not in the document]
dfarrell07: Talking about this with @SteveMattar on the Automation WG sync today, the plan is to try to validate the version (if easy, ie doesn't have "-") and fail more quickly/clearly.

Imho the ideal is to have targets that pass by default (without any var overrides), but quickly and clearly failing is also okay.

Makefile
dfarrell07: Thanks @SteveMattar, that makes sense. :+1: 

Are there any updates we need to make to https://submariner.io/contributing/release-process/?

SteveMattar: @dfarrell07 yes in [step 9](https://submariner.io/contributing/release-process/#step-9-update-the-submariner-operator-on-operatorhubio) I have already prepared a commit locally for the submariner-website but waiting for this one to be merged first

Makefile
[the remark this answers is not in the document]
SteveMattar: following our discussion I've added a validation using regex

Makefile
dfarrell07: Imho it would be best to go ahead and push it, mark as draft, and we can start reviewing but hold on merging until this lands.

dfarrell07 (changes requested):

Makefile
SteveMattar: ACK

Makefile
```diff
@@ -127,11 +163,46 @@ preload-images:
 		import_image quay.io/submariner/$${image}; \
 	done
 
+# test if VERSION matches the semantic versioning rule
+is-semantic-version:
+	[[ $(VERSION) =~ $(PATTERN) ]] || \
+	(printf '\nerror: VERSION does not match the semantic versioning rule\n\n' && exit 1)
```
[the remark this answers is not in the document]
SteveMattar: Not exactly, kubebuilder is just the framework. The `operator-sdk` tool which generates the bundle fails as it expects a semantic version. I understand what you are saying, what do you think of this message?
```suggestion
	(printf '\nerror: Invalid VERSION - Please supply a semantic version for your bundle or package.\n\n' && exit 1)
```

skitt:
@SteveMattar could you rebase this and resolve the conflicts? @dfarrell07 could you see if your comments have been addressed?

Thanks!

Makefile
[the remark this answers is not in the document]
dfarrell07: I guess what I'm trying to clarify is that a version "major.minor.patch-pre" is still a valid semantic version, per https://github.com/submariner-io/submariner-operator/pull/786.

dfarrell07 (changes requested):
When I run

```
make bundle VERSION=0.7.0 CHANNELS=alpha DEFAULT_CHANNEL=alpha
```

I see

```
--- a/bundle/manifests/submariner.io_submariners.yaml
+++ b/bundle/manifests/submariner.io_submariners.yaml
@@ -473,6 +473,8 @@ spec:
               type: object
             natEnabled:
               type: boolean
+            networkPlugin:
+              type: string
             routeAgentDaemonSetStatus:
               properties:
                 lastResourceVersion:
diff --git a/config/crd/bases/submariner.io_submariners.yaml b/config/crd/bases/submariner.io_submariners.yaml
index c7a6aab..7d3bc58 100644
--- a/config/crd/bases/submariner.io_submariners.yaml
+++ b/config/crd/bases/submariner.io_submariners.yaml
@@ -539,6 +539,8 @@ spec:
               type: object
             natEnabled:
               type: boolean
+            networkPlugin:
+              type: string
             routeAgentDaemonSetStatus:
               properties:
                 lastResourceVersion:
```

It seems like we need manual updates after https://github.com/submariner-io/submariner-operator/pull/779.

dfarrell07:
I see a lot of minor formatting changes (indentation mostly) when I run `make packagemanifests VERSION=0.7.0 FROM_VERSION=0.6.1 CHANNEL=alpha`. Is that something that should be updated?

dfarrell07:
Regarding the PR message, I don't see a command `make docker-build`.

Makefile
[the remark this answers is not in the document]
SteveMattar: @dfarrell07 you are right although in the operator-sdk [official documentation](https://sdk.operatorframework.io/docs/cli/operator-sdk_generate_bundle/) they use the same term.
I will use your suggestion and replace kubebuilder by operator-sdk.
Thanks!

SteveMattar:
> When I run
> 
> ```
> make bundle VERSION=0.7.0 CHANNELS=alpha DEFAULT_CHANNEL=alpha
> ```
> 
> I see
> 
> ```
> --- a/bundle/manifests/submariner.io_submariners.yaml
> +++ b/bundle/manifests/submariner.io_submariners.yaml
> @@ -473,6 +473,8 @@ spec:
>                type: object
>              natEnabled:
>                type: boolean
> +            networkPlugin:
> +              type: string
>              routeAgentDaemonSetStatus:
>                properties:
>                  lastResourceVersion:
> diff --git a/config/crd/bases/submariner.io_submariners.yaml b/config/crd/bases/submariner.io_submariners.yaml
> index c7a6aab..7d3bc58 100644
> --- a/config/crd/bases/submariner.io_submariners.yaml
> +++ b/config/crd/bases/submariner.io_submariners.yaml
> @@ -539,6 +539,8 @@ spec:
>                type: object
>              natEnabled:
>                type: boolean
> +            networkPlugin:
> +              type: string
>              routeAgentDaemonSetStatus:
>                properties:
>                  lastResourceVersion:
> ```
> 
> It seems like we need manual updates after #779.

I rebased today. In my opinion these changes should be included in v0.8.0

SteveMattar:
> Should we somehow handle the version replaced by a new bundle? When I run `make bundle VERSION=0.7.0 CHANNELS=alpha DEFAULT_CHANNEL=alpha`, I see:
> 
> ```
> --- a/bundle/manifests/submariner.clusterserviceversion.yaml
> +++ b/bundle/manifests/submariner.clusterserviceversion.yaml
>    maturity: alpha
>    provider:
>      name: submariner.io
> -  replaces: submariner.v0.6.1
> +  replaces: submariner.v0.0.0
> ```

@dfarrell07 if the previous version is smaller than the new version then everything should work well.
Otherwise, if you run the command more than once using the same `VERSION` then the value for `replaces` is reset. 
I tested it several times but let me know if you find any scenario which doesn't seem to make sense then I will handle this field manually.

SteveMattar:
> I see a lot of minor formatting changes (indentation mostly) when I run `make packagemanifests VERSION=0.7.0 FROM_VERSION=0.6.1 CHANNEL=alpha`. Is that something that should be updated?

Yes, indentation and white space. Unfortunately, this is how these tools output the generated YAML files. I've added those files to the ignore list in `.yamllint`. Do you think we should use `yamlfmt` to automatically modify the offset.

SteveMattar:
> Regarding the PR message, I don't see a command `make docker-build`.

@dfarrell07 I moved this operation to shipyard and forgot to update the PR message. Thanks for noticing :+1:

skitt (approved):
LGTM, let’s get this in and address any remaining issues in follow-up patches (nothing we rely on currently is broken).

dfarrell07:
>  Do you think we should use yamlfmt to automatically modify the offset.

That's a good idea, but I also wonder if this is something we could help fix upstream.

> In my opinion these changes should be included in the next bundle v0.8.0

Yep, that makes sense.

Agreed, let's get this in and iterate on these minor details. Thanks for all the good work @SteveMattar!

dfarrell07 (approved):

PR #726, opened 2020-10-05
Add repo/version override commands

mangelajo:
This will enable subctl to install specific versions of each component
during CI. The parameters are hidden to avoid overwhelming the user
while looking for help.


pkg/subctl/cmd/image_overrides/image_overrides.go
```diff
@@ -0,0 +1,77 @@
...
+		return override
+	} else {
+		return defaultVal
+	}
+}
+
+func (img *ImageOverrides) GetSubmarinerRegistryAndVersion(defaultRegistry, defaultVersion string) (registry, version string) {
+	defaultVersion = defaultStr(defaultVersion, versions.DefaultSubmarinerVersion)
+	registry = defaultStr(defaultRegistry, img.submariner.registry)
+	version = defaultStr(defaultVersion, img.submariner.version)
```
mkolesnik, line 53: Wondering why you chose this approach instead of just setting the defaults in the flag declarations?

mkolesnik (commented):
Some thoughts

pkg/subctl/cmd/image_overrides/image_overrides.go
```diff
@@ -0,0 +1,77 @@
...
+		"Lighthouse container registry")
+	cmd.Flags().StringVar(&img.lighthouse.version, "lighthouse-version", "",
+		"Lighthouse container version")
+
+	_ = cmd.Flags().MarkHidden("operator-registry")
+	_ = cmd.Flags().MarkHidden("operator-version")
+	_ = cmd.Flags().MarkHidden("submariner-registry")
+	_ = cmd.Flags().MarkHidden("submariner-version")
+	_ = cmd.Flags().MarkHidden("lighthouse-registry")
+	_ = cmd.Flags().MarkHidden("lighthouse-version")
```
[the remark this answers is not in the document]
mangelajo: That's a good question, I have no idea, I'll check that, we can always leave the test to make sure it has help if un-hidden, but I wonder if it's possible to show hidden options.

pkg/subctl/cmd/image_overrides/image_overrides.go
mangelajo: Hmm, that could be an option, we still need to check how it works. We have two sources of "versions", 

The --version parameter, and the specific overrides, I guess it would work... I'll try that.

mangelajo:
@mkolesnik still didn't handle any of your comments, but updated the patch with a missing file I had.

pkg/subctl/cmd/image_overrides/image_overrides.go
mkolesnik: Hmm I see, though I wonder if a `--version` parameter make much sense. E.G can you deploy v0.1.0 with a v0.5.0 subctl? :)

skitt: Shouldn’t we leave the defaults to the operator, *i.e.* use blank defaults here? (Apart from the operator...)

mangelajo:
Dropping this in favor of the more flexible approach @SteveMattar is taking: https://github.com/submariner-io/submariner-operator/pull/829

PR #727, opened 2020-10-05
Add images make target to simplify UX

dfarrell07:
To create a more unified Submariner UX, in the process of standardizing
all repos on "build" target for building all binaries and "images"
target for building all images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ GOOS = $(shell go env GOOS)
 
 # Targets to make
 
-clusters: build
+clusters: build images
 
 deploy: clusters preload-images
 
@@ -45,7 +45,7 @@ test: unit-test
 clean:
 	rm -f $(BINARIES) $(CROSS_BINARIES) $(CROSS_TARBALLS)
 
-build: operator-image $(BINARIES)
+build: $(BINARIES)
 
 build-cross: $(CROSS_TARBALLS)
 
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

Makefile
```diff
@@ -58,6 +58,8 @@ endif
 operator-image: vendor/modules.txt pkg/subctl/operator/common/embeddedyamls/yamls.go
 	operator-sdk build quay.io/submariner/submariner-operator:$(DEV_VERSION) --go-build-args "$(OSDK_BUILD_ARGS)"
 
+images: build operator-image
```
skitt, line 61: Shouldn’t this be

```suggestion
images: operator-image
```

only?

skitt (changes requested):

Makefile
dfarrell07: At least in the main repo and Lighthouse, `images` depends on building the binaries.

skitt: I suspect the Submariner and Lighthouse repos have dependencies on `build` in the `images` target because the individual image targets are insufficiently specified (each image ships binaries, and should depend on those binaries directly). In the operator, the binary shipped in the image is built by the Operator SDK in the `operator-image` recipe.

skitt: Put another way, meta-targets such as `images` should only depend on what they really want, not targets related to _how_ to build what they want.

dfarrell07: Okay, so should `clusters` also depend on `build` then? Otherwise it seems we're changing behavior there.

skitt (approved):

PR #731, opened 2020-10-05
Make the daemonset statuses optional

skitt:
... otherwise updates fail in the controller when the status isn't
available.


skitt:
Not sure this is actually needed to fix the issue I’m seeing (with the Helm charts).

Issue #663, opened 2020-08-31
subctl show shows cluster information several times in some circumstances


skitt:
After running `make deploy` in Shipyard, I get the following output from `subctl show gateways`:

```

Showing information for cluster "cluster2":
NODE                HA STATUS       SUMMARY                         
cluster2-worker     active          All connections (1) are established

Showing information for cluster "cluster1":
NODE                HA STATUS       SUMMARY                         
cluster1-worker     active          All connections (1) are established

Showing information for cluster "cluster2":
NODE                HA STATUS       SUMMARY                         
cluster2-worker     active          All connections (1) are established
```

`KUBECONFIG` is `/home/skitt/go/src/github.com/submariner-io/shipyard/output/kubeconfigs/kind-config-cluster2:/home/skitt/go/src/github.com/submariner-io/shipyard/output/kubeconfigs/kind-config-cluster1:`, which is the result of the command recommended by the deployment, `export KUBECONFIG=$(find $(git rev-parse --show-toplevel)/output/kubeconfigs/ -type f -printf %p:)`.

PR #734, opened 2020-10-07
Ignore empty KUBECONFIG entries in subctl show cmds

maayanf24:
Fixes issue #663

### DIFF
--- a/pkg/subctl/cmd/show.go
+++ b/pkg/subctl/cmd/show.go
@@ -64,13 +64,15 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
 	}
 
 	for _, item := range rules.Precedence {
-		rules.ExplicitPath = item
-		config, err := getClientConfigAndClusterName(rules, overrides)
-		if err != nil {
-			return nil, err
-		}
+		if item != "" {
+			rules.ExplicitPath = item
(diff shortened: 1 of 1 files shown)


skitt (approved):
Thanks for taking care of this! The change in the code seems fine, but perhaps the description could be more accurate: “Ignore empty KUBECONFIG entries” (it technically doesn’t ensure `subctl show` doesn’t duplicate data, since duplicate KUBECONFIG entries will still result in duplicate output).

maayanf24:
> Thanks for taking care of this! The change in the code seems fine, but perhaps the description could be more accurate: “Ignore empty KUBECONFIG entries” (it technically doesn’t ensure `subctl show` doesn’t duplicate data, since duplicate KUBECONFIG entries will still result in duplicate output).

@skitt - Thanks, that make sense.
I'll update the description

nyechiel (approved):

skitt:
> > Thanks for taking care of this! The change in the code seems fine, but perhaps the description could be more accurate: “Ignore empty KUBECONFIG entries” (it technically doesn’t ensure `subctl show` doesn’t duplicate data, since duplicate KUBECONFIG entries will still result in duplicate output).
> 
> @skitt - Thanks, that make sense.
> I'll update the description

Could you update the commit message too? That’s what survives in the project history.

Issue #733, opened 2020-10-06
The first run of the operator crashes


skitt:
After deploying with `subctl`, the first startup of the operator crashes:

```
{"level":"error","ts":1601970756.7683287,"logger":"controller-runtime.source","msg":"if kind is a CRD, it should be installed before calling Start","kind":"ServiceDiscovery.submariner.io","error":"no matches for kind \"ServiceDiscovery\" in version \"submariner.io/v1alpha1\"","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\tsubmariner-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/source.(*Kind).Start\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/source/source.go:116\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:153\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:188\nsigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).startLeaderElectionRunnables.func1\n\tsubmariner-operator/vendor/sigs.k8s.io/controller-runtime/pkg/manager/internal.go:473"}
{"level":"error","ts":1601970756.768453,"logger":"cmd","msg":"Manager exited non-zero","error":"no matches for kind \"ServiceDiscovery\" in version \"submariner.io/v1alpha1\"","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\tsubmariner-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\tsubmariner-operator/cmd/manager/main.go:166\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:203"}
```

What happens is that the operator creates the ServiceDiscovery CRD, but apparently too late; then it crashes, and the next time it starts, the CRD is present, and the operator starts correctly.

PR #740, opened 2020-10-08
Revert "Handle Lighthouse CRDs in the operator"

skitt:
This reverts commit f47dc5105e7acf97bd4a327a775471005c61c111.

Attempting to create the ServiceDiscovery CRD from the operator fails:
the operator starts, creates the CRD, then the manager starts, and
doesn't find the CRD; the operator then stops, is restarted, and
finally finds the CRD. It seems better to require the CRD to be
available before the operator starts.

Fixes: #733

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,9 +29,6 @@ import (
 
 	"github.com/submariner-io/submariner-operator/pkg/apis"
 	"github.com/submariner-io/submariner-operator/pkg/controller"
-	"github.com/submariner-io/submariner-operator/pkg/engine"
-	"github.com/submariner-io/submariner-operator/pkg/lighthouse"
-	crdutils "github.com/submariner-io/submariner-operator/pkg/utils/crds"
 	"github.com/submariner-io/submariner-operator/pkg/version"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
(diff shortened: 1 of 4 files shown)


mangelajo:
Isn't this more cosmetical than something else?

I think handling the CRDs in the operator is better.  May be a workaround for now could be exiting gracefully after creating/updating the CRD instead?

Why does it happen? Controller and CRD creation start concurrently?

skitt:
> Isn't this more cosmetical than something else?

It’s largely cosmetic, but it doesn’t inspire confidence.

> I think handling the CRDs in the operator is better. May be a workaround for now could be exiting gracefully after creating/updating the CRD instead?

That doesn’t change the event sequence in the OpenShift console for example; it’s still weird that the operator starts, stops, and starts again.

> Why does it happen? Controller and CRD creation start concurrently?

The controller is started after the CRD creation, but doesn’t see the new CRD. I haven’t yet figured out why that happens.

mangelajo (changes requested):
Instead of reverting, can we try something like this: 

https://play.golang.org/p/Kiyr2vuDeah

This replaces the running process by a restarted version of himself, same environment, etc

```go
package main

import "syscall"
import "os"
import "fmt"

func main() {

	binary, lookErr := os.Executable()
	if lookErr != nil {
		panic(lookErr)
	}

	env := os.Environ()

	fmt.Println("Calling ourselves again...")

	execErr := syscall.Exec(binary, os.Args, env)
	if execErr != nil {
		panic(execErr)
	}

	fmt.Println("This should not ever happen!")
}
```

skitt:
See https://github.com/submariner-io/submariner-operator/pull/742

skitt:
Closing in favour of #742.

PR #742, opened 2020-10-09
Restart the operator if we update Lighthouse CRDs

skitt:
Instead of crashing and then restarting, or requiring the deployment
to create the ServiceDiscovery CRD, restart the operator after
creating or updating the Lighthouse CRDs.

Thanks to Miguel Angel Ajo Pelayo for the suggestion!

Fixes: #733

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


mangelajo (approved):
Thank you @skitt  :-) , with the help of the OS our code is simpler than trying to decipher the internal details of the operator SDK initialization :-)

nyechiel (approved):

tpantelis (approved):

Issue #753, opened 2020-10-13
Deleting the ServiceDiscovery CR doesn’t delete the corresponding deployments


skitt:
Creating a ServiceDiscovery CR causes the operator to start the corresponding deployments; deleting it should remove them, but doesn’t.

PR #756, opened 2020-10-14
Ensure that helpers set ownership correctly

skitt:
Helpers for ConfigMap, Deployment and Service need to set the
ownership when handling conflicts. This ensures that the updated
objects are correctly owned and garbage-collected when the owning
object is deleted.

The DaemonSet helper is handled in #752.

Fixes: #753

### DIFF
--- a/pkg/controller/helpers/helpers.go
+++ b/pkg/controller/helpers/helpers.go
@@ -82,7 +82,8 @@ func ReconcileDeployment(owner metav1.Object, deployment *appsv1.Deployment, req
 			for k, v := range deployment.Labels {
 				toUpdate.Labels[k] = v
 			}
-			return nil
+			// Set the owner and controller
+			return controllerutil.SetControllerReference(owner, toUpdate, scheme)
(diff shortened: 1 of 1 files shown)


pkg/controller/helpers/helpers.go
```diff
@@ -82,7 +82,8 @@ func ReconcileDeployment(owner metav1.Object, deployment *appsv1.Deployment, req
 			for k, v := range deployment.Labels {
 				toUpdate.Labels[k] = v
 			}
-			return nil
+			// Set the owner and controller
```
mangelajo, line 85: Github is not letting me add comments up there, but doesn't L68-L71 become unnecessary after this change?

skitt: I thought so too, but I’ve seen tests fail with those lines removed; I wanted to get the fix in, I’ll look into removing the theoretically redundant lines in a follow-up.

tpantelis (approved):

PR #762, opened 2020-10-19
Remove unnecessary git fetching

dfarrell07:
It seems these git fetches are not necessary.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -17,11 +17,6 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
-      - name: Fetch git tags
-        run: git fetch origin +refs/tags/*:refs/tags/*
 
       - name: Reclaim free space
         run: |


tpantelis (approved):

mkolesnik (commented):
As I mentioned in https://github.com/submariner-io/releases/pull/33#discussion_r507779871 This is actually necessary here to calculate the versions correctly (since we're using `git revision`) but it's not necessary in most other CI jobs

You'll only see it on the release once it merges (we'll get some weird version probably), perhaps the e2e can live without it

dfarrell07:
> As I mentioned in [submariner-io/releases#33 (comment)](https://github.com/submariner-io/releases/pull/33#discussion_r507779871) This is actually necessary here to calculate the versions correctly (since we're using `git revision`) but it's not necessary in most other CI jobs
> 
> You'll only see it on the release once it merges (we'll get some weird version probably), perhaps the e2e can live without it

Okay, thanks @mkolesnik! I removed the `release.yml` change but kept the `e2e.yml` change in the latest.

.github/workflows/e2e.yml
```diff
@@ -17,11 +17,6 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
-        with:
```
mangelajo, line 20: We added it because otherwise the shipyard scripts get a little bit confused about tags, but
may be it's not necessary for E2E, and just for releasing/tagging images?... let's see how it goes

dfarrell07: Yeah, I think that's the case. It's hard to imagine how there could be an issue with removing this from the E2E workflow that wouldn't already have shown up.

skitt (approved):

Issue #748, opened 2020-10-12
subctl benchmark: add “--quiet” option


nyechiel:
For users that may want to integrate the subctl benchmark tests in some other flow/system, it would be great to support a “--quiet” flag that will only present the results, without full logs.

/cc @mkolesnik

nyechiel:
Other alternative is to add a "verbose" option and default to the "quiet" version, were no logs are presented.

Issue #649, opened 2020-08-25
Make the subctl boolean flags more consistent


skitt:
Currently `subctl` has the following flags, across all sub-commands:

* `--globalnet` (default `false`);
* `--service-discovery` (default `true` now);
* `--disable-nat` (default `false`; the flag is slightly inaccurate too, it should refer to NAT-T);
* `--ipsec-debug` (default `false`);
* `--subm-debug` (default `false`, deprecated in favour of `--enable-pod-debugging`);
* `--enable-pod-debugging` (default `false`);
* `--no-label` (default `false`);
* `--enable-disruptive` (default `false`);
* `--verbose` (default `false`).

It would be more consistent if we followed these rules:
* boolean flags should be “positive”, *i.e.* setting them should enable;
* `no` or `enable`/`disable` prefixes shouldn’t be hard-coded (I’m working on adding them to the flag library);
* we should rely on `=false` to disable features.

This would result in the following changes:
* `--disable-nat` becomes `--natt`, default `true` (the current flag is deprecated, equivalent to `--natt=false`);
* `--enable-pod-debugging` becomes `--pod-debug` to match the other debug flags, default `false` (the current flag is deprecated, equivaltn to `--pod-debug`);
* `--no-label` becomes `--label-gateway`, default `true` (the current flag is deprecated, equivalent to `--label-gateway=false`);
* `--enable-disruptive` becomes `--disruptive-tests`, default `false`) (the current flag is deprecated, equivalent to `--enable-disruptive`).

skitt:
@manosnoam I’d appreciate your feedback on this one!

mangelajo:
Makes sense, coherence always helps.

manosnoam:
Thanks for asking @skitt 
Since I did not use all flags in my tests yet, can you update the descriptions with sections above related flags: 

- Flags for any subctl command
 ...
- Flags for deploy command
 ...
- Flags for join command
 ...

Regarding --ipsec-debug, is it like --enable-pod-debugging, or more like --verbose ?
In general, having few **flags for verbosity/debug**, is somewhat confusing. 
Can you have these flags with same naming convention ?

Regarding "--disable-option" (in conjunction with "--enable-option") flags, it seems better than "**--no**-option" flags.

We might even consider having just "disable" flags, and remove "enable", as it is obvious that --option defaults to option=true.
For example, change `--enable-pod-debugging` to `--pod-debugging`. 
[...]

PR #766, opened 2020-10-20
Subctl benchmark - add verbose option

maayanf24:
Add a verbose flag to subctl benchmark cmds. default=false
Fixes issue #748


pkg/subctl/cmd/benchmark.go
```diff
@@ -36,16 +37,19 @@ var (
...
+	addBenchmarkFlags(benchmarkThroughputCmd)
 
 	benchmarkCmd.AddCommand(benchmarkThroughputCmd)
 	benchmarkCmd.AddCommand(benchmarkLatencyCmd)
 	rootCmd.AddCommand(benchmarkCmd)
 }
 
+func addBenchmarkFlags(cmd *cobra.Command) {
+	cmd.PersistentFlags().BoolVar(&intraCluster, "intra-cluster", false, "Runs the test within a single cluster")
+	cmd.PersistentFlags().BoolVar(&verbose, "verbose", true, "Produce verbose logs during benchmark tests")
```
mangelajo, line 50: We normally have verbose = false in other parts
```suggestion
	cmd.PersistentFlags().BoolVar(&verbose, "verbose", false, "Produce verbose logs during benchmark tests")
```

mangelajo (changes requested):
To do the same as in verify I'd default verbose to false. (I tend to enable it always though..)

pkg/subctl/cmd/benchmark.go
skitt: Yes, see https://github.com/submariner-io/submariner-operator/issues/649 for general discussion.

skitt (approved):

pkg/subctl/benchmark/throughput.go
```diff
@@ -9,7 +9,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func StartThroughputTests(intraCluster bool) {
+var benchmark_verbose bool
```
tpantelis, line 12: I think we should only have  global var for this.  I suggest renaming this to `Verbose` and reference it in _pkg/subctl/cmd/benchmark.go_ and remove the `verbose` var there. Then no need to pass it around.

tpantelis (changes requested):

tpantelis (approved):

mangelajo (approved):
Looks good, but now in conflict because of the go.mod files sorry

PR #767, opened 2020-10-20
Use the MCS SIG APIs instead of our own

aswinsuryan:
Fixes:submariner-io/lighthouse#336


skitt (changes requested):
The Dockerfile shouldn’t be part of this PR; apart from that, LGTM!

You’ll also need to add the MCS YAML files to the linting exclusion list in `.yamllint.yml`, and it might be worth separating the old MCS removal in a separate PR.

mangelajo:
@aswinsuryan @vthapar question independent to this specific patch. At this point, where we are starting to care about updates, probably in Lighthouse we may want to look for old "ServiceExports", converting them to new services exports, and deleting the old ones. So existing deployments are not affected.

Dockerfile.dapper783451420
```diff
@@ -0,0 +1,22 @@
+FROM quay.io/submariner/shipyard-dapper-base:devel
```
mangelajo, line 1: This file should not be added :)

mangelajo (changes requested):
Looks good, just removing the dockerfile from dapper.

mangelajo:
Ahh I see @skitt already commented about the dockerfile ':D

Dockerfile.dapper783451420
aswinsuryan: Done . it was due to a  `git add .` . :)

aswinsuryan:
> The Dockerfile shouldn’t be part of this PR; apart from that, LGTM!
> 
> You’ll also need to add the MCS YAML files to the linting exclusion list in `.yamllint.yml`, and it might be worth separating the old MCS removal in a separate PR.

WIll make MCS removal a separate PR.

skitt (approved):

vthapar (approved):

tpantelis (approved):

mangelajo (approved):

PR #786, opened 2020-11-09
Add periodic E2E tests to find flaky failures

dfarrell07:
Because this runs against code that has already passed CI to be merged
to the main branch, any failures are likely to be sporadic flakes.

The frequency of this job should be turned up or down depending on how
actively we're hunting for flaky failures

### DIFF
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -3,7 +3,8 @@ name: Flake Finder
 
 on:
   schedule:
-    - cron: "0 0 * * *"
+    - cron: "0 0,1,2 * * *"
+    - cron: "0 0/2 * * 6"
 
 jobs:
   e2e:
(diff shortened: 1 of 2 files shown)


.github/workflows/flake_finder.yml
```diff
@@ -0,0 +1,54 @@
+---
+name: Flake Finder
```
dfarrell07, line 2: Do y'all like this job naming? If so, I'll make a parallel name change to the periodic E2E job in the main repo.

tpantelis: Sounds appropriate.

tpantelis (approved):

mangelajo (approved):

.github/workflows/flake_finder.yml
```diff
@@ -0,0 +1,55 @@
...
+    - cron: "0 0/2 * * 6"
+
+jobs:
+  e2e:
+    name: E2E
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
```
dfarrell07, line 16: This will run 12 jobs (2*2*3), not the 8 mentioned previously.

PR #802, opened 2020-11-12
Add ClusterRole for the networkplugin_syncer

tpantelis:


config/rbac/kustomization.yaml
```diff
@@ -9,6 +9,8 @@ resources:
   - role.yaml
   - role_binding.yaml
   - service_account.yaml
+  - networkplugin_syncer_cluster_role.yaml
+  - networkplugin_syncer_cluster_role_binding.yaml
```
tpantelis, line 13: What is this file for?

skitt: @SteveMattar you’ll probably be able to explain this better than me...

tpantelis: I put these in here assuming they're needed.

SteveMattar: @tpantelis its a list of rbac resources which `kustomize` is using to build the final bundle for OLM.
It's needed for operatorhub.io basically.

mangelajo: I believe that if you run `kubectl kustomize .`  in this directory, you will get the bundle.yaml out (which is used to install the operator just from a yaml)

mangelajo: but only the rbacs part

config/rbac/networkplugin_syncer_cluster_role.yaml
```diff
@@ -0,0 +1,25 @@
...
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+      - secrets
+      - nodes
+    verbs:
+      - get
+      - list
```
mangelajo, line 16: Can we remove - list for "secrets and configmaps", for nodes we need it.

We will limit a little bit the security this way.. at least the name of the secret/configmap needs to be know to be accessible.

pkg/subctl/operator/submarinerop/scc/ensure.go
```diff
@@ -23,5 +23,11 @@ import (
 )
 
 func Ensure(restConfig *rest.Config, namespace string) (bool, error) {
-	return scc.UpateSCC(restConfig, namespace, serviceaccount.OperatorServiceAccount)
+	operatorUserAdded, err := scc.UpateSCC(restConfig, namespace, serviceaccount.OperatorServiceAccount)
+	if err != nil {
+		return false, err
+	}
+
+	npSyncerUserAdded, err := scc.UpateSCC(restConfig, namespace, serviceaccount.NPSyncerServiceAccount)
```
mangelajo, line 31: This one doesn't need to be on the SCC, it hopefully won't run as a privileged pod. (as far as I have identified)

pkg/subctl/operator/submarinerop/scc/ensure.go
```diff
@@ -23,5 +23,11 @@ import (
 )
 
 func Ensure(restConfig *rest.Config, namespace string) (bool, error) {
-	return scc.UpateSCC(restConfig, namespace, serviceaccount.OperatorServiceAccount)
+	operatorUserAdded, err := scc.UpateSCC(restConfig, namespace, serviceaccount.OperatorServiceAccount)
```
mangelajo, line 26: unrelated to your patch, we have a typo in UpateSCC -> UpdateSCC

mangelajo (changes requested):
Thank you @tpantelis just a couple of changes.

pkg/subctl/operator/submarinerop/scc/ensure.go
tpantelis: ok - I wasn't sure what this stuff was for 😄

mangelajo: In openshift there are security contexts for service accounts, and for being allowed to escalate for use of hostnetwork pod, or privileged ports, you need to be on the privileged security context. ':)

pkg/subctl/operator/submarinerop/scc/ensure.go
tpantelis: Fixed the typo

config/rbac/kustomization.yaml
SteveMattar: You can run `make bundle`

mangelajo (approved):

mangelajo:
I have checked with : https://github.com/submariner-io/submariner/pull/943 and it's working as expected, the service account has the required permissions to work.

SteveMattar (commented):
lgtm
Just one suggestion:
I would rename and add a prefix "is" to those variables inside every ensure function just to indicate they are type boolean
for example: `operatorSACreated` --> `isOperatorSACreated`

tpantelis:
> lgtm
> Just one suggestion:
> I would rename and add a prefix "is" to those variables inside every ensure function just to indicate they are type boolean
> for example: `operatorSACreated` --> `isOperatorSACreated`

Or `createdOperatorSA` to be consistent with other parts of the code. I think that reads a little better.

SteveMattar (approved):

PR #829, opened 2020-11-16
Image override

### LINKED ISSUE #781: Support 'Image' in submariners API spec
In `submariners` API spec we allow currently to set a custom repository and custom image tag but the containers image **names** are not supported.
In some cases the components image names could be different than the default ones but currently those names are hard coded in the operator.
It would be a good idea to let the user decide which image names to use.
These options should be available in both the CRD and `subctl`

SteveMattar:
Add `ImageOverrides` to Submariners CRD which is a string map that overrides the images by component.
This field is meant to be used by CI, development, testing or hotfixes.

In addition add a list of variables for `go build` to support multiple default builds.

### Generates the default image names by adding a prefix/postfix
* ImagePrefix
* ImagePostfix

### Override default repository and default versions
* DefaultRepo
* DefaultSubmarinerOperatorVersion
* DefaultSubmarinerVersion
* DefaultLighthouseVersion

Closes #781 


pkg/versions/versions.go
```diff
@@ -6,3 +6,10 @@ const (
 	DefaultSubmarinerVersion         = "0.7.0"
 	DefaultLighthouseVersion         = "0.7.0"
 )
+
+var (
```
mangelajo, line 10: This complicates the code and we still would need to maintain those ^ constants.

My proposal is to simply use DefaultSubmarinerRepo, .... DefaultLighthouseVersion as variables and override those at compilation time.

mangelajo:
```var (
	DefaultSubmarinerRepo            = "quay.io/submariner"		DefaultSubmarinerRepo            = "quay.io/submariner"
	DefaultSubmarinerOperatorVersion = "0.7.0"		DefaultSubmarinerOperatorVersion = "0.7.0"
	DefaultSubmarinerVersion         = "0.7.0"		DefaultSubmarinerVersion         = "0.7.0"
	DefaultLighthouseVersion         = "0.7.0"		DefaultLighthouseVersion         = "0.7.0"
)	)

File: apis/submariner/v1alpha1/submariner_types.go [hunk 2]
```diff
@@ -120,11 +121,19 @@ func init() {
 func (submariner *Submariner) SetDefaults() {
 	if submariner.Spec.Repository == "" {
 		// An empty field is converted to the default upstream submariner repository where all images live
-		submariner.Spec.Repository = versions.DefaultSubmarinerRepo
+		if versions.OverrideDefaultRepo != "" {
```
mangelajo, line 124: See my comment down in versions for a simpler approach.

mangelajo (approved):
@SteveMattar thanks for working on this. A couple of comments.

mangelajo (changes requested):
I meant, requested changes.

pkg/versions/versions.go
SteveMattar: Great! I believe we have to change the release process as well right?

apis/submariner/v1alpha1/submariner_types.go
SteveMattar: Thanks @mangelajo I will get rid of these Override variables

skitt (approved):

pkg/names/names.go
```diff
@@ -4,5 +4,14 @@ const (
 	RouteAgentImage        = "submariner-route-agent"
 	EngineImage            = "submariner"
 	GlobalnetImage         = "submariner-globalnet"
+	ServiceDiscoveryImage  = "lighthouse-agent"
+	LighthouseCoreDNSImage = "lighthouse-coredns"
```
mangelajo, line 8: :+1:

mangelajo (approved):
Looks great :)

mkolesnik:
Thanks @SteveMattar, how to supply it from the command line?
cc @mangelajo

SteveMattar:
> Thanks @SteveMattar, how to supply it from the command line?
> cc @mangelajo

@mkolesnik  #869

PR #845, opened 2020-11-18
Exclude openshift-dns namespace in Globalnet

sridhargaddam:
In an OCP Cluster, openshift-dns namespace has dns-default
service and this is controlled by its operator.
When Globalnet is deployed on OCP, it was seen that globalip
annotation added to it is periodically getting deleted by the
operator, so Globalnet tries to re-add the annotation and this
goes on forever. This will cause Globalnet to consume CPU
unnecessarily and could affect user-experience with Submariner
Globalnet. We have plans to enhance Globalnet to improve its
scalability, but until then we can exclude annotating services
in openshift-dns namespace.


skitt (approved):

nyechiel:
@sridhargaddam While this makes total sense, should we just add the rest of the OCP default namespaces to the exclude list for now?

nyechiel (approved):

Issue #797, opened 2020-11-12
subctl benchmark to support Globalnet deployments


sridhargaddam:
**What would you like to be added**:
Currently, "subctl benchmark" supports throughput and latency tests for Vanilla Submariner.
We need support in "subctl benchmark" to support even Globalnet deployments.

**Why is this needed**:
This is necessary so that we can use "subctl benchmark" for Globalnet deployments.

sridhargaddam:
In order to validate latency between the clusters, subctl internally uses netperf tool.
Netperf is a tool that is designed as a client-server model. When a netperf client connects to a server, the first thing will be an establishment of the control connection. This control connection happens on TCP/12865 port.
Once the control connection is setup and configuration information is passed to the server, a separate connection is opened for validating the actual latency tests. The port numbers used for the new connection are not fixed and are exchanged dynamically during the initial control connection phase. Because of this, we cannot use netperf/netserver with K8s ClusterIP services where the port numbers have to be explicitly specified during service creation.

Currently, Globalnet supports only pod to remoteService communication. So, until we support Headless-Services/Stateful-Sets/Pod-to-Pod communication in Globalnet, we cannot benchmark latency using netperf.

skitt:
@sridhargaddam `netperf` can be used without a control connection, using the `-N` option. When this is invoked, the tests use well-known ports by default (echo, chargen, discard), and those ports can also be specified to different values on the command line. So I think it should be possible to fix the set of ports to use ahead of time.

sridhargaddam:
> @sridhargaddam `netperf` can be used without a control connection, using the `-N` option. When this is invoked, the tests use well-known ports by default (echo, chargen, discard), and those ports can also be specified to different values on the command line. So I think it should be possible to fix the set of ports to use ahead of time.

Thanks @skitt, but it does not seem to work.

Running the netperf client, normally:
```
bash-5.0# netperf -H 10.3.128.3 -t TCP_RR -p 12865 -- -o min_latency,mean_latency,max_latency,stddev_latency,transaction_rate
MIGRATED TCP REQUEST/RESPONSE TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 10.3.128.3 (10.3.12) port 0 AF_INET : first burst 0
Minimum Latency Microseconds,Mean Latency Microseconds,Maximum Latency Microseconds,Stddev Latency Microseconds,Transaction Rate Tran/s
86,118.91,2874,53.18,8390.342
```

[...]

PR #903, opened 2020-11-23
Update subctl benchmark to support Globalnet

maayanf24:
Fixes issue #797


pkg/subctl/benchmark/latency.go
```diff
@@ -65,9 +71,8 @@ func StartLatencyTests(intraCluster bool) {
 func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
 	clusterAName := framework.TestContext.ClusterIDs[testParams.ClientCluster]
 	clusterBName := framework.TestContext.ClusterIDs[testParams.ServerCluster]
-	var connectionTimeout uint = 20
-	var connectionAttempts uint = 3
-	var netperfPort = 12865
+	var connectionTimeout uint = 5
```
sridhargaddam: Using 5 seconds can cause timing issues on some platforms. The default seems to be 60 secs and connectionAttempts as 2. Can we re-use the defaults here?

Also, IMO we should support these arguments so that users can override them if required. WDYS?

pkg/subctl/benchmark/latency.go
```diff
@@ -32,6 +32,12 @@ func StartLatencyTests(intraCluster bool) {
 	f = initFramework("latency")
 
 	if !intraCluster {
+		if framework.TestContext.GlobalnetEnabled {
```
sridhargaddam, line 35: What would happen if user tries to run latency tests with intracluster on a Globalnet deployment?

pkg/subctl/benchmark/latency.go
maayanf24: We can revert to use defaults.
Do you mean to allow users to supply those arguments from command-line? 
I think it's a good idea

sridhargaddam: Yes, I was thinking it might be a good idea to allow users to override these params during runs (if required). But for now, IMO lets use defaults.

pkg/subctl/benchmark/latency.go
maayanf24: The tests should work properly in this case

pkg/subctl/benchmark/latency.go
maayanf24: currently, the code is not using connectionTimeout and connectionAttempts params:
https://github.com/submariner-io/shipyard/blob/6a023e910387a2bf9f17ad9255efb14439b66b37/test/e2e/framework/network_pods.go#L339

Do you think it's need to be added?

sridhargaddam: Basically when running tests aganist Globalnet deployment, the initial connection might fail because of the way we are creating the connectorPods as it takes few secs for the globalIP annotation to be applied on the connectorPods. So, its more important in Globalnet deployments that we retry atleast twice (i.e., connectionAttempts = 2).

I see that we have disabled Latency tests for Globalnet deployments, so its fine for now. But in a future PR please consider supporting connectionTimeout/connectionAttempts when Latency tests are enabled for Globalnet.

skitt (approved):

pkg/subctl/benchmark/throughput.go
```diff
@@ -95,20 +96,33 @@ func runThroughputTest(f *framework.Framework, testParams benchmarkTestParams) {
...
 	})
 
 	podsClusterB := framework.KubeClients[testParams.ServerCluster].CoreV1().Pods(f.Namespace)
 	p1, _ := podsClusterB.Get(nettestServerPod.Pod.Name, metav1.GetOptions{})
 	framework.By(fmt.Sprintf("Nettest Server Pod %q was created on node %q", nettestServerPod.Pod.Name, nettestServerPod.Pod.Spec.NodeName))
 
 	remoteIP := p1.Status.PodIP
+
+	if framework.TestContext.GlobalnetEnabled && testParams.ClientCluster != testParams.ServerCluster {
+		By(fmt.Sprintf("Pointing a service ClusterIP to the listener pod in cluster %q",
```
sridhargaddam, line 109:
```suggestion
		By(fmt.Sprintf("Pointing a ClusterIP service to the listener pod in cluster %q",
```

sridhargaddam (approved):
One small nit, otherwise LGTM. Thanks @maayanf24

pkg/subctl/benchmark/throughput.go
maayanf24: done

PR #904, opened 2020-11-24
Remove hard-coded override of cable driver flag

dfarrell07:
Don't pass cable driver flag, instead use new Shipyard using= flag.

As a side effect, this will change the default cable driver in this repo
from strongswan to libreswan.

Relates-to: submariner-io/shipyard#364

### DIFF
--- a/scripts/kind-e2e/lib_operator_verify_subm.sh
+++ b/scripts/kind-e2e/lib_operator_verify_subm.sh
@@ -35,7 +35,7 @@ function create_subm_vars() {
     subm_colorcodes=blue
     subm_debug=false
     subm_broker=k8s
-    subm_cabledriver=strongswan
+    subm_cabledriver=libreswan
     ce_ipsec_debug=false
     ce_ipsec_ikeport=500
     ce_ipsec_nattport=4500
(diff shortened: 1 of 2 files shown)


dfarrell07:
This needs a version of Shipyard with https://github.com/submariner-io/shipyard/pull/365.

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -35,7 +35,7 @@ function create_subm_vars() {
     subm_colorcodes=blue
     subm_debug=false
     subm_broker=k8s
-    subm_cabledriver=strongswan
+    subm_cabledriver=libreswan
```
mangelajo, line 38: So we will default to libreswan?

I think if we want to make this change it should be on the operator itself, when we do it.

mangelajo (dismissed):

scripts/kind-e2e/lib_operator_verify_subm.sh
dfarrell07: https://github.com/submariner-io/shipyard/pull/365 will set libreswan as the default. You're saying we should set it to strongswan there?

skitt: The _default_ default (_i.e._, the cable driver used when Submariner isn’t told which to use) is determined by Submariner itself, see https://github.com/submariner-io/submariner/pull/972 which changes it to Libreswan. The operator can have a different default, as can `subctl` and the various test suites ;-). But it’s better IMO if we align everything on Libreswan for 0.8.

skitt (approved):

skitt:
The tests fails because of

```
[submariner-operator]$ [cluster2] validate_equals .spec.cableDriver libreswan
[submariner-operator]$ [cluster2] validate_equals .spec.cableDriver libreswan
[submariner-operator]$ [cluster2] local json_filter=.spec.cableDriver
[submariner-operator]$ [cluster2] local expected=libreswan
[submariner-operator]$ [cluster2] local actual=null
[submariner-operator]$ [cluster2] jq -r -M .spec.cableDriver /tmp/submariner.cluster2.json
[cluster2] Expected 'libreswan' but got 'null'
```

mkolesnik (approved):

PR #938, opened 2020-11-29
feat(rbac): rationalise service accounts

### LINKED ISSUE #679: Rationalise service accounts
We have a number of service accounts with different configurations, but probably not enough (and the permissions requested aren’t well-documented).

My thoughts on this are that the operator should have a single service account with all the appropriate privileges, and then our applications should have their own service accounts. If privileges vary depending on the deployment (*e.g.* with or without Lighthouse or Globalnet), we should modify the service account accordingly instead of adding a `:`-prefixed variant (or is that idiomatic K8s?).

SteveMattar:
Create a service account for each component.
The operator will have a single service account with all the appropriate privileges, and same for the applications which will have their own service accounts.
Modify the RBAC definitions respectively.

Closes #679

BREAKING CHANGE


SteveMattar:
The e2e test is failing because it does not use my PR version of `subctl`.

skitt (changes requested):
The breakage should be handled in this PR as far as possible, and what isn’t, should be split into a separate PR (_i.e._ this PR should add whatever new roles are needed, and change code to use them, without reducing existing roles; a subsequent PR can then reduce the existing roles, and handle upgrades).

mangelajo (approved):
Thanks for handling this Steve!

skitt (approved):

tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
Have we verified the Helm side of this? I guess it will not be as big a deal with the new submariner-operator Chart, but the other Charts probably will be impacted (and might should just be deprecated/removed).

SteveMattar:
> Have we verified the Helm side of this? I guess it will not be as big a deal with the new submariner-operator Chart, but the other Charts probably will be impacted (and might should just be deprecated/removed).

Thanks for your comment @dfarrell07, I will create another PR for handling the Helm charts

mangelajo (approved):

PR #939, opened 2020-11-29
fix: update image pull policy when using image override

SteveMattar:
The `ImagePullPolicy` will be set to the correct value according to the image tag
in case of using `ImageOverride` flag.

### DIFF
--- a/controllers/servicediscovery/servicediscovery_controller.go
+++ b/controllers/servicediscovery/servicediscovery_controller.go
@@ -184,7 +184,7 @@ func newLighthouseAgent(cr *submarinerv1alpha1.ServiceDiscovery) *appsv1.Deploym
 						{
 							Name:            "submariner-lighthouse-agent",
 							Image:           getImagePath(cr, names.ServiceDiscoveryImage),
-							ImagePullPolicy: images.GetPullPolicy(cr.Spec.Version),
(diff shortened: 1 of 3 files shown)


controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -472,6 +472,14 @@ func getImagePath(submariner *submarinerv1alpha1.ServiceDiscovery, componentImag
 		submariner.Spec.ImageOverrides)
 }
 
+func getPullPolicy(submariner *submarinerv1alpha1.ServiceDiscovery, componentImage string) corev1.PullPolicy {
```
skitt, line 475: Can’t this be shared?

skitt (changes requested):

controllers/servicediscovery/servicediscovery_controller.go
SteveMattar: Sure

mangelajo (approved):

dfarrell07 (approved):

skitt (approved):

skitt:
The CRD linting failure is fixed by https://github.com/submariner-io/submariner-operator/pull/1011.

PR #961, opened 2020-12-07
Add controller-gen as a pkg

SteveMattar:
This should be handy for the d/s ci only.
The idea is to use `go run pkg/controller-gen/main.go` instead.
`make embeddedyamls` is based on `controller-gen` which unfortunately does not
have an RPM or Brew package at the moment and can only be downloaded as part of
the go module `sigs.k8s.io/controller-tools`.
When running the d/s CI there is no access to the internet and the only way to
use this tool is by explicitly adding this package to the project.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20200204173128-addea2498afe
 	sigs.k8s.io/controller-runtime v0.6.2
+	sigs.k8s.io/controller-tools v0.3.0
 	sigs.k8s.io/mcs-api v0.0.0-20200908023942-d26176718973
 	sigs.k8s.io/yaml v1.2.0
 )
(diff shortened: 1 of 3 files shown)


dfarrell07:
Maybe we can expand on the reasoning for how it will be useful downstream and why we should include it here? It would be good to have that info in the commit history imho.

SteveMattar:
`make embeddedyamls` is based on `controller-gen` which unfortunately does not have an RPM or Brew package at the moment and can only be downloaded as part of the go module `sigs.k8s.io/controller-tools`
When running the d/s CI there is no access to the internet and the only way to use this tool is by explicitly adding this package to the project.
I can then use `go run pkg/controller-gen/main.go`

dfarrell07:
> `make embeddedyamls` is based on `controller-gen` which unfortunately does not have an RPM or Brew package at the moment and can only be downloaded as part of the go module `sigs.k8s.io/controller-tools`
> When running the d/s CI there is no access to the internet and the only way to use this tool is by explicitly adding this package to the project.
> I can then use `go run pkg/controller-gen/main.go`

Thanks @SteveMattar, can you add that to the commit message? :pray:

dfarrell07 (approved):

skitt (approved):

PR #962, opened 2020-12-08
Fix cable driver false test failure

dfarrell07:
The verification scripts from when the operator was being developed
aren't able to respond to the using=<cable driver> flag, and so fail
when the cable driver doesn't match their default.


dfarrell07:
I'm surprised the CI passed in the PR that introduced this https://github.com/submariner-io/submariner-operator/pull/904 but failed in the Flake Finder https://github.com/submariner-io/submariner-operator/actions/runs/407204797.

mkolesnik (approved):
I think generally a lot of this stuff belongs in a more "unit test" type of testing rather than E2E, we should consider moving it out.

tpantelis (approved):

PR #971, opened 2020-12-09
Make boolean flags consistent

skitt:
This applies the following rules:
* boolean flags should be positive, i.e. setting them should enable;
* "no" and "enable"/"disable" prefixes shouldn't be hard-coded;
* we should rely on "=false" to disable features.

As a result:
* --disable-nat is replaced by --natt=false
* --enable-pod-debugging is replaced by --pod-debug
* --no-label is replaced by --label-gateway=false
* --health-check-enable is replaced by --health-check
* --enable-disruptive is replaced by --disruptive-tests

The old variants are still supported but will result in a deprecation
warning.

Fixes: #649


pkg/subctl/cmd/join.go
```diff
@@ -106,8 +121,13 @@ func addJoinFlags(cmd *cobra.Command) {
 		"List of domains to use for multicluster service discovery")
 	cmd.Flags().StringSliceVar(&imageOverrideArr, "image-override", nil,
 		"Override component image")
+	cmd.Flags().BoolVar(&healthCheckEnable, "health-check", true,
+		"enable Gateway health check")
```
tpantelis, line 125: This one hasn't released yet so I think we should just change this for 0.8.0 rather than deprecating.

skitt: Good point, I’ll split it out so we can merge it for 0.8.

skitt: https://github.com/submariner-io/submariner-operator/pull/975

tpantelis (approved):

dfarrell07 (approved):

nyechiel:
@skitt our docs need to be updated to reflect these changes: https://submariner.io/operations/deployment/subctl/

skitt:
> @skitt our docs need to be updated to reflect these changes: https://submariner.io/operations/deployment/subctl/, https://submariner.io/getting-started/quickstart/kind/#join-cluster2-and-cluster3-to-the-broker

https://github.com/submariner-io/submariner-website/pull/406

tpantelis:
This was merged as part of https://github.com/submariner-io/submariner-operator/pull/976.

PR #973, opened 2020-12-10
chore: do not merge

ccamacho:
This patch is only used
for testing the CI integration
with KubeInit.

Please, do not merge.


skitt:
@ccamacho for patches to test CI, does it work if you publish them as drafts? That way there’s no risk they can be merged, and people don’t get notified about them.

ccamacho:
> @ccamacho for patches to test CI, does it work if you publish them as drafts? That way there’s no risk they can be merged, and people don’t get notified about them.

yes, sorry for that Ill send them as drafts.

Issue #999, opened 2020-12-18
Detect OVNKubernetes when no cluster network operator present


mangelajo:
**What happened**:

When installing ovn-kubernetes in kind [1] it's not done via the cluster network operator,
so we need to improve our network plugin detection heuristics to support OVNKubernetes
in such case.

The OVN_NBDB and OVN_SBDB endpoints should be detected correctly and passed down to the network plugin syncer.
[1] https://github.com/submariner-io/shipyard/pull/387

**What you expected to happen**:

OVNKubernetes detected correctly.

**How to reproduce it (as minimally and precisely as possible)**:

`make clusters` with [1] 
**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`):
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #1001, opened 2020-12-18
Detect OVNKubernetes without Cluster Network Operator

mangelajo:
When running in kind, and possibly other clusters OVNKubernetes
works standalone and we need to find the details in the system,
so we can pass them down to the other services.

Fixes-Issue: #999

### DIFF
--- a/controllers/submariner/submariner_controller.go
+++ b/controllers/submariner/submariner_controller.go
@@ -127,7 +127,8 @@ func (r *SubmarinerReconciler) Reconcile(request reconcile.Request) (reconcile.R
 
 	// discovery is performed after Update to avoid storing the discovery in the
 	// struct beyond status
-	if err := r.discoverNetwork(instance); err != nil {
+	clusterNetwork, err := r.discoverNetwork(instance)
+	if err != nil {
 		return reconcile.Result{}, err
 	}
 
@@ -148,7 +149,7 @@ func (r *SubmarinerReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		}
 	}
 
-	if _, err := r.reconcileNetworkPluginSyncerDeployment(instance, reqLogger); err != nil {
+	if _, err := r.reconcileNetworkPluginSyncerDeployment(instance, clusterNetwork, reqLogger); err != nil {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/discovery/network/ovnkubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		}
+	}
+
+	dbConnection := fmt.Sprintf("%s://%s.%s", dbConnectionProtocol, ovnKubeService, ovnDbPod.Namespace)
+
+	clusterNetwork := &ClusterNetwork{
+		NetworkPlugin: "OVNKubernetes",
+		PluginSettings: map[string]string{
+			OvnNBDB: dbConnection + ":6641",
```
vthapar, line 45: Shouldn't we use a constant for these?

pkg/discovery/network/ovnkubernetes_test.go
```diff
@@ -0,0 +1,105 @@
+/*
+© 2019 Red Hat, Inc. and others.
```
vthapar, line 2: 2020

vthapar (changes requested):
minor nitpicks

controllers/submariner/submariner_controller.go
```diff
@@ -306,10 +307,11 @@ func (r *SubmarinerReconciler) reconcileEngineDaemonSet(instance *submopv1a1.Sub
 }
 
 func (r *SubmarinerReconciler) reconcileNetworkPluginSyncerDeployment(instance *submopv1a1.Submariner,
-	reqLogger logr.Logger) (*appsv1.Deployment, error) {
+	clusterNetwork *network.ClusterNetwork, reqLogger logr.Logger) (*appsv1.Deployment, error) {
 	// Only OVNKubernetes needs networkplugin-syncer so far
 	if instance.Status.NetworkPlugin == "OVNKubernetes" {
```
sridhargaddam, line 312: Can we create a constant for OVNKubernetes and use it here as well as pkg/discovery/network/ovnkubernetes.go file

sridhargaddam (approved):
Verified locally and it works fine. Small nit, otherwise LGTM. Thanks @mangelajo

pkg/discovery/network/ovnkubernetes.go
mangelajo: We don't replay them in more places, for OVNKubernetes yes but we talked about adding all those to admiral as a common place (it's operator + subm depending on those), also the default port for openshift is different, but that one is detected with the openshift cluster network operator.

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+		for _, envVar := range container.Env {
+			if envVar.Name == "OVN_SSL_ENABLE" {
+				if strings.ToUpper(envVar.Value) != "NO" {
+					dbConnectionProtocol = "ssl"
+				}
+			}
+		}
+	}
+
+	dbConnection := fmt.Sprintf("%s://%s.%s", dbConnectionProtocol, ovnKubeService, ovnDbPod.Namespace)
```
mangelajo, line 40:
```suggestion
	dbConnection := fmt.Sprintf("%s:%s.%s", dbConnectionProtocol, ovnKubeService, ovnDbPod.Namespace)
```

pkg/discovery/network/ovnkubernetes.go
mangelajo: Ok, converted them to constants, no kitties were harmed :D :-P :)

pkg/discovery/network/ovnkubernetes_test.go
mangelajo: oops wait that one.

mangelajo: Done

controllers/submariner/submariner_controller.go
mangelajo: oh let me handle that while I'm at it...

dfarrell07 (approved):

controllers/submariner/submariner_controller.go
mangelajo: done

skitt (approved):

PR #1008, opened 2020-12-22
Ensure that CRDs are up-to-date

skitt:
CRDs should be updated along with any change which causes them to go
out-of-date, e.g. a version bump on Submariner which results in an
update to the underlying types.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -5,6 +5,17 @@ on:
   pull_request:
 
 jobs:
+  crds:
+    name: CRDs up-to-date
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):
Nice check, thanks @skitt!

mkolesnik (approved):

skitt:
It should fail though, the CRDs here _aren’t_ up-to-date...

skitt:
> It should fail though, the CRDs here _aren’t_ up-to-date...

Except it shouldn’t, CI is testing the _merge_ with master which is fixed.

PR #1013, opened 2020-12-23
test(linting): find broken links in the bundle csv yaml file

SteveMattar:
Create a temporary markdown file called `bundle.md` by extracting and composing the `.spec.description`, `.spec.links` fields from the csv yaml file.
Include this file as part of markdown-link-check test.

**Note**: The combination of both `check-modified-files-only=true` and `file-path=bundle.md` does not work.
see this [issue](https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/72).
I had to create an additional step for calculating the modified files and append my `bundle.md` to that list.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -63,16 +63,34 @@ jobs:
   markdown-link-check:
     name: Markdown Links (modified files)
     runs-on: ubuntu-latest
+    env:
+      BASE_BRANCH: ${{ github.base_ref }}
+      FILE_EXTENSION: .md
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
-
+      - name: Extract markdown from CSV.yaml
+        run: |-
+          sudo snap install yq --channel=v4/stable && \
+          yq e '.spec.description, .spec.links' config/manifests/bases/submariner.clusterserviceversion.yaml > extracted_yaml_links.md
+      - name: Prepare a list of md files
+        run: |-
+          git fetch origin "${BASE_BRANCH}" --depth=1 > /dev/null
+          MASTER_HASH=$(git rev-parse origin/"${BASE_BRANCH}")
+          mapfile -t FILE_ARRAY < <( git diff --name-only --diff-filter=AM "$MASTER_HASH" )
+          for i in "${FILE_ARRAY[@]}"; do
+            if [ "${i##*.}" == "${FILE_EXTENSION#.}" ]; then
+              MD_FILES="${MD_FILES}${i},"
+            fi
+          done
+          export MD_FILES="${MD_FILES}extracted_yaml_links.md"
+          echo "MD_FILES=${MD_FILES}" >> $GITHUB_ENV
       - name: Run markdown-link-check
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
           config-file: ".markdownlinkcheck.json"
-          check-modified-files-only: "yes"
-          base-branch: ${{ github.base_ref }}
+          check-modified-files-only: "no"
+          file-path: ${{ env.MD_FILES }}
 
   markdownlint:
     name: Markdown


.github/workflows/linting.yml
```diff
@@ -45,12 +45,16 @@ jobs:
...
+      - name: Extract markdown from CSV.yaml
+        run: |-
+          sudo snap install yq && \
+          yq r config/manifests/bases/submariner.clusterserviceversion.yaml '.spec.description, .spec.links' > bundle.md
       - name: Run markdown-link-check
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
           config-file: ".markdownlinkcheck.json"
           check-modified-files-only: "yes"
+          file-path: "bundle.md"
```
dfarrell07, line 57: I think this will cause markdownlink-check to **only** check this file, instead of all the files **and** this file.

SteveMattar: @dfarrell07 according to [this documentation](https://github.com/gaurav-nelson/github-action-markdown-link-check) it says "specify additional files" but I think you are right I just saw the logs and it looks like it works only on this file. Confusing a little bit...

skitt (commented):
Given https://github.com/submariner-io/submariner-operator/pull/1014, shouldn’t this PR cause the linting GHA to fail?

SteveMattar:
> Given #1014, shouldn’t this PR cause the linting GHA to fail?

@dfarrell07 can you tell why `markdown-link-check` is not catching those broken links?
example: `curl -I https://submariner.io/architecture/route-agent` returns 404

SteveMattar:
@skitt @dfarrell07 please see the note in the description

skitt (changes requested):
Please include the information from the PR description in the commit message itself.

skitt (approved):
Let’s see if merging #1014 lets this PR pass ;-).

SteveMattar:
@skitt it works. The bundle broken links pass the test now, but it found one additional broken link in `README.md` file.
I will open another PR for that.

.github/workflows/linting.yml
```diff
@@ -63,16 +63,34 @@ jobs:
...
+    env:
+      BASE_BRANCH: ${{ github.base_ref }}
+      FILE_EXTENSION: .md
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
-
+      - name: Extract markdown from CSV.yaml
+        run: |-
+          sudo snap install yq --channel=v4/stable && \
```
dfarrell07, line 74: If we just did this link extraction, creating a new (uncommitted) file, would enabling the GHA's built-in `check-modified-files-only` flag include it in the checks?

If so, it seems we could do that and avoid creating our own modified-f-MD-iles finding logic.

.github/workflows/linting.yml
```diff
@@ -63,16 +63,34 @@ jobs:
...
+          yq e '.spec.description, .spec.links' config/manifests/bases/submariner.clusterserviceversion.yaml > bundle.md
```
dfarrell07, line 75:
```suggestion
          yq e '.spec.description, .spec.links' config/manifests/bases/submariner.clusterserviceversion.yaml > extracted_yaml_links.md
```

That `bundle.md` name is not super clear to me, maybe something more verbose would be quicker to understand.

.github/workflows/linting.yml
```diff
@@ -63,16 +63,34 @@ jobs:
   markdown-link-check:
     name: Markdown Links (modified files)
```
dfarrell07, line 64: I guess we're not strictly checking only modified files in the current setup.  PRs that don't touch the CSV file still cause that file to be checked.

.github/workflows/linting.yml
dfarrell07, line 75: Maybe we should only create/check the YAML links if that file is modified? Maybe we should always check it in the periodic link check test?

dfarrell07 (changes requested):

.github/workflows/linting.yml
SteveMattar: @dfarrell07 unfortunately no, this was my first approach but this action compares between committed files

.github/workflows/linting.yml
SteveMattar: Yes, you are right and I can fix that really easy, however don't you think we should check all the links every period of time?
The links could be broken even if the file has never been touched.

.github/workflows/linting.yml
SteveMattar: I think the second option covers more.

.github/workflows/linting.yml
dfarrell07: Per discussion on Automation Sync, the idea is to check all files in the [Periodic workflow](https://github.com/submariner-io/submariner-operator/blob/master/.github/workflows/periodic.yml) and only modified files here.

.github/workflows/linting.yml
dfarrell07: Per discussion on the Automation Sync, an option is to commit the extracted links file locally in the GHA runner (`git commit -am "Extracted links in YAML for linting"`). It shouldn't have any side-effects, and would allow us to avoid creating modified-MD-files logic ourselves.

.github/workflows/linting.yml
```diff
@@ -63,16 +63,34 @@ jobs:
...
+          export MD_FILES="${MD_FILES}extracted_yaml_links.md"
+          echo "MD_FILES=${MD_FILES}" >> $GITHUB_ENV
       - name: Run markdown-link-check
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
           config-file: ".markdownlinkcheck.json"
-          check-modified-files-only: "yes"
-          base-branch: ${{ github.base_ref }}
+          check-modified-files-only: "no"
+          file-path: ${{ env.MD_FILES }}
```
dfarrell07, line 93: If we do the links file creation and commit design, I guess we don't need this files-path flag and should revert the modified-files-only flag change.

nyechiel:
@SteveMattar what's the status of this one?

SteveMattar:
Marked as Draft.
The implementation should be done in the Periodic workflow.

Issue #958, opened 2020-12-03
Subctl output when executed on a cluster without Submariner


sridhargaddam:
**What happened**:
This is not really a bug, but reporting it just to give a friendly message to the users when "subctl" commands are executed on a cluster that does not have Submariner installed.

Currently, running subctl commands on a cluster where Submariner is not installed returns the following message.
Can we include a note or a tip in the error message saying that Submariner may not be installed on the cluster?

[sgaddam@localhost submariner-operator]$ subctl show all
Showing information for cluster "cluster2":
    Discovered network details:
        Network plugin:  weave-net
        Service CIDRs:   [100.0.0.0/16]
        Cluster CIDRs:   [10.0.0.0/16]
Error obtaining the Submariner resource: the server could not find the requested resource (get submariners.submariner.io submariner)
subctl version: v0.8.0-rc0

Issue #957, opened 2020-12-03
When `subctl show X` fails, it stops


mkolesnik:
**What happened**:
Running `subctl show X` (Except `subctl show networks` which seems unaffected) when a cluster doesn't have submariner installed, causes it to fail and abort enumerating further clusters.

```shell
[root@6a37e385eb23 submariner-operator]# subctl show versions

Showing information for cluster "cluster1":
Unable to get the Submariner versions: the server could not find the requested resource (get submariners.submariner.io submariner)

subctl version: v0.8.0-rc0-1-g601ba7c
```

**What you expected to happen**:
It should fail for that cluster, but continue enumerating the other clusters.
Perhaps a different return code can be used to indicate this type of failure.

```shell
[root@6a37e385eb23 submariner-operator]# subctl show versions

Showing information for cluster "cluster1":
Unable to get the Submariner versions: the server could not find the requested resource (get submariners.submariner.io submariner)

Showing information for cluster "cluster2":
COMPONENT                       REPOSITORY                                            VERSION         
submariner                      quay.io/submariner                                    0.8.0-rc0       
submariner-operator             quay.io/submariner                                    0.8.0-rc0       
service-discovery               quay.io/submariner                                    0.8.0-rc0       

subctl version: v0.8.0-rc0
```

**How to reproduce it (as minimally and precisely as possible)**:
Deploy clusters manually according to the KIND quickstart:
```shell
make shell
make clusters
subctl deploy-broker --kubeconfig output/kubeconfigs/kind-config-cluster1 --service-discovery
subctl join --kubeconfig output/kubeconfigs/kind-config-cluster2 broker-info.subm --clusterid cluster2 --disable-nat
```

Then set KUBECONFIG so that the "broker" cluster is first:
```shell
export KUBECONFIG=/go/src/github.com/submariner-io/submariner-operator/output/kubeconfigs/kind-config-cluster1:/go/src/github.com/submariner-io/submariner-operator/output/kubeconfigs/kind-config-cluster2
```

Now any `subctl show` command (except networks) will fail on the first cluster and exit.

**Anything else we need to know?**:

**Environment**:
- Submariner version (use `subctl version`): v0.8.0-rc0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

[...]

PR #1017, opened 2020-12-31
Proper error on Submariner not installed

Jaanki:
This patch does the follwong:
1. Print proper error message while running
   subctl show * when submariner is not
   installed/removed in/from the cluster
2. Don't exit when subctl show * fails on
   a cluster rather enumerate other clusters

Closes: #957
Closes: #958


pkg/subctl/cmd/show_connections.go
```diff
@@ -36,30 +36,35 @@ func getConnectionsStatus(config *rest.Config) []connectionStatus {
...
+		if ! SubmarinerNotInstalled(err) {
+			exitOnError("Error obtaining the Submariner resource", err)
+		}
 	}
 
-	gateways := existingCfg.Status.Gateways
-	if gateways == nil {
-		exitWithErrorMsg("No endpoints found")
-	}
+	if SubmarinerInstalled == true {
```
maayanf24, line 49: It would be more readable if you change to:
`if SubmarinerInstalled == false {
    return status
  }`
  
and then continue the rest of the code.
Otherwise looking good

maayanf24 (changes requested):

pkg/subctl/cmd/show_connections.go
```diff
@@ -36,30 +36,35 @@ func getConnectionsStatus(config *rest.Config) []connectionStatus {
 	submarinerClient, err := submarinerclientset.NewForConfig(config)
 	exitOnError("Unable to get the Submariner client", err)
 
+	SubmarinerInstalled = true
 	var status []connectionStatus
 
 	existingCfg, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, v1.GetOptions{})
 	if err != nil {
-		exitOnError("error reading from submariner client", err)
+		if ! SubmarinerNotInstalled(err) {
```
mkolesnik, line 44: This "double negative" is a knows anti patten, why not name the function "submarinerMissing" and the "if" here will only be true if it is.

pkg/subctl/cmd/show.go
```diff
@@ -37,6 +39,10 @@ type restConfig struct {
 	clusterName string
 }
 
+var submarinerNotInstalledError = errors.New("the server could not find the requested resource (get submariners.submariner.io submariner)")
+var submarinerUninstalledError = errors.New("submariners.submariner.io \"submariner\" not found")
```
mkolesnik, line 43: Why save there as Error types and then just use `.Error()` to get the string, instead of just saving the string?
In addition, why save as variables at all? They don't seem to be set anywhere, nor are they used more than once. Why not just hard code the strings?

pkg/subctl/cmd/show_connections.go
```diff
@@ -36,30 +36,35 @@ func getConnectionsStatus(config *rest.Config) []connectionStatus {
 	submarinerClient, err := submarinerclientset.NewForConfig(config)
 	exitOnError("Unable to get the Submariner client", err)
 
+	SubmarinerInstalled = true
```
mkolesnik, line 39: This seems like a weird way to implement conditional flow control, I'm sure you can do it without the variable.
In fact you can just return the empty list `status` and the code will work just fine.

mkolesnik (changes requested):

pkg/subctl/cmd/show.go
Jaanki: for easy reading

pkg/subctl/cmd/show_connections.go
Jaanki: yes, it will work but the output when submariner is not installed won't be pretty.

```
$ ./../submariner-operator/bin/subctl show endpoints

Showing information for cluster "cluster3":
Submariner is not installed
subctl version: v0.8.0-18-g53608ad-dev
No resources found.

Showing information for cluster "cluster2":
CLUSTER ID                    ENDPOINT IP     PUBLIC IP       CABLE DRIVER        TYPE            
cluster2                      172.17.0.9      49.206.37.124   libreswan           local           

Showing information for cluster "cluster1":
Submariner is not installed
subctl version: v0.8.0-18-g53608ad-dev
No resources found.
```
Notice subctl version line before 'No resources found'

Whereas in current implementation, the output is
```
$ ./../submariner-operator/bin/subctl show endpoints

Showing information for cluster "cluster3":
Submariner is not installed
subctl version: v0.8.0-18-g53608ad-dev

Showing information for cluster "cluster2":
CLUSTER ID                    ENDPOINT IP     PUBLIC IP       CABLE DRIVER        TYPE            
cluster2                      172.17.0.9      49.206.37.124   libreswan           local           

Showing information for cluster "cluster1":
Submariner is not installed
subctl version: v0.8.0-18-g53608ad-dev
```

I guess it comes down to the way we want the output.

pkg/subctl/cmd/show_connections.go
Jaanki: The logic here is ```if there is error and submariner is installed, exitOnError otherwise exit current cluster and continue with other clusters``` Function name, IMO, won't matter. Instead, I changed the function to ```SubmarinerIsInstalled```.

pkg/subctl/cmd/show_connections.go
Jaanki: I personally would prefer ```subctl version``` line to be the last which sort of says thats the end of information for the given cluster.

mkolesnik: There should be no `subctl version` printout at all as it's only printed out on error (and this is expected)

pkg/subctl/cmd/show.go
mkolesnik: Please inline them

pkg/subctl/cmd/show_connections.go
```diff
@@ -36,30 +36,35 @@ func getConnectionsStatus(config *rest.Config) []connectionStatus {
...
 	var status []connectionStatus
 
 	existingCfg, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, v1.GetOptions{})
 	if err != nil {
-		exitOnError("error reading from submariner client", err)
+		if SubmarinerIsInstalled(err) {
+			exitOnError("Error obtaining the Submariner resource", err)
+		}
```
mkolesnik, line 46:
```suggestion
		if ! SubmarinerIsInstalled(err) {
			return status
		}
		
		exitOnError("Error obtaining the Submariner resource", err)
```

pkg/subctl/cmd/show.go
```diff
@@ -98,3 +104,13 @@ func getClientConfigAndClusterName(rules *clientcmd.ClientConfigLoadingRules, ov
 
 	return restConfig{config: clientConfig, clusterName: *clusterName}, nil
 }
+
+func SubmarinerIsInstalled(err error) bool {
```
mkolesnik, line 108:
```suggestion
func submarinerMissing(err error) bool {
```

As suggested, this name makes more sense as you're checking the error type that you're sending to the function.
If you'd have an independent function checking somehow that submariner is or isn't installed (not receiving an `error` variable) then the name you use would make sense.

mkolesnik (changes requested):
Please revise to make the code more maintainable

pkg/subctl/cmd/show_connections.go
tpantelis: I think it would be cleaner to verify if submariner is install at a higher level and not even call `getConnectionsStatus` and `printConnections`. Similar for endpoints et al. That would eliminate the need for `SubmarinerInstalled` and `show all` can bail early instead of printing "_Submariner is not installed_" multiple times. I suggest to create a function to retrieve the `*v1alpha1.Submariner` and pass it into `getConnectionsStatus`, `getEndpointsStatus` et al. 

In _show.go_:

```
func GetSubmarinerResource(config *rest.Config) *v1alpha1.Submariner {
	submarinerClient, err := submarinerclientset.NewForConfig(config)
	exitOnError("Unable to get the Submariner client", err)

	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, v1.GetOptions{})
	if err != nil {
		if apierrors.IsNotFound(err) {
			fmt.Println("Submariner is not installed")
			return nil
		}

		exitOnError("Error obtaining the Submariner resource", err)
	}

	return submariner
}

```
In _show_connections.go_:

```
func getConnectionsStatus(submariner *v1alpha1.Submariner) []connectionStatus {
	gateways := submariner.Status.Gateways
	...
}

func showConnections(cmd *cobra.Command, args []string) {
	...
	for _, item := range configs {
		fmt.Println()
		fmt.Printf("Showing information for cluster %q:\n", item.clusterName)

		showConnectionsFor(GetSubmarinerResource(item.config))
	}
}

func showConnectionsFor(submariner *v1alpha1.Submariner) {
	if submariner != nil {
		status := getConnectionsStatus(submariner)
		printConnections(status)
	}
}

```
In _show_all.go_:

```
func showAll(cmd *cobra.Command, args []string) {
	...
	for _, item := range configs {
		fmt.Println()
		fmt.Printf("Showing information for cluster %q:\n", item.clusterName)

                submariner := GetSubmarinerResource(item.config)
                if submariner == nil {
                    continue
                }

                 ...
		fmt.Println("\nShowing Connections details")
		showConnectionsFor(submariner)
		...
	}
}
```

tpantelis (changes requested):

pkg/subctl/cmd/show.go
Jaanki: Done

pkg/subctl/cmd/show.go
Jaanki: Done

pkg/subctl/cmd/show.go
```diff
@@ -98,3 +104,13 @@ func getClientConfigAndClusterName(rules *clientcmd.ClientConfigLoadingRules, ov
...
+	if (err.Error() == submarinerNotInstalledError.Error()) || (err.Error() == submarinerUninstalledError.Error()) {
```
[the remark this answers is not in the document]
Jaanki: Done

tpantelis (approved):

maayanf24 (approved):

mkolesnik (changes requested):
Thanks, it looks much better, just a couple more suggestions

pkg/subctl/cmd/show_connections.go
```diff
@@ -72,14 +63,19 @@ func showConnections(cmd *cobra.Command, args []string) {
...
-		printConnections(status)
+		showConnectionsFor(GetSubmarinerResource(item.config))
 	}
 }
 
 func showConnectionsFromConfig(config *rest.Config) {
-	status := getConnectionsStatus(config)
-	printConnections(status)
+	showConnectionsFor(GetSubmarinerResource(config))
+}
```
tpantelis, line 72: I don't think we need this function anymore. Same with similar functions for endpoints et al.

pkg/subctl/cmd/show_connections.go
```diff
@@ -72,14 +63,19 @@ func showConnections(cmd *cobra.Command, args []string) {
...
+
+func showConnectionsFor(submariner *v1alpha1.Submariner) {
+	if submariner != nil {
```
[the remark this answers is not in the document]
tpantelis: It can be nil when called from L66 above.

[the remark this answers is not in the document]
mkolesnik: Sure, in that case I'd expect an error so that a bug can be opened, instead of a silent failure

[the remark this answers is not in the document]
tpantelis: No - nil means submariner isn't installed.

pkg/subctl/cmd/show_all.go
```diff
@@ -26,14 +26,23 @@ func showAll(cmd *cobra.Command, args []string) {
...
+		fmt.Println("")
+
+		submariner := GetSubmarinerResource(item.config)
+
+		if submariner == nil {
+			continue
+		}
+
+		fmt.Println("\nShowing Endpoint details")
 		showEndpointsFromConfig(item.config)
```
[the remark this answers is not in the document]
Jaanki: Agree. Changed it.

pkg/subctl/cmd/show_all.go
```diff
@@ -26,14 +26,23 @@ func showAll(cmd *cobra.Command, args []string) {
...
-		fmt.Println()
+		fmt.Println("\nShowing Connection details")
 		showConnectionsFromConfig(item.config)
-		fmt.Println()
+		fmt.Println("\nShowing Gateway details")
 		showGatewaysFromConfig(item.config)
```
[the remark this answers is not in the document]
Jaanki: showGatewaysFor(submariner) will again check for submariner != nil (which will always be true) because L44 will be executed only if submariner != nil.

1 way to do it is `printGateways(getGatewaysStatus(submariner))` which I dont like much. Another way is to keep the current implementation as is.

pkg/subctl/cmd/show_all.go
```diff
@@ -26,14 +26,23 @@ func showAll(cmd *cobra.Command, args []string) {
...
+
+		if submariner == nil {
+			continue
+		}
+
+		fmt.Println("\nShowing Endpoint details")
 		showEndpointsFromConfig(item.config)
-		fmt.Println()
+		fmt.Println("\nShowing Connection details")
 		showConnectionsFromConfig(item.config)
```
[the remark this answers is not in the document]
Jaanki: Dtitto

pkg/subctl/cmd/show_connections.go
Jaanki: This is called in show_all.go. See comment https://github.com/submariner-io/submariner-operator/pull/1017#discussion_r553229721

pkg/subctl/cmd/show_all.go
```diff
@@ -26,14 +26,23 @@ func showAll(cmd *cobra.Command, args []string) {
...
-		fmt.Println()
+		fmt.Println("\nShowing Gateway details")
 		showGatewaysFromConfig(item.config)
-		fmt.Println()
+		fmt.Println("\nShowing version details")
 		showVersionsFromConfig(item.config)
```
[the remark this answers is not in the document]
Jaanki: See comment https://github.com/submariner-io/submariner-operator/pull/1017#discussion_r553927639

pkg/subctl/cmd/show_all.go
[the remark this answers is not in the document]
tpantelis: It's fine if `showGatewaysFor` checks it again as `submariner` may be nil from the other call site. The other alternative is to check nil at the other call site - either way, doesn't matter.

pkg/subctl/cmd/show_connections.go
tpantelis: Right but that should call `showConnectionsFor` instead as Mike pointed out which renders `showConnectionsFromConfig` unused/unnecessary.

pkg/subctl/cmd/show_all.go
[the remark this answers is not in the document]
mkolesnik: I'm not sure whats the problem you expect exactly?

pkg/subctl/cmd/show_connections.go
mkolesnik: All these `show*FromConfig` functions are only called once from `showAll` and not being used elsewhere.
As they're just one line functions and don't add any value, I made suggestions to replace them with proper calls.

pkg/subctl/cmd/show_connections.go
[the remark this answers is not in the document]
mkolesnik: `showConnectionsFor` (or any other such function) is a private function and is not expected to be called with a `nil` value.
The value checking should be done outside or it can hide bugs as I already explained.

mkolesnik (changes requested):

pkg/subctl/cmd/show_connections.go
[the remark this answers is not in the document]
tpantelis: Well there are no private functions in Go - it's package scoped. The API contract for `showConnectionsFor` is what ever one wants it to be. It doesn't really matter who checks for nil in this case - if you feel strongly about it outside then that's fine.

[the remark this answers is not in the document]
mangelajo: I slightly agree with @mkolesnik here, if we are already checking for nil... it's not really necessary to check twice here... we don't expect this to be called with nil ever.

Anyway it will also work with nil here also, just being silent.

pkg/subctl/cmd/show_all.go
```diff
@@ -41,14 +41,23 @@ func showAll(cmd *cobra.Command, args []string) {
...
-		fmt.Println()
-		showGatewaysFromConfig(item.config)
-		fmt.Println()
-		showVersionsFromConfig(item.config)
+		fmt.Println("")
+
+		submariner := getSubmarinerResource(item.config)
+
+		if submariner == nil {
+			continue
```
mangelajo, line 51: Can we print a line here before continuing explaining that this specific item.clusterName does not have submariner installed?

pkg/subctl/cmd/show_endpoints.go
```diff
@@ -82,14 +72,19 @@ func showEndpoints(cmd *cobra.Command, args []string) {
 	for _, item := range configs {
 		fmt.Println()
 		fmt.Printf("Showing information for cluster %q:\n", item.clusterName)
-		status := getEndpointsStatus(item.config)
+		showEndpointsFor(GetSubmarinerResource(item.config))
+	}
+}
+
+func showEndpointsFor(submariner *v1alpha1.Submariner) {
+	if submariner != nil {
```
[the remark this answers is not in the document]
mangelajo: +1

pkg/subctl/cmd/show_gateways.go
```diff
@@ -82,14 +74,19 @@ func showGateways(cmd *cobra.Command, args []string) {
 	for _, item := range configs {
 		fmt.Println()
 		fmt.Printf("Showing information for cluster %q:\n", item.clusterName)
-		status := getGatewaysStatus(item.config)
+		showGatewaysFor(GetSubmarinerResource(item.config))
+	}
+}
+
+func showGatewaysFor(submariner *v1alpha1.Submariner) {
+	if submariner != nil {
```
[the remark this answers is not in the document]
mangelajo: +1

pkg/subctl/cmd/show_versions.go
```diff
@@ -99,20 +95,26 @@ func getVersions(config *rest.Config) []versionImageInfo {
 	return versions
 }
 
+func getVersionsFor(config *rest.Config) {
+	submariner := GetSubmarinerResource(config)
+	if submariner != nil {
```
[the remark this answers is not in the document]
mangelajo: +1

mangelajo (changes requested):
some comments. Janki thanks for improving this!!! :)

pkg/subctl/cmd/show_all.go
Jaanki: getSubmarinerResource prints "Submariner is not installed" line.

pkg/subctl/cmd/show.go
```diff
@@ -98,3 +104,20 @@ func getClientConfigAndClusterName(rules *clientcmd.ClientConfigLoadingRules, ov
...
 }
+
+func getSubmarinerResource(config *rest.Config) *v1alpha1.Submariner {
+	submarinerClient, err := submarinerclientset.NewForConfig(config)
+	exitOnError("Unable to get the Submariner client", err)
+
+	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, v1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			fmt.Println("Submariner is not installed")
```
mangelajo, line 115: Can you move this out of the function after the call to getSubmarinerResource ?
Thanks

pkg/subctl/cmd/show_all.go
mangelajo: let's move that line here instead before continue, will make it more evident handling of the result/prints.

pkg/subctl/cmd/show_connections.go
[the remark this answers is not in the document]
mangelajo: Oh but wait, we can call those independently (not with the show all..)

I think it's good as it is then, avoids a lot of duplication around.

mangelajo (approved):

mangelajo (changes requested):
The history is very weird now, can you rebase on master and squash it?

pkg/subctl/cmd/show_connections.go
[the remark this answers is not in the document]
mkolesnik: > Oh but wait, we can call those independently (not with the show all..)
> 
> I think it's good as it is then, avoids a lot of duplication around.

On the contrary, this is exactly duplication of this exact line all over the place.
If not nil check is really needed (which it isnt), it can happen on line 80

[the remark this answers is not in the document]
mangelajo: Right, ok, finally I settle on... handling that nil (and printing) outside.

For show_all.go it would only happen once ...

then for show_xxx.go it needs to happen on each of them before attempting to pass down the submariner object. Then... the nil checks on each function can be removed.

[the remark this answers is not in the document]
tpantelis: "_...(and printing) outside._". I assume you mean printing of "_Submariner is not installed_"?  So you want to move that out of `getSubmarinerResource` and duplicate it 4-5 times?

mangelajo:
mkolesni posted you an example here: https://github.com/mkolesnik/submariner-operator/commit/ba2a8e85d187bf2ac824a4f3a2c034e138cd10fd

tpantelis:
> mkolesni posted you an example here: [mkolesnik@ba2a8e8](https://github.com/mkolesnik/submariner-operator/commit/ba2a8e85d187bf2ac824a4f3a2c034e138cd10fd)

So `getSubmarinerResource` also has paths that print an error and exit.  So if we move the printing of the "not installed" message to the caller, shouldn't we be consistent and move the error handling as well?

Jaanki:
> > mkolesni posted you an example here: [mkolesnik@ba2a8e8](https://github.com/mkolesnik/submariner-operator/commit/ba2a8e85d187bf2ac824a4f3a2c034e138cd10fd)
> 
> So `getSubmarinerResource` also has paths that print an error and exit. So if we move the printing of the "not installed" message to the caller, shouldn't we be consistent and move the error handling as well?

@mangelajo  @mkolesnik WDYT?

Jaanki:
Closing in favor of https://github.com/submariner-io/submariner-operator/pull/1030

pkg/subctl/cmd/show_connections.go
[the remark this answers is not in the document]
mangelajo: if status == nil { 
    print submariner not installed
} else {
    printConnections(status)
}

PR #1031, opened 2021-01-19
Change `devel` tag to `subctl-devel`

mkolesnik:
Since we're changing the `master` branch name to `devel`, this tag will
collide, so change it something sensible.


mangelajo (approved):
Looks good, but we will need to make changes in get.submariner.io I guess.

skitt:
Why `subctl`? Is the tag supposed to be `subctl`-specific?

mkolesnik:
> Why `subctl`? Is the tag supposed to be `subctl`-specific?

It's subctl bleeding edge specific, if you have other suggestions I don't mind.

skitt (changes requested):
Thanks, just a minor fix: the commit message needs to be updated too ;-).

skitt (approved):

PR #1054, opened 2021-01-29
AKS: Use coredns-custom configmap for lighthouse

### LINKED ISSUE #1046: CoreDNS in AKS uses two configmaps.  the lighthouse forwarding domain needs added to coredns-custom
**Platform:** AKS (Azure Kubernetes Service) 

**Current Behavior:** Upon joining the clusterset, the configmap named "coredns" is modified. This file is soft-locked (allows the write, but immediately overwrites the file with infrastructure-layer-only data).

**Desired behavior:** when a configmap by the name of "coredns-custom" exists,  it should be the one written to. It also requires that the map be named `{anyWordHere}.server`  (lighthouse.server would serve nicely IMHO) instead of "coreFile".  See official documentation for formatting: https://docs.microsoft.com/en-us/azure/aks/coredns-custom#custom-forward-server

Additional notes: 
a little branch logic to UpdateDNSConfigMap should fix this up.

https://github.com/submariner-io/submariner-operator/blob/31f861cfc341f981998077d7ed88a67ca920c411/controllers/servicediscovery/servicediscovery_controller.go#L354

https://github.com/submariner-io/submariner-operator/blob/31f861cfc341f981998077d7ed88a67ca920c411/controllers/servicediscovery/servicediscovery_controller.go#L366

https://github.com/submariner-io/submariner-operator/blob/31f861cfc341f981998077d7ed88a67ca920c411/controllers/servicediscovery/servicediscovery_controller.go#L396

Related: 
   AKS fix slated for #0.8.1: https://github.com/submariner-io/submariner/issues/1106  (resolving these two issues equals AKS support, or at least a platform we can start testing on.)

vthapar:
Lighthouse makes changes to `coredns` configmap to forward
`clusterset.local` queries to lighthouse coredns server. But
AKS doesn't allow changes to this configmap by overwriting any
changes.

AKS provides `custom-coredns` for such changes, so lighthouse
should use that.

Fixes #1046

### DIFF
--- a/controllers/servicediscovery/servicediscovery_controller.go
+++ b/controllers/servicediscovery/servicediscovery_controller.go
@@ -62,6 +62,7 @@ const (
 	lighthouseForwardPluginName   = "lighthouse"
 	coreDNSNamespace              = "kube-system"
 	coreDNSName                   = "coredns"
+	customCoreDNSName             = "coredns-custom"
 )
 
 // NewReconciler returns a new ServiceDiscoveryReconciler
@@ -155,7 +156,14 @@ func (r *ServiceDiscoveryReconciler) Reconcile(request reconcile.Request) (recon
 			return reconcile.Result{}, err
 		}
 	}
-	err = updateDNSConfigMap(r.client, r.k8sClientSet, instance, reqLogger)
+	err = updateDNSCustomConfigMap(r.client, r.k8sClientSet, instance, reqLogger)
+	if errors.IsNotFound(err) {
+		err = updateDNSConfigMap(r.client, r.k8sClientSet, instance, reqLogger)
+	} else if err != nil {
+		reqLogger.Error(err, "Error updating the 'custom-coredns' ConfigMap")
+		return reconcile.Result{}, err
+	}
+
 	if errors.IsNotFound(err) {
 		// Try to update Openshift-DNS
 		return reconcile.Result{}, updateOpenshiftClusterDNSOperator(instance, r.client, r.operatorClientSet, reqLogger)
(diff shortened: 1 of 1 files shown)


controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -351,6 +359,39 @@ func newLighthouseCoreDNSService(cr *submarinerv1alpha1.ServiceDiscovery) *corev
 	}
 }
 
+func updateDNSCustomConfigMap(client controllerClient.Client, k8sclientSet clientset.Interface, cr *submarinerv1alpha1.ServiceDiscovery,
```
skitt, line 362: Couldn’t this share a lot of `updateDNSConfigMap`?

skitt (changes requested):
It feels rather brittle to just try `coredns-custom` and assume that it’s the right thing to update if it’s present... Is there anything better we could do? At some point we’ll probably end up with “infrastructure drivers” similar to cable drivers, but without going there yet, perhaps we could tie this more specifically to AKS.

controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -351,6 +359,39 @@ func newLighthouseCoreDNSService(cr *submarinerv1alpha1.ServiceDiscovery) *corev
...
+		if _, ok := configMap.Data["lighthouse.server"]; ok {
+			reqLogger.Info("Overwriting existing lighthouse.server data in coredns-custom")
+		}
+
+		coreFile := ""
+		for _, domain := range append([]string{"clusterset.local"}, cr.Spec.CustomDomains...) {
+			coreFile = fmt.Sprintf("%s%s:53 {\n    forward . %s\n}\n",
+				coreFile, domain, lighthouseClusterIp)
+		}
+		log.Info("Updated coredns-custom ConfigMap for lighthouse.server" + coreFile)
```
sridhargaddam, line 386:
```suggestion
		log.Info("Updating coredns-custom ConfigMap for lighthouse.server" + coreFile)
```

sridhargaddam (approved):

vthapar:
> It feels rather brittle to just try `coredns-custom` and assume that it’s the right thing to update if it’s present... Is there anything better we could do? At some point we’ll probably end up with “infrastructure drivers” similar to cable drivers, but without going there yet, perhaps we could tie this more specifically to AKS.

Yes, that is the plan. This is a temporary workaround till we have better way to detect AKS and use that.

controllers/servicediscovery/servicediscovery_controller.go
vthapar: Except for section that adds forward to lighthouse, it is different. Note how with way AKS sets it up, we modify just our part of configmap and don't have to worry about retaining rest of existing configmap.

aswinsuryan (approved):

controllers/servicediscovery/servicediscovery_controller.go
skitt: OK, let’s get this in, I’ll follow up with a PR to show what I mean.

skitt (approved):

PR #1055, opened 2021-01-29
Allow service to be deployed without connectivity

mangelajo:
This cover the use case of routed networks where pods and
services can already talk to remote clusters by other means
and the submariner routing and gateways are not required,
but service discovery is necessary.

Continues in: https://github.com/submariner-io/submariner-operator/pull/1087  , sorry for the mess, I deleted the original branch by error.

### DIFF
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+	"github.com/submariner-io/admiral/pkg/stringset"
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/submariner-io/submariner-operator/pkg/discovery/globalnet"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/components"
 
 	"github.com/submariner-io/submariner-operator/pkg/broker"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
@@ -34,10 +36,13 @@ var (
 	globalnetEnable             bool
 	globalnetCidrRange          string
 	defaultGlobalnetClusterSize uint
-	serviceDiscovery            bool
+	serviceDiscoveryEnabled     bool
+	componentArr                []string
 	GlobalCIDRConfigMap         *v1.ConfigMap
 	defaultCustomDomains        []string
 )
+
+var validComponents = []string{components.ServiceDiscovery, components.Connectivity}
 
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
@@ -50,11 +55,16 @@ func init() {
 	deployBroker.PersistentFlags().StringVar(&ipsecSubmFile, "ipsec-psk-from", "",
 		"import IPsec PSK from existing submariner broker file, like broker-info.subm")
 
-	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", true,
+	deployBroker.PersistentFlags().BoolVar(&serviceDiscoveryEnabled, "service-discovery", true,
 		"enable multi-cluster service discovery")
+
+	_ = deployBroker.PersistentFlags().MarkDeprecated("service-discovery", "please use --components instead")
 
 	deployBroker.PersistentFlags().StringSliceVar(&defaultCustomDomains, "custom-domains", nil,
 		"list of domains to use for multicluster service discovery")
+
+	deployBroker.PersistentFlags().StringSliceVar(&componentArr, "components", validComponents,
+		fmt.Sprintf("The components to be installed - any of #v. The default is all components", validComponents))
 
 	addKubeconfigFlag(deployBroker)
 	rootCmd.AddCommand(deployBroker)
@@ -67,10 +77,21 @@ var deployBroker = &cobra.Command{
 	Short: "Set the broker up",
 	Run: func(cmd *cobra.Command, args []string) {
 
+		componentSet := stringset.New(componentArr...)
+
+		// TODO: Remove this in the future, while service-discovery is marked as
+		//       deprecated we should still provide a consistent broker config file
+		if !serviceDiscoveryEnabled {
+			componentSet.Remove(components.ServiceDiscovery)
+		}
+
+		if err := isValidComponents(componentSet); err != nil {
+			exitOnError("Invalid components parameter", err)
+		}
+
 		if valid, err := isValidGlobalnetConfig(); !valid {
 			exitOnError("Invalid GlobalCidr configuration", err)
 		}
-
 		config, err := getRestConfig(kubeConfig, kubeContext)
 		exitOnError("The provided kubeconfig is invalid", err)
 
@@ -102,7 +123,8 @@ var deployBroker = &cobra.Command{
 			status.QueueSuccessMessage(fmt.Sprintf("Backed up previous %s to %s", brokerDetailsFilename, newFilename))
 		}
 
-		subctlData.ServiceDiscovery = serviceDiscovery
+		subctlData.ServiceDiscovery = serviceDiscoveryEnabled
+		subctlData.SetComponents(componentSet)
 
 		if len(defaultCustomDomains) > 0 {
 			subctlData.CustomDomains = &defaultCustomDomains
@@ -119,6 +141,22 @@ var deployBroker = &cobra.Command{
 		exitOnError("Error writing the broker information", err)
 
 	},
+}
+
+func isValidComponents(componentSet stringset.Interface) error {
+	validComponentSet := stringset.New(validComponents...)
+
+	if componentSet.Size() < 1 {
+		return fmt.Errorf("at least one component must be provided for deployment")
+	}
+
+	for _, component := range componentSet.Elements() {
+		if !validComponentSet.Contains(component) {
+			return fmt.Errorf("unknown component: %s", component)
+		}
+	}
+
+	return nil
 }
 
 func isValidGlobalnetConfig() (bool, error) {
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -43,6 +43,7 @@ import (
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	"github.com/submariner-io/submariner-operator/pkg/names"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/datafile"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/servicediscoverycr"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/submarinercr"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/submarinerop"
 	"github.com/submariner-io/submariner-operator/pkg/versions"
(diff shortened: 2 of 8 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -277,16 +278,31 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, contextName string, su
...
+			status.QueueSuccessMessage("Submariner is up and running")
+			status.End(cli.Success)
+		} else {
+			status.QueueFailureMessage("Submariner deployment failed")
+			status.End(cli.Failure)
+		}
+
+		exitOnError("Error deploying Submariner", err)
+	} else if subctlData.IsServiceDiscoveryEnabled() {
+		status.Start("Deploying Lighthouse only")
```
tpantelis, line 294:
```suggestion
		status.Start("Deploying service discovery only")
```

pkg/subctl/cmd/join.go
```diff
@@ -277,16 +278,31 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, contextName string, su
...
+		if err == nil {
+			status.QueueSuccessMessage("Submariner is up and running")
+			status.End(cli.Success)
+		} else {
+			status.QueueFailureMessage("Submariner deployment failed")
+			status.End(cli.Failure)
+		}
+
+		exitOnError("Error deploying Submariner", err)
+	} else if subctlData.IsServiceDiscoveryEnabled() {
```
tpantelis, line 293: How about adding a `ConnectivityEnabled` flag to the `Submariner` CR so we can let the operator controller handle this normally?

pkg/subctl/operator/servicediscoverycr/ensure.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func createServiceDiscovery(clientSet submarinerclientset.Interface, namespace string, serviceDiscoveryCR *submariner.ServiceDiscovery) (
+	bool, error) {
+	_, err := clientSet.SubmarinerV1alpha1().ServiceDiscoveries(namespace).Create(serviceDiscoveryCR)
+	if err == nil {
+		return true, nil
+	} else if errors.IsAlreadyExists(err) {
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			// We can’t always handle existing resources, and we want to overwrite them anyway, so delete them
```
tpantelis, line 64: `RetryOnConflict` is not applicable for Create. Also the Delete and Create operations aren't atomic so could be problematic. What exactly is the issue using the common `CreateOrUpdate` pattern? An Update does overwrite the resource.

tpantelis (changes requested):

pkg/subctl/cmd/deploybroker.go
```diff
@@ -34,11 +36,14 @@ var (
 	globalnetEnable             bool
 	globalnetCidrRange          string
 	defaultGlobalnetClusterSize uint
-	serviceDiscovery            bool
+	serviceDiscoveryEnabled     bool
+	componentArr                []string
```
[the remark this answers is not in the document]
mangelajo: It overlapps with the "components" import, otherwise I had to give a weird name to that import.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -50,12 +55,17 @@ func init() {
 	deployBroker.PersistentFlags().StringVar(&ipsecSubmFile, "ipsec-psk-from", "",
 		"import IPsec PSK from existing submariner broker file, like broker-info.subm")
 
-	deployBroker.PersistentFlags().BoolVar(&serviceDiscovery, "service-discovery", true,
+	deployBroker.PersistentFlags().BoolVar(&serviceDiscoveryEnabled, "service-discovery", true,
 		"enable multi-cluster service discovery")
 
+	deployBroker.PersistentFlags().MarkDeprecated("service-discovery", "please use --componentArr instead")
```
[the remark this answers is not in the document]
mangelajo: Done, a "refactor" went bad...

pkg/subctl/cmd/deploybroker.go
```diff
@@ -50,12 +55,17 @@ func init() {
...
+
 	deployBroker.PersistentFlags().StringSliceVar(&defaultCustomDomains, "custom-domains", nil,
 		"list of domains to use for multicluster service discovery")
 
+	deployBroker.PersistentFlags().StringSliceVar(&componentArr, "componentArr", validComponents,
+		"componentArr to be installed")
```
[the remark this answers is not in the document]
mangelajo: Done

pkg/subctl/cmd/deploybroker.go
```diff
@@ -67,10 +77,21 @@ var deployBroker = &cobra.Command{
...
+		componentSet := stringset.New(componentArr...)
+
+		// TODO: Remove this in the future, while service-discovery is marked as
+		//       deprecated we should still provide a consistent broker config file
+		if !serviceDiscoveryEnabled {
+			componentSet.Remove(components.ServiceDiscovery)
+		}
+
+		if err := isValidComponents(componentSet); err != nil {
+			exitOnError("Invalid componentArr parameter", err)
```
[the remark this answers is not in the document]
mangelajo: Done

pkg/subctl/operator/servicediscoverycr/ensure.go
```diff
@@ -0,0 +1,78 @@
+/*
+© 2019 Red Hat, Inc. and others.
```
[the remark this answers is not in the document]
mangelajo: oops :D

pkg/subctl/cmd/join.go
```diff
@@ -277,16 +278,31 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, contextName string, su
...
+		status.Start("Deploying Lighthouse only")
+		err = servicediscoverycr.Ensure(clientConfig, OperatorNamespace, populateServiceDiscoverySpec(subctlData))
+		if err == nil {
+			status.QueueSuccessMessage("Lighthouse is up an running")
+			status.End(cli.Success)
+		} else {
+			status.QueueFailureMessage("Lighthouse deployment failed")
+			status.End(cli.Failure)
+		}
+		exitOnError("Error deploying Lighthouse", err)
```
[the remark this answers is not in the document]
mangelajo: Done

pkg/subctl/cmd/join.go
```diff
@@ -277,16 +278,31 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, contextName string, su
...
+
+		exitOnError("Error deploying Submariner", err)
+	} else if subctlData.IsServiceDiscoveryEnabled() {
+		status.Start("Deploying Lighthouse only")
+		err = servicediscoverycr.Ensure(clientConfig, OperatorNamespace, populateServiceDiscoverySpec(subctlData))
+		if err == nil {
+			status.QueueSuccessMessage("Lighthouse is up an running")
+			status.End(cli.Success)
+		} else {
+			status.QueueFailureMessage("Lighthouse deployment failed")
```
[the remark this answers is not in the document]
mangelajo: Done

pkg/subctl/cmd/join.go
```diff
@@ -277,16 +278,31 @@ func joinSubmarinerCluster(config clientcmd.ClientConfig, contextName string, su
...
+			status.QueueFailureMessage("Submariner deployment failed")
+			status.End(cli.Failure)
+		}
+
+		exitOnError("Error deploying Submariner", err)
+	} else if subctlData.IsServiceDiscoveryEnabled() {
+		status.Start("Deploying Lighthouse only")
+		err = servicediscoverycr.Ensure(clientConfig, OperatorNamespace, populateServiceDiscoverySpec(subctlData))
+		if err == nil {
+			status.QueueSuccessMessage("Lighthouse is up an running")
```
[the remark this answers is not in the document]
mangelajo: Done

pkg/subctl/cmd/join.go
mangelajo: Done

pkg/subctl/cmd/join.go
mangelajo: It's the operator controller itself :)  we designed it that way. This is why we split it in two objects "Submariner "  & "ServiceDiscovery"

But in most cases submariner will bring service discovery, so we generate that object automatically. But we wanted to allow ServiceDiscovery being created alone, but never added support to subctl.

pkg/subctl/operator/servicediscoverycr/ensure.go
mangelajo: I'm I just copied what we had for submarinercr.

pkg/subctl/cmd/join.go
tpantelis: The `SubmarinerReconciler` in the operator controller already creates the `ServiceDiscovery` CR based on the `ServiceDiscoveryEnabled` flag in the `Submariner` CR.  It seems to make sense to keep that workflow.

pkg/subctl/operator/servicediscoverycr/ensure.go
tpantelis: OK but I think `CreateOrUpdate` is the better way to do it.

pkg/subctl/cmd/deploybroker.go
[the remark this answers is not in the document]
tpantelis: I think the import should be `component`.

pkg/subctl/operator/servicediscoverycr/ensure.go
mangelajo: I'll move those two to the CreateOrUpdate pattern (sorry I was on a run when I wrote the previous comment)

pkg/subctl/cmd/join.go
mangelajo: But then it doesn't make sense to keep a separate API on the submariner-operator controller. It could just be a sub-package handling those details.

It's the current design, not written in stone anyway, but if we do what you suggest I would kill the ServiceDiscovery CR altogether. @vthapar @skitt   ^

I lean more towards keeping ServiceDiscover TBH.

pkg/subctl/cmd/deploybroker.go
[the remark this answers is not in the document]
mangelajo: > I think the import should be `component`.

Makes sense.

pkg/subctl/cmd/join.go
tpantelis: I think we should do it one way or the other.  I lean towards removing the `ServiceDiscoveryEnabled` flag altogether and having `subctl` create the `ServiceDiscovery` CR as you've done here. We can do the flag removal part in another PR.

mangelajo: Right, "ServiceDiscoveryEnabled" flag exists so the people using the bare operator doesn't need to create 2 objects. I think it would be fine if we move all those pesky credentials to a kubernetes secret instead, so the broker credentials live in a single place.

@vthapar thoughts ^

mangelajo:
Sorry @tpantelis it seems like I destroyed the original branch: 


mangelajo wants to merge 1 commit into submariner-io:devel from unknown **repository**

I'll re-create this PR applying the comments first :/

mangelajo:
Continues in: #1087 , sorry for the mess, I deleted the original branch by error.

PR #1059, opened 2021-02-02
Do not merge, multiple CRD version support

mangelajo:


mangelajo:
This is just uploaded for reference about how to check API versions. Closing.

PR #1061, opened 2021-02-03
Fix network plugin detection 

mkimuram:
This PR fixes network plugin detection by:
- Making detection logic of each network plugin or cidrs not to fail on internal error to properly fall through,
- Adding logic to detect pod cidrs and cluster cidrs for k8s w/o k8s pods(api-server, controller, kube-proxy)

Closes: https://github.com/submariner-io/submariner/issues/1116


mkimuram:
I've tested by below way and it seems to work well:

<details>

[Build environment]
```
# make build
# make images
# docker save quay.io/submariner/submariner-operator:dev > /tmp/submariner-operator.tar
```
(Copy bin/subctl and /tmp/submariner-operator.tar to test environment.)

[Test environment (after k3s is deployed)]
```
# k3s ctr images import submariner-operator.tar
# k3s ctr images tag quay.io/submariner/submariner-operator:dev quay.io/submariner/submariner-operator-dev:dev
# k3s ctr images list | grep submariner

# ./subctl deploy-broker --globalnet --globalnet-cidr-range 169.254.0.0/16

# ./subctl join broker-info.subm --clusterid cl1 --clustercidr 10.44.0.0/16 --servicecidr 10.45.0.0/16 --natt=false --globalnet-cidr 169.254.44.0/24 --image-override submariner-operator=quay.io/submariner/submariner-operator-dev:dev

# kubectl edit deploy -n submariner-operator submariner-operator
```
(Change `imagePullPolicy` to `Never`.)
```
# watch kubectl get pod -n submariner-operator

# ./subctl show networks

Showing network details for cluster "default":
    Discovered network details:
        Network plugin:  generic
        Service CIDRs:   [10.45.0.0/16]
        Cluster CIDRs:   [10.44.0.0/24]

# iptables -t nat -S POSTROUTING | head -n 2
-P POSTROUTING ACCEPT
-A POSTROUTING -j SUBMARINER-POSTROUTING

# kubectl logs -n submariner-operator -l app=submariner-routeagent
I0202 22:58:31.139620       1 routes_iface.go:121] Found nil gw or dst
I0202 22:58:31.139631       1 gw_transition.go:40] Creating the vxlan interface: vx-submariner on the gateway node
I0202 22:58:31.140418       1 vxlan.go:164] Successfully added the bridge fdb entry 192.168.122.67 00:00:00:00:00:00
I0202 22:58:31.140562       1 vxlan.go:251] Successfully configured rp_filter to loose mode(2) on vx-submariner
I0202 22:58:33.297912       1 handler.go:78] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
I0202 22:58:33.297953       1 node_handler.go:30] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
I0202 22:58:38.924489       1 handler.go:78] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
I0202 22:58:38.924530       1 node_handler.go:30] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
I0202 23:03:41.255994       1 handler.go:78] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
I0202 23:03:41.256036       1 node_handler.go:30] A Node with name "k3s" and addresses []v1.NodeAddress{v1.NodeAddress{Type:"InternalIP", Address:"192.168.122.67"}, v1.NodeAddress{Type:"Hostname", Address:"k3s"}} has been updated
```
</details>

pkg/discovery/network/generic.go
```diff
@@ -62,13 +50,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+		return "", fmt.Errorf("parsing (%s) failed. it doesn't match with %v", msg, re)
+	}
+
+	// returns first matching string
+	return match[1], nil
+}
+
+func findPodIPRange(clientSet kubernetes.Interface) (string, error) {
+	podIPRange, err := findPodIPRangeKubeController(clientSet)
+	if err == nil && podIPRange != "" {
```
tpantelis: I don't think we should silently dismiss the error. An error should mean that `findPodIPRangeKubeController` couldn't determine whether or not the  desired data exists, in which case we should fail and propagate the error. If it is definitively determined that the data doesn't exist, then it should return an empty string and nil error. However the underlying function `findPod` returns an error if no pod is found in the list which I don't think is right - it should return nil (which `findPodCommandParameter` handles properly).

pkg/discovery/network/generic.go
```diff
@@ -62,13 +50,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+		return podIPRange, nil
+	}
+
+	podIPRange, err = findPodIPRangeNodeSpec(clientSet)
+	if err == nil && podIPRange != "" {
+		return podIPRange, nil
+	}
+
+	return "", fmt.Errorf("podIPRange can't be found by any methods")
```
tpantelis:
```suggestion
	return "", fmt.Errorf("No Pod IP range could be determined")
```

pkg/discovery/network/generic.go
```diff
@@ -62,13 +50,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+}
+
+func parseToPodCidr(nodes []v1.Node) (string, error) {
+	for _, node := range nodes {
+		if node.Spec.PodCIDR != "" {
+			return node.Spec.PodCIDR, nil
+		}
+	}
+
+	return "", fmt.Errorf("no node with Spec.PodCIDR found")
```
tpantelis: I don't think we should return an error here if not found. We should let the caller handle not found as it sees fit (it so happens this is the last resort for the caller but that could change).

pkg/discovery/network/generic.go
```diff
@@ -62,13 +50,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+
+	return "", fmt.Errorf("clusterIPRange can't be found by any methods")
+}
+
+func findClusterIPRangeApiserver(clientSet kubernetes.Interface) (string, error) {
 	return findPodCommandParameter(clientSet, "component=kube-apiserver", "--service-cluster-ip-range")
 }
 
+func findClusterIPRangeSvcCreation(clientSet kubernetes.Interface) (string, error) {
+	// find service cidr based on https://stackoverflow.com/questions/44190607/how-do-you-find-the-cluster-service-cidr-of-a-kubernetes-cluster
```
tpantelis, line 80: I think this method is a bit fragile. It depends on a specific error message format and content returned from the API server that could change. In the link above, the most upvoted answer and the one that got accepted explained to parse the output from `kubectl cluster-info dump`. So I think we should  try to go in that direction, ie look at the `kubectl` code to see how they obtain the  cluster info.

pkg/discovery/network/generic.go
```diff
@@ -62,13 +50,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+	if err == nil && clusterIPRange != "" {
+		return clusterIPRange, nil
+	}
+
+	clusterIPRange, err = findClusterIPRangeSvcCreation(clientSet)
+	if err == nil && clusterIPRange != "" {
+		return clusterIPRange, nil
+	}
+
+	return "", fmt.Errorf("clusterIPRange can't be found by any methods")
```
tpantelis:
```suggestion
	return "", fmt.Errorf("No cluster IP range could be determined")
```

pkg/discovery/network/generic.go
```diff
@@ -25,32 +32,13 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
-
-	if podIPRange == "" {
-		podIPRange, err = findPodIPRangeKubeProxy(clientSet)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	if podIPRange != "" {
+	podIPRange, err := findPodIPRange(clientSet)
```
tpantelis, line 36: As commented elsewhere, I think we should propagate an error.

pkg/discovery/network/generic.go
```diff
@@ -25,32 +32,13 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
-	// our discovery method of looking for the kube-apiserver pod is useless, and we won't be
-	// able to return such detail
 	clusterIPRange, err := findClusterIPRange(clientSet)
-
-	if err != nil {
-		return nil, err
-	}
-
-	if clusterIPRange != "" {
+	if err == nil && clusterIPRange != "" {
```
tpantelis: ...same here

pkg/discovery/network/canal.go
```diff
@@ -52,14 +52,10 @@ func discoverCanalFlannelNetwork(clientSet kubernetes.Interface) (*ClusterNetwor
...
-	genNetwork, err := discoverGenericNetwork(clientSet)
-	if err != nil {
-		return nil, err
-	}
-
-	if genNetwork != nil {
-		clusterNetwork.ServiceCIDRs = genNetwork.ServiceCIDRs
+	// Try to detect the service CIDRs using the generic functions
+	clusterIPRange, err := findClusterIPRange(clientSet)
+	if err == nil && clusterIPRange != "" {
```
tpantelis: As commented elsewhere, I think we should propagate an error.

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		var clusterNet *ClusterNetwork
+
+		BeforeEach(func() {
+			clusterNet = testDiscoverGenericWith(
+				fakeNode("node1", ""),
+				fakeNode("node2", testPodCIDR),
+			)
+		})
+
+		It("Should return the ClusterNetwork structure with PodCIDR", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with the pod CIDR", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		It("Should return nil cluster network", func() {
+			clusterNet := testDiscoverGenericWith(
+				fakeNode("node1", ""),
+				fakeNode("node2", ""),
+			)
+			Expect(clusterNet).To(BeNil())
+		})
+	})
+
+	When("There is pod cidr information on any of nodes", func() {
```
tpantelis:
```suggestion
	When("pod CIDR information exists on a node", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
 			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
 		})
 	})
+
+	When("There is no pod cidr information on node", func() {
```
tpantelis:
```suggestion
	When("no pod CIDR information exists on any node", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		It("Should identify the networkplugin as generic", func() {
+			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
+		})
+
+		It("Should return the ClusterNetwork structure with empty service CIDR", func() {
+			Expect(clusterNet.ServiceCIDRs).To(BeEmpty())
+		})
+	})
+
+	When("There is pod cidr information on any of nodes and there is kubeapi pod", func() {
```
tpantelis:
```suggestion
	When("both pod and service CIDR information exists", func() {
```

pkg/discovery/network/network_suite_test.go
```diff
@@ -67,3 +67,16 @@ func fakeService(namespace, name, component string) *v1.Service {
...
+	spec := v1.NodeSpec{}
+	if podCIDR != "" {
+		spec.PodCIDR = podCIDR
+	}
+	return &v1.Node{
+		ObjectMeta: v1meta.ObjectMeta{
+			Name: name,
+		},
+		Spec: spec,
+	}
```
tpantelis:
```suggestion
	return &v1.Node{
		ObjectMeta: v1meta.ObjectMeta{
			Name: name,
		},
		Spec: v1.NodeSpec{
		    PodCIDR: podCIDR,
		},
	}
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		It("Should identify the networkplugin as generic", func() {
+			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
+		})
+	})
+
+	When("No resource and creating invalid svc doesn't return error", func() {
+		It("Should return nil cluster network", func() {
+			clusterNet := testDiscoverGenericEmulatedErrWith(nil)
+			Expect(clusterNet).To(BeNil())
+		})
```
tpantelis, line 268: This test case essentially the same as the one on L42 above.

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+	})
+
+	When("No resource and creating invalid svc returns unexpected error", func() {
+		It("Should return nil cluster network", func() {
```
tpantelis: This case should return the error as per my comments.

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,115 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
+		})
+
+		It("Should return the ClusterNetwork structure with empty service CIDR", func() {
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDR}))
+		})
+
+	})
+
+	When("Node with cidr information and creating invalid svc returns expected error", func() {
```
tpantelis: This is essentially the same as the case on L204 when both exist in some manner.

pkg/discovery/network/generic_test.go
```diff
@@ -165,3 +284,23 @@ func testDiscoverGenericWith(objects ...runtime.Object) *ClusterNetwork {
...
+	if expectedErr != nil {
+		clientSet.CoreV1().(*fakecorev1.FakeCoreV1).PrependReactor("create", "services",
+			func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+				return true, nil, expectedErr
+			})
+	} else {
+		clientSet.CoreV1().(*fakecorev1.FakeCoreV1).PrependReactor("create", "services",
+			func(action testing.Action) (handled bool, ret runtime.Object, err error) {
+				return true, &v1.Service{}, nil
+			})
```
tpantelis: Since this function is intended to simulate an error, it seems we should expect an error passed in. The `else` here doesn't see to make sense.

tpantelis (changes requested):

pkg/discovery/network/generic.go
mangelajo: At this point it's not possible out of cluster-info dump: that's what our other heuristics do already. But in some cases, the api server services don't run inside the cluster itself (for example they are provided directly by GKE, AKS... or in the case of rancher they run in a container outside of kubernetes, and not inside), so the details can't be found.

I believe that the extra heuristic that mkimura added is great, and will find the range in many cases. If this still fails, it's OK, then the admin will be asked for the range, and life goes on, but this will catch 90% of the remaining cases for AKS,GKE,rancher...

mangelajo (commented):
see inline comments, but the new service heuristic is good IMO, "cluster-dump" equivalent is what we do on the other heuristics.

pkg/discovery/network/generic.go
sridhargaddam: I tend to agree with @mangelajo as we have some implementations of K8s which run the API server outside of the cluster and if we really want to purse cluster-info dump, it can be done in a separate PR.

pkg/discovery/network/generic.go
mkimuram: Thank you for your comment.

I found the same discussion [here](https://github.com/submariner-io/submariner-operator/pull/581) and agree with it. I will make a change as suggested.

pkg/discovery/network/generic.go
mkimuram: Changed it to return nil, so that `findClusterIPRange` won't return error when no internal find functions return error.

pkg/discovery/network/generic.go
mkimuram: Changed it to return nil, so that `findPodIPRange` won't return error when no internal find functions return error.

pkg/discovery/network/generic_test.go
mkimuram: Deleted this duplicated test.

pkg/discovery/network/generic_test.go
mkimuram: Changed to check the error case.

pkg/discovery/network/generic_test.go
mkimuram: Deleted nil error case.

pkg/discovery/network/pods.go
```diff
@@ -55,8 +55,13 @@ func findPod(clientSet kubernetes.Interface, labelSelector string) (*v1.Pod, err
 		Limit:         1,
 	})
 
-	if err != nil || len(pods.Items) == 0 {
+	if err != nil {
 		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", labelSelector)
```
mkimuram, line 59: `WithMessagef` actually returns `nil` if `err = nil`, but it is confusing. So, let's separate `err != nil` case and the other case.

pkg/discovery/network/generic.go
mkimuram: Fixed as suggested.

pkg/discovery/network/generic.go
```diff
@@ -89,7 +97,7 @@ func findClusterIPRangeSvcCreation(clientSet kubernetes.Interface) (string, erro
 
 	// creating invalid svc didn't fail as expected
 	if err == nil {
-		return "", fmt.Errorf("creating invalid service(%v) didn't fail", invalidSvcSpec)
+		return "", nil
```
mkimuram: Keeping this return error makes unit tests complex. All existing tests for network plugin detection will need to rely on creating svc always return error.

pkg/discovery/network/generic.go
```diff
@@ -53,13 +55,64 @@ func findClusterIPRange(clientSet kubernetes.Interface) (string, error) {
...
+				{
+					Port: 443,
+					TargetPort: intstr.IntOrString{
+						IntVal: 443,
+					},
+				},
+			},
+		},
+	}
+	_, err := clientSet.CoreV1().Services("submariner-operator").Create(invalidSvcSpec)
```
mkimuram: Creating svc in `submariner-operator` namespace might fail if it is called before the namespace is created. It seems to happen even on `subctl join` command. Other candidate will be `kube-system`, `default`, and random namespace created just for this check. Any ideas?

mkimuram (commented):
@tpantelis 

I've fixed almost as suggested. Please see my inline comments.

pkg/discovery/network/canal_test.go
```diff
@@ -50,6 +52,18 @@ var _ = Describe("discoverCanalFlannelNetwork", func() {
 		})
 
 	})
+
+	When("No kubeapi pod but svc creation fails properly", func() {
```
tpantelis:
```suggestion
	When("No kube-api pod exists and service creation returns the expected error", func() {
```

pkg/discovery/network/generic.go
```diff
@@ -62,13 +58,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+
+	clusterIPRange, err = findClusterIPRangeSvcCreation(clientSet)
+	if err != nil || clusterIPRange != "" {
+		return clusterIPRange, err
+	}
+
+	return "", nil
+}
+
+func findClusterIPRangeApiserver(clientSet kubernetes.Interface) (string, error) {
```
tpantelis:
```suggestion
func findClusterIPRangeFromApiserver(clientSet kubernetes.Interface) (string, error) {
```

pkg/discovery/network/generic.go
```diff
@@ -62,13 +58,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
 	return findPodCommandParameter(clientSet, "component=kube-apiserver", "--service-cluster-ip-range")
 }
 
+func findClusterIPRangeSvcCreation(clientSet kubernetes.Interface) (string, error) {
```
tpantelis:
```suggestion
func findClusterIPRangeFromServiceCreation(clientSet kubernetes.Interface) (string, error) {
```

pkg/discovery/network/generic.go
tpantelis: I think we could use empty namespace but, if not, `default`.

pkg/discovery/network/generic.go
```diff
@@ -62,13 +58,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+
+	// creating invalid svc didn't fail as expected
+	if err == nil {
+		return "", nil
+	}
+
+	return parseToServiceCidr(err.Error())
+}
+
+func parseToServiceCidr(msg string) (string, error) {
```
tpantelis:
```suggestion
func parseServiceCIDRFrom(msg string) (string, error) {
```

pkg/discovery/network/generic.go
tpantelis: So this means the API server didn't behave as we expected, likely something changed. I think it would be useful to at least log a message:

`status.QueueWarningMessage("Could not determine the service IP range via service creation - the expected error was not returned")`

pkg/discovery/network/generic.go
```diff
@@ -62,13 +58,112 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+	//   "The Service \"invalid-svc\" is invalid: spec.clusterIPs: Invalid value: []string{\"1.1.1.1\"}:
+	//   failed to allocated ip:1.1.1.1 with error:provided IP is not in the valid range.
+	//   The range of valid IPs is 10.45.0.0/16"
+	// expected matched string is below:
+	//   10.45.0.0/16
+	re := regexp.MustCompile(".*valid IPs is (.*)$")
+
+	match := re.FindStringSubmatch(msg)
+	if match == nil {
+		return "", fmt.Errorf("parsing (%s) failed. it doesn't match with %v", msg, re)
```
tpantelis, line 127: We don't return an error above if no error was returned from service creation so we should be consistent here. I think that's the correct behavior as it means we didn't find the desired data. As above I think we should log a message:

`status.QueueWarningMessage(fmt.Sprintf"Could not determine the service IP range via service creation - the expected error was not returned. The actual error was %q", msg))`

pkg/discovery/network/generic_test.go
```diff
@@ -28,6 +32,12 @@ import (
 const testPodCIDR = "1.2.3.4/16"
 const testServiceCIDR = "4.5.6.7/16"
 
+var errExpected = fmt.Errorf("The Service \"invalid-svc\" is invalid: "+
+	"spec.clusterIPs: Invalid value: []string{\"1.1.1.1\"}: failed to "+
+	"allocated ip:1.1.1.1 with error:provided IP is not in the valid range. "+
+	"The range of valid IPs is %s", testServiceCIDR)
+var errUnexpected = fmt.Errorf("%s", testServiceCIDR)
```
tpantelis: This is used in one place and can be inlined.

pkg/discovery/network/generic_test.go
```diff
@@ -28,6 +32,12 @@ import (
 const testPodCIDR = "1.2.3.4/16"
 const testServiceCIDR = "4.5.6.7/16"
 
+var errExpected = fmt.Errorf("The Service \"invalid-svc\" is invalid: "+
```
tpantelis:
```suggestion
var expectedServiceErr = fmt.Errorf("The Service \"invalid-svc\" is invalid: "+
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,89 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDR}))
+			Expect(clusterNet.PodCIDRs).To(Equal([]string{testPodCIDR}))
+		})
+
+		It("Should identify the networkplugin as generic", func() {
+			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
+		})
+	})
+
+	When("No resource and creating invalid svc returns unexpected error", func() {
```
tpantelis:
```suggestion
	When("No kube-api pod exists and invalid service creation returns an unexpected error", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,89 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		It("Should return nil cluster network", func() {
+			clusterNet, err := doTestDiscoverGenericEmulatedErrWith(errUnexpected)
+			Expect(err).To(HaveOccurred())
+			Expect(clusterNet).To(BeNil())
+		})
+	})
+
+	When("No resource and creating invalid svc returns expected error", func() {
```
tpantelis:
```suggestion
	When("No kube-api pod exists and invalid service creation returns the expected error", func() {
```

pkg/discovery/network/weavenet_test.go
```diff
@@ -73,6 +75,26 @@ var _ = Describe("discoverWeaveNetwork", func() {
 			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("weave-net"))
 		})
 	})
+
+	When("There are weave and no kube api pods, but svc creation fails properly", func() {
```
tpantelis:
```suggestion
	When("The weave pod exists but no kube-api pod exists and invalid service creation returns the expected error", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,89 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+		var clusterNet *ClusterNetwork
+
+		BeforeEach(func() {
+			clusterNet = testDiscoverGenericEmulatedErrWith(errExpected)
+		})
+
+		It("Should return the ClusterNetwork structure with the pod CIDR", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with empty pod CIDR", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -157,6 +167,89 @@ var _ = Describe("discoverGenericNetwork", func() {
...
+			Expect(clusterNet.PodCIDRs).To(BeEmpty())
+		})
+
+		It("Should identify the networkplugin as generic", func() {
+			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo("generic"))
+		})
+
+		It("Should return the ClusterNetwork structure with empty service CIDR", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with the service CIDR", func() {
```

pkg/discovery/network/canal_test.go
```diff
@@ -50,6 +52,18 @@ var _ = Describe("discoverCanalFlannelNetwork", func() {
...
+		It("Should return the ClusterNetwork structure with ServiceCIDRs too", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with Pod and Service CIDRs", func() {
```

tpantelis (changes requested):

pkg/discovery/network/generic.go
mkimuram: Changed it to return an error properly and changed unit tests as well.

pkg/discovery/network/generic.go
mkimuram: This code is called from both command line and operator.
subctl actually calls it on subctl join before submariner-namespace is created, so it shouldn't create service in submariner-operator namespace. On the other hand, the operator is only allowed to create resources inside submariner-operator namespace, so it shouldn't create service in default namespace.

Therefore, the code was changed to check whether it is running in operator and to create service in the proper namespace.

mkimuram:
@tpantelis

Thank you so much for your review with detailed suggestions. PTAL

pkg/discovery/network/generic.go
```diff
@@ -62,13 +59,122 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+					TargetPort: intstr.IntOrString{
+						IntVal: 443,
+					},
+				},
+			},
+		},
+	}
+
+	// decide which namespace to create the service
+	ns := os.Getenv("WATCH_NAMESPACE")
```
tpantelis, line 98: What is `WATCH_NAMESPACE` and how is it set?

pkg/discovery/network/generic.go
```diff
@@ -62,13 +59,122 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+	// use "submariner-operator" if WATCH_NAMESPACE is "submariner-operator" (in operator)
+	if ns != "submariner-operator" {
```
tpantelis: There's a constant for this in _root.go_.

pkg/discovery/network/generic.go
```diff
@@ -62,13 +59,122 @@ func discoverGenericNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, er
...
+		// otherwise, use "default" (in subctl command)
+		ns = "default"
+	}
+
+	// create service to the namespace
+	_, err := clientSet.CoreV1().Services(ns).Create(invalidSvcSpec)
+
+	// creating invalid service didn't fail as expected
+	if err == nil {
+		return "", fmt.Errorf("creating invalid service(%v) didn't fail", invalidSvcSpec)
```
tpantelis, line 110: I prefer this the way you had it, ie not return error here or on L128. This makes it consistent with the other discovery methods and, on join, will result in the user being prompted. But, as I noted earlier, I think it's worth logging a message as the API server didn't behave as expected.

pkg/discovery/network/canal_test.go
```diff
@@ -20,25 +20,25 @@ import (
...
 			clusterNet := testDiscoverCanalFlannelWith(&canalFlannelCfgMap)
 			Expect(clusterNet).NotTo(BeNil())
 			Expect(clusterNet.NetworkPlugin).To(Equal("canal-flannel"))
 			Expect(clusterNet.PodCIDRs).To(Equal([]string{testCannalFlannelPodCIDR}))
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDRFromService}))
 		})
 	})
 
-	When("There is a kubeapi pod at least ", func() {
+	When("There is a kube-api pod at least ", func() {
```
tpantelis:
```suggestion
	When("There is a kube-api pod", func() {
```

pkg/discovery/network/generic.go
mkimuram: This is set [here](https://github.com/submariner-io/submariner-operator/blob/d48d547be4ec4ed7c17dbcbc2e70dd89eb7d6fcc/pkg/subctl/operator/common/operatorpod/ensure.go#L72) as an environment variable of the operator pod. I think that this is used by operator-sdk's `GetWatchNamespace`.

pkg/discovery/network/generic.go
mkimuram: The callers of `discoverGenericNetwork` are as follows:

```
[Operator]
Reconcile
  discoverNetwork
    getClusterNetwork
      Discover
        discoverGenericNetwork

[Command line]
getNetworkDetails
  Discover
    discoverGenericNetwork
  if err != nil {
    status.QueueWarningMessage(fmt.Sprintf("Error trying to discover network details: %s", err))
  }
```

So, as for command line case, the error is already been passed to `status.QueueWarningMessage`.

And, other functions called from `discoverGenericNetwork`, like `findClusterIPRangeFromApiserver` actually returns error and not catching it with `status.QueueWarningMessage`.

Therefore, the code should be already as you expected.

pkg/discovery/network/generic.go
mkimuram: It seems that importing constant from `pkg/subctl/cmd` to `pkg/discovery/network/generic.go` makes `import cycle not allowed` error. Is it OK to just define constant in `pkg/discovery/network/generic.go` or do we need to move the constant to the other place to share it?

Note that as explained, subctl create the operator pod in the namespace and pass the env variable via the downward API (`metadata.namespace`). Then inside the operator, it checks whether env variable is set to check if it runs inside the operator. So, this constant is not directly shared across these functions.

pkg/discovery/network/generic.go
tpantelis: OK - could you add a comment explaining this? eg "_WATCH_NAMESPACE will be present if running in the operator pod_".

Another approach is to pass in the NS so we don't rely on the presence of an env var. The operator code implicitly knows its own namespace. I think this would make it clearer and although it would mean passing a param down the chain.

pkg/discovery/network/generic.go
tpantelis: If we pass the NS into the function as described above, this becomes moot.

pkg/discovery/network/generic.go
tpantelis: My point was about whether it should return an error in this case, ie should we treat it like we didn't find the data similar to the other discovery methods. So for `subclt join`, this would at least allow it to prompt the user as a last resort. After all, this is just another attempt to find the data. This seems reasonable but it would be beneficial to print a warning. @mangelajo WDYT?

pkg/discovery/network/generic.go
mkimuram: Changed and squashed.

pkg/discovery/network/generic.go
mkimuram: Waiting for @mangelajo 's comment. But I'm now inclined to return nil and log error here, only for this case.

Meanwhile, I would like to confirm that it is safe to call `status.QueueWarningMessage` outside CLI's context, or from operator. It looks like a package under `pkg/internal/cli` and I'm not familiar with its implementation details.

tpantelis: Unfortunately `status.QueueWarningMessage` wouldn't be appropriate in the operator. The operator uses a different logging API. We could pass in a function to log warnings.

pkg/discovery/network/generic_test.go
```diff
@@ -17,54 +17,93 @@ limitations under the License.
...
 	})
 
 	When("There is a kube-controller with no expected parameters", func() {
-		It("Should return nil cluster network", func() {
-			clusterNet := testDiscoverGenericWith(
+		var clusterNet *ClusterNetwork
+
+		BeforeEach(func() {
+			clusterNet = testDiscoverGenericWith(
 				fakePod("kube-controller", []string{"kube-controller", "--cluster-ABCD=1.2.3.4"}, []v1.EnvVar{}),
```
sridhargaddam:
```suggestion
				fakePod("kube-controller-manager", []string{"kube-controller-manager", "--cluster-ABCD=1.2.3.4"}, []v1.EnvVar{}),
```

sridhargaddam: In the code, we are actually checking for component `kube-controller-manager` and this test-case is about validating the arguments (i.e., no expected params).

pkg/discovery/network/generic_test.go
```diff
@@ -17,54 +17,93 @@ limitations under the License.
...
 	})
 
 	When("There is a kube-api with no expected parameters", func() {
-		It("Should return nil cluster network", func() {
-			clusterNet := testDiscoverGenericWith(
+		var clusterNet *ClusterNetwork
+
+		BeforeEach(func() {
+			clusterNet = testDiscoverGenericWith(
 				fakePod("kube-controller", []string{"kube-api", "--cluster-ABCD=1.2.3.4"}, []v1.EnvVar{}),
```
sridhargaddam:
```suggestion
				fakePod("kube-apiserver", []string{"kube-apiserver", "--cluster-ABCD=1.2.3.4"}, []v1.EnvVar{}),
```

pkg/discovery/network/generic_test.go
mkimuram: Thank you for pointing it out. Fixed.

pkg/discovery/network/generic_test.go
mkimuram: Fixed as well.

pkg/discovery/network/generic.go
mkimuram: Added logging code in `controllers/submariner/submariner_networkdiscovery.go` for operator to log it (See the first commit). Also, re-organized and squashed commits to review easier.

sridhargaddam (approved):
Thanks for the PR and also addressing the comments @mkimuram

mangelajo (approved):
This looks good to me now, the additional heuristics will make this much more robust. Thank you @mkimuram .

mangelajo:
@tpantelis have an eye, I belive mkimura handled all your comments.

tpantelis:
> @tpantelis have an eye, I belive mkimura handled all your comments.

Mostly.  So we're going to fail fast on `join` if the service creation API doesn't fail with the CIDR info as expected. I'm fine with that but I'd like to make the messages clearer and add a comment wrt to the `WATCH_NAMESPACE`.  I can push a follow-up PR.

tpantelis (approved):

PR #1062, opened 2021-02-03
Log status update errors without failing

skitt:
On OpenShift, we’re seeing reconciliation loops with continuous errors
when updating the status. To avoid this (temporarily), only log the
error, and indicate success; the status should end up being updated at
some point...

### DIFF
--- a/controllers/submariner/submariner_controller.go
+++ b/controllers/submariner/submariner_controller.go
@@ -216,7 +216,10 @@ func (r *SubmarinerReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		err := r.client.Status().Update(context.TODO(), instance)
 		if err != nil {
 			reqLogger.Error(err, "failed to update the Submariner status")
-			return reconcile.Result{}, err
+			// Log the error, but indicate success, to avoid reconciliation storms
+			// TODO skitt determine what we should really be doing for concurrent updates to the Submariner CR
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

controllers/submariner/submariner_controller.go
```diff
@@ -216,7 +216,10 @@ func (r *SubmarinerReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		err := r.client.Status().Update(context.TODO(), instance)
 		if err != nil {
 			reqLogger.Error(err, "failed to update the Submariner status")
-			return reconcile.Result{}, err
+			// Log the error, but indicate success, to avoid reconciliation storms
+			// TODO skitt determine what we should really be doing for concurrent updates to the Submariner CR
+			// Updates fail here because the instance is updated between the .Update() at the start of the function
+			// and the status update here
```
tpantelis, line 221: Perhaps we should just do one `Update` call and use `CreateOrUpdate` semantics here.

mangelajo (approved):

controllers/submariner/submariner_controller.go
skitt: The two updates touch two different objects, one inside the other. The first updates the `Submariner` object, before we do much work at all (see the attached comment); the second only updates the status. It seems to me now that if we end up with a version conflict at this point, we _shouldn’t_ update the status at all, because we know that the status is outdated.

PR #1074, opened 2021-02-10
Enable linters: stylecheck, goconst, prealloc, dogsled

Jaanki:


dfarrell07:
Hey @Jaanki, thanks for the PR! Can we mark this as a Draft while you're still working on getting CI passing?

There's more information here, including a nice tip Stephen added to mark PRs as Drafts while creating them, avoiding automatically adding the code owners.

https://submariner.io/development/code-review/#mark-work-in-progress-prs-as-drafts

.golangci.yml
```diff
@@ -58,3 +60,18 @@ issues:
     - linters:
         - lll
       path: "pkg/subctl/operator/common/embeddedyamls/yamls.go"
+
+    # Allow dot imports for `fake` directory and e2e tests
+    - linters:
+        - stylecheck
+      text: "ST1001: "
+      path: _test_suite\.go
```
tpantelis: The only file I see that matches this pattern is _pkg/broker/broker_test_suite.go_. You should rename this file properly to _broker_suite_test.go_ instead so you don't need this rule.

.golangci.yml
```diff
@@ -58,3 +60,18 @@ issues:
...
+
+    - linters:
+        - stylecheck
+      text: "ST1003: "
+      path: pkg/subctl/operator/common/embeddedyamls/yamls.go
```
tpantelis: Can you add a comment explaining this error and why it's ignored?

.golangci.yml
```diff
@@ -58,3 +60,18 @@ issues:
...
+
+    - linters:
+        - stylecheck
+      text: "ST1003: struct field BrokerK8sApiServer"
```
tpantelis, line 69: What is this rule for? It's not associated with a path.

tpantelis (changes requested):

.golangci.yml
Jaanki: Its to ignore warnings suggesting changing it to `BrokerK8sAPIServer`

.golangci.yml
Jaanki: Not exactly. This is in accordance with [GIngko framework](https://onsi.github.io/ginkgo/#bootstrapping-a-suite). However, there are no specs attached with this suite so basically this file is not needed unless we add any specs to this suite.

mangelajo (approved):

vthapar (approved):

.golangci.yml
tpantelis: The file name is not in accordance with the Go test framework nor Gingko, hence the stylecheck error.  There's already a rule (somewhere) that ignores dot import for _*_test.go_  so renaming the file fixes the error. However, since this file isn't really needed, I suggest to remove it - if we ever add unit tests in the future then it can be added with the correct name.

nyechiel:
@Jaanki this one is a bit stale, let's get it merged :)

.golangci.yml
Jaanki: That rule to ignore dot imports in _test.go is in `submariner` repo.

For now, I will just remove the file.

tpantelis: > That rule to ignore dot imports in _test.go is in `submariner` repo.
> 

hmm... then how are dot imports ignored for the other __test.go_ files in this repo (and in admiral, lighthouse etc)? There must be something in shipyard ...  @dfarrell07 do you know?

dfarrell07: We have rules like this to ignore dot imports for Go test frameworks:

https://github.com/submariner-io/admiral/blob/devel/.golangci.yml#L69

tpantelis: > We have rules like this to ignore dot imports for Go test frameworks:
> 
> https://github.com/submariner-io/admiral/blob/devel/.golangci.yml#L69

ok but I don't see those rules in other projects. Not a big deal - just wondering...

dfarrell07: I think for any repos without that ignore, it's just that golint and stylecheck are't enabled.

https://github.com/submariner-io/lighthouse/blob/devel/.golangci.yml#L43

tpantelis: ahh right except this PR enables stylecheck but doesn't add a general dot import rule like admiral has and yet  the test files aren't being flagged, except _pkg/broker/broker_test_suite.go_ which doesn't follow the test file naming convention. Also there is no dot import rule in submariner either. As a test I removed the rules in admiral and ran golangci-lint and it only reported dot import errors for test utils and fakes that don't end in __test.go_.  So a general rule is coming from somewhere.

mangelajo: But can't we just rename it to suite_test, remove this and move on? 

```
[majopela@bluehat submariner-operator]$ find ./ -name "*suite_test.go"
./pkg/discovery/globalnet/globalnet_suite_test.go
./pkg/discovery/network/network_suite_test.go
./controllers/servicediscovery/servicediscovery_suite_test.go
./controllers/submariner/submariner_suite_test.go

[majopela@bluehat submariner-operator]$ find ./ -name "*test_suite.go"
./pkg/broker/broker_test_suite.go
```

I honestly agree with @tpantelis initial comment, see above ^

mangelajo (changes requested):
Please rename broker_test_suite.go
to broker_suite_test.go

and remove the exception.

.golangci.yml
tpantelis: yeah we can except there's no tests.  If we plan to add them then we can rename and keep it. I'm fine either way.

mangelajo (approved):

tpantelis (approved):

mangelajo:
@Jaanki something is failing on the unit tests ^

dfarrell07 (approved):

skitt (approved):

PR #1085, opened 2021-02-12
Handle the broker in the operator

### LINKED ISSUE #203: Move broker handling to the operator (out of subctl)
Currently the broker creation is managed from subctl, while it should go into the operator itself (for future coordinated upgrade handling, etc...)

skitt:
Fixes: #203


skitt:
This PR updates `subctl` so that `deploy-broker` goes through the operator, but it doesn’t remove all the non-operator-based code — I’m wondering whether we should add a flag to allow broker deployment without the operator, or if we should just forget that and only support the operator.

mangelajo:
> This PR updates `subctl` so that `deploy-broker` goes through the operator, but it doesn’t remove all the non-operator-based code — I’m wondering whether we should add a flag to allow broker deployment without the operator, or if we should just forget that and only support the operator.

+1 for operator only for now...

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -124,8 +124,54 @@ type SubmarinerList struct {
 	Items           []Submariner `json:"items"`
 }
 
+// BrokerSpec defines the desired state of Broker
+// +k8s:openapi-gen=true
+type BrokerSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Components                  []string `json:"components,omitempty"`
```
mangelajo, line 135: Thanks for updating last moment, I just realized you were into this too :/

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -124,8 +124,54 @@ type SubmarinerList struct {
...
+	DefaultCustomDomains        []string `json:"defaultCustomDomains,omitempty"`
+	GlobalnetCidrRange          string   `json:"globalnetCidrRange,omitempty"`
+	DefaultGlobalnetClusterSize uint     `json:"defaultGlobalnetClusterSize"`
+	GlobalnetEnabled            bool     `json:"globalnetEnabled"`
```
mangelajo: Can we make GlobalNetEnabled an optional/omitempty argument?, same for DefaultGlobalnetClusterSize?, is it possible -an alternative- to provide default values and allow omission?

Is it really necessary?, could it be that "GlobalnetCidrRange == GlobalnetEnabled?" (not sure about it though).

skitt: How about making Globalnet a component? I’ll try that in a separate PR... Also, Globalnet can be enabled without a CIDR range, can’t it?

config/samples/submariner_v1alpha1_broker.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: submariner.io/v1alpha1
+kind: Broker
+metadata:
+  name: broker-sample
+spec:
+#  defaultCustomDomains:
+#  globalnetCidrRange:
+  defaultGlobalnetClusterSize: 8192
```
mangelajo: Is there a way to allow omission of 'defaultGlobanetClusterSize', and also, provide a default value? I guess that can be done in code.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -98,9 +101,26 @@ var deployBroker = &cobra.Command{
 		exitOnError("The provided kubeconfig is invalid", err)
 
 		status := cli.NewStatus()
-		status.Start("Deploying broker")
-		err = broker.Ensure(config)
+
+		status.Start("Setting up broker RBAC")
+		err = broker.Ensure(config, false)
```
mangelajo, line 106: I was expecting that the operator would create the RBAC details too, I guess it's not very secure though, but thinking of usability and the concept of doing it on the operator... it's weird that we would need to setup RBAC manually if not doing it with subctl.

pkg/subctl/operator/brokercr/ensure.go
```diff
@@ -0,0 +1,77 @@
...
+	_, err = createBroker(clientSet, namespace, brokerCR)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createBroker(clientSet submarinerclientset.Interface, namespace string, brokerCR *submarinerv1a1.Broker) (bool, error) {
```
mangelajo: Per @tpantelis I recently used the admiral util CreateOrUpdate, which resulted in more compact code.

https://github.com/submariner-io/submariner-operator/pull/1087/files#diff-8d3de7853ec49b469f54faa23b887aaea257e63e6cf1d5bfb891de5c2813685cR65

It has the disadvantage of needing the dynamic client, but the advantage that we don't need a different createOrUpdate per api type :)

pkg/subctl/operator/submarinerop/crds/ensure.go
```diff
@@ -40,7 +45,12 @@ func Ensure(restConfig *rest.Config) (bool, error) {
 	// Attempt to update or create the CRD definition
 	// TODO(majopela): In the future we may want to report when we have updated the existing
 	//                 CRD definition with new versions
-	return utils.CreateOrUpdateCRD(crdUpdater, submarinerCrd)
+	submarinerCreated, err := utils.CreateOrUpdateCRD(crdUpdater, submarinerCrd)
```
mangelajo: You can use 
```
 CreateOrUpdateEmbeddedCRD(updater crdutils.CRDUpdater, crdYaml string) (bool, error
```
In this case, and save the getBrokerCRD / getSubmarinerCRD :)

mangelajo (changes requested):
Some comments

apis/submariner/v1alpha1/submariner_types.go
mangelajo: Hmm, that'd be rather neat I think @skitt 

Without a CIDR range I'm not sure, I guess we need something at least for auto-assignment of clusters. But if you don't want auto-assignment I guess you don't need it @sridhargaddam  ^

mangelajo:
Sorry, the second batch of comments came separate.

pkg/subctl/cmd/deploybroker.go
skitt: We discussed this in the past, and decided to forgo dealing with RBAC in the operator, because it requires granting too many privileges to the operator.

config/samples/submariner_v1alpha1_broker.yaml
skitt: This is just an example, intended for documentation in the operator when shown in the catalog.

apis/submariner/v1alpha1/submariner_types.go
sridhargaddam: Correct, the globalCIDR is required for auto-assignment to clusters while the cluster is joining the Broker. We provide flexibility to users to specify custom globalCIDR, but IMO we should continue to have a default CIDR to support clusters that do not explicitly specify globalCIDR while joining.

pkg/subctl/operator/brokercr/ensure.go
skitt: In this case though I don’t want to create or update, I want to create or delete then create.

pkg/subctl/operator/submarinerop/crds/ensure.go
skitt: Indeed, thanks!

apis/submariner/v1alpha1/submariner_types.go
skitt: The fields are now all optional.

skitt:
This needs #1096 first.

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -124,8 +124,54 @@ type SubmarinerList struct {
...
+// +k8s:openapi-gen=true
+type BrokerSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Components                  []string `json:"components,omitempty"`
+	DefaultCustomDomains        []string `json:"defaultCustomDomains,omitempty"`
+	GlobalnetCidrRange          string   `json:"globalnetCidrRange,omitempty"`
+	DefaultGlobalnetClusterSize uint     `json:"defaultGlobalnetClusterSize,omitempty"`
+	GlobalnetEnabled            bool     `json:"globalnetEnabled,omitempty"`
```
sridhargaddam, line 139: We might require changes to subctl to prompt the user to provide the GlobalnetCIDR/GlobalnetClusterSize while joining a new Cluster, if there are no defaults configured when Broker is deployed.

apis/submariner/v1alpha1/submariner_types.go
skitt: https://github.com/submariner-io/submariner-operator/issues/1102 tracks the componentisation of Globalnet.

apis/submariner/v1alpha1/submariner_types.go
skitt: This is already the case, I’ll do that in a separate PR. https://github.com/submariner-io/submariner-operator/issues/1103 tracks this.

pkg/subctl/cmd/deploybroker.go
mangelajo: I know, I remember, but I start to see issues with our thinking as we move the broker here too, which is different to just "provide RBAC to submariner services", ... now we need to provide RBAC / Accounts to signed up clusters...

We will need to document the RBAC steps unless we change that.

pkg/subctl/operator/brokercr/ensure.go
mangelajo: Fair enough

apis/submariner/v1alpha1/submariner_types.go
mangelajo: I suggest we do something like:

https://github.com/submariner-io/submariner-operator/blob/devel/apis/submariner/v1alpha1/submariner_types.go#L131

https://github.com/submariner-io/submariner-operator/blob/devel/apis/submariner/v1alpha1/servicediscovery_types.go#L84

mangelajo (changes requested):

apis/submariner/v1alpha1/submariner_types.go
mangelajo: Well, or may be not, not sure, if they are going through the process of using the operator directly.... may be it's better to let them be explicit

mangelajo (approved):

config/crd/kustomization.yaml
```diff
@@ -5,19 +5,22 @@
 resources:
 #  - bases/submariner.io_servicediscoveries.yaml
   - bases/submariner.io_submariners.yaml
+#  - bases/submariner.io_brokers.yaml
```
SteveMattar: Do we want to hide this CRD? how do we install the broker when working with the bundle only?

main.go
```diff
@@ -206,6 +209,15 @@ func main() {
...
 	}
+	if err = (&submariner.BrokerReconciler{
+		Client: mgr.GetClient(),
+		Config: mgr.GetConfig(),
+		Log:    ctrl.Log.WithName("controllers").WithName("Broker"),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		log.Error(err, "unable to create controller", "controller", "Broker")
+		os.Exit(1)
+	}
```
SteveMattar, line 220: I think you can add this function to `controllers.AddToManager` this is how we usually start a new Reconciler

SteveMattar (commented):
What about the broker RBAC? Currently the only way it works is by using `subctl`.
Should we export these roles to `config/rbac`?

SteveMattar:
Sorry @skitt just saw the previous comments

pkg/subctl/cmd/deploybroker.go
SteveMattar: Can we remove the `rbac.go` in favour of working with embedded yaml files?
The yaml files could be tracked in `config/rbac`. This way I can use them in the operator bundle.

main.go
skitt: This is what the operator SDK generated for me...

config/crd/kustomization.yaml
skitt: Ah, right, we do want to show the CRD!

skitt:
> What about the broker RBAC? Currently the only way it works is by using `subctl`.
> Should we export these roles to `config/rbac`?

This is what I’d asked you about a couple of weeks ago; I couldn’t get the SDK to generate the appropriate `role.yaml` file, or rather, I could, but it overwrote all the existing RBAC setup that’s in there.

SteveMattar:
> > What about the broker RBAC? Currently the only way it works is by using `subctl`.
> > Should we export these roles to `config/rbac`?
> 
> This is what I’d asked you about a couple of weeks ago; I couldn’t get the SDK to generate the appropriate `role.yaml` file, or rather, I could, but it overwrote all the existing RBAC setup that’s in there.

Sorry you might have missed my message on slack... the thread is confusing sometimes :)
I'll just drop my comment here:
"I found the problem... when controller-gen generates the rbac manifests the default output is `config/rbac/role.yaml` but the `+kubebuilder:rbac` markers for `submariner` and `servicediscovery` never worked... The reason is because there is a missing empty line between those markers and the class definition. That means the `role.yaml` file was never updated."

skitt:
> > > What about the broker RBAC? Currently the only way it works is by using `subctl`.
> > > Should we export these roles to `config/rbac`?
> > 
> > 
> > This is what I’d asked you about a couple of weeks ago; I couldn’t get the SDK to generate the appropriate `role.yaml` file, or rather, I could, but it overwrote all the existing RBAC setup that’s in there.
> 
> Sorry you might have missed my message on slack... the thread is confusing sometimes :)
> I'll just drop my comment here:
> "I found the problem... when controller-gen generates the rbac manifests the default output is `config/rbac/role.yaml` but the `+kubebuilder:rbac` markers for `submariner` and `servicediscovery` never worked... The reason is because there is a missing empty line between those markers and the class definition. That means the `role.yaml` file was never updated."

Right, that’s what I found out too (which is how I disabled the Broker RBAC handling, see the `TODO` there). I hesitated to re-enable the full RBAC handling because it produces `ClusterRole`s, not `Role`s as we currently have...

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -124,8 +124,54 @@ type SubmarinerList struct {
...
 
+// BrokerSpec defines the desired state of Broker
+// +k8s:openapi-gen=true
+type BrokerSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Components                  []string `json:"components,omitempty"`
+	DefaultCustomDomains        []string `json:"defaultCustomDomains,omitempty"`
+	GlobalnetCidrRange          string   `json:"globalnetCidrRange,omitempty"`
```
tpantelis: We've been trending from `Cidr` -> `CIDR` elsewhere... 

```suggestion
	GlobalnetCIDRRange          string   `json:"globalnetCIDRRange,omitempty"`
```

apis/submariner/v1alpha1/submariner_types.go
skitt: I prefer https://github.com/submariner-io/submariner-operator/pull/1104 ;-)

pkg/subctl/operator/brokercr/ensure.go
tpantelis: Do we need to return `bool` here instead of just `err`?

pkg/subctl/operator/brokercr/ensure.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			_, err = clientSet.SubmarinerV1alpha1().Brokers(namespace).Create(brokerCR)
+			if err != nil {
+				return fmt.Errorf("failed to create cfg  %s : %s", brokerCR.Name, err)
+			}
+			return nil
+		})
+		return false, retryErr
+	}
+	return false, err
```
tpantelis: We really don't need `RetryOnConflict` as it applies to Update. We could refactor to a loop as per below but not a big deal either way. I can add a CreateAnew function in admiral that does the DeleteIfExistsThenCreate pattern.

```suggestion
            for {
		_, err := clientSet.SubmarinerV1alpha1().Brokers(namespace).Create(brokerCR)
		if err == nil {
			return nil
		}

		if errors.IsAlreadyExists(err) {
			err := clientSet.SubmarinerV1alpha1().Brokers(namespace).Delete(brokerCR.Name, &metav1.DeleteOptions{})
			if err != nil && !errors.IsNotFound(err) {
				return fmt.Errorf("failed to delete pre-existing Broker instance %q : %v", brokerCR.Name, err)
			}
			
			continue
		} 
		
                 return err
            }

            return nil
```

apis/submariner/v1alpha1/submariner_types.go
skitt: Ah yes, fixed.

pkg/subctl/operator/brokercr/ensure.go
skitt: Ah yes, a creation can never return an error which RetryOnConflict will recognise as a conflict. We’d really have to loop ourselves, with a backoff...

skitt: (I’ll submit an appropriate function for Admiral tomorrow.)

tpantelis: I'm actually working on that now 😄

tpantelis: https://github.com/submariner-io/admiral/pull/185

pkg/broker/ensure.go
```diff
@@ -34,7 +34,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-func Ensure(config *rest.Config) error {
```
mangelajo, line 37: Why is the crds bool added, where is it used? :-?

skitt: Oops, something got lost in a rebase it seems...

skitt: Now restored.

pkg/subctl/cmd/deploybroker.go
skitt: @SteveMattar this is tracked in https://github.com/submariner-io/submariner-operator/issues/1105.

SteveMattar (approved):

sridhargaddam (approved):

tpantelis (approved):

PR #1093, opened 2021-02-16
When replacing a Submariner CR, wait for deletion

skitt:
Before creating the new Submariner CR, wait for the old one to be
completely deleted, including all its owned resources. This should
make upgrades much more reliable.

(Ultimately we need to ensure we cope with fast re-creations...)

### DIFF
--- a/pkg/subctl/operator/submarinercr/ensure.go
+++ b/pkg/subctl/operator/submarinercr/ensure.go
@@ -47,6 +47,9 @@ func Ensure(config *rest.Config, namespace string, submarinerSpec submariner.Sub
 		Group:    submariner.SchemeGroupVersion.Group,
 		Version:  submariner.SchemeGroupVersion.Version,
 		Resource: "submariners"}).Namespace(namespace)
+	propagationPolicy := metav1.DeletePropagationForeground
 
-	return util.CreateAnew(client, submarinerCR)
+	return util.CreateAnew(client, submarinerCR, &metav1.DeleteOptions{
+		PropagationPolicy: &propagationPolicy,
+	})
 }
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/prometheus/client_golang v1.9.0
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
-	github.com/submariner-io/admiral v0.9.0-m1
+	github.com/submariner-io/admiral v0.9.0-m1.0.20210302163417-57a3ab841d27
 	github.com/submariner-io/lighthouse v0.9.0-m1
 	github.com/submariner-io/shipyard v0.9.0-m1
 	github.com/submariner-io/submariner v0.9.0-m1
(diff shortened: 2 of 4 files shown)


pkg/subctl/operator/submarinercr/ensure.go
```diff
@@ -61,10 +63,26 @@ func createSubmariner(clientSet submarinerclientset.Interface, namespace string,
...
 			}
+			// We need to wait for the deletion to be complete
+			err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
+				_, err := clientSet.SubmarinerV1alpha1().Submariners(namespace).Get(submarinerCR.Name, metav1.GetOptions{})
+				if errors.IsNotFound(err) {
+					return true, nil
+				} else if err != nil {
+					return false, err
+				}
+				return false, nil
```
tpantelis:
```suggestion
				if errors.IsNotFound(err) {
					return true, nil
				} 
				
				return false, err
```

tpantelis (commented):
Just a minor comment.

mangelajo:
@skitt I'm trying to remember why didn't we update the objects (instead of delete/re-create).

pkg/subctl/operator/submarinercr/ensure.go
skitt: Indeed, thanks for the suggestion!

skitt:
> @skitt I'm trying to remember why didn't we update the objects (instead of delete/re-create).

Because there’s nothing we ever want to carry over from the old CR, and by deleting it, we get a lot of processing done for us (cascade-deleting all the dependent resources).

tpantelis (approved):

tpantelis (approved):

pkg/subctl/operator/submarinercr/ensure.go
```diff
@@ -61,10 +63,24 @@ func createSubmariner(clientSet submarinerclientset.Interface, namespace string,
...
-			err := clientSet.SubmarinerV1alpha1().Submariners(namespace).Delete(submarinerCR.Name, &metav1.DeleteOptions{})
+			propagationPolicy := metav1.DeletePropagationForeground
+			err := clientSet.SubmarinerV1alpha1().Submariners(namespace).Delete(submarinerCR.Name, &metav1.DeleteOptions{
+				PropagationPolicy: &propagationPolicy,
+			})
 			if err != nil {
 				return fmt.Errorf("failed to delete pre-existing cfg %s : %s", submarinerCR.Name, err)
 			}
+			// We need to wait for the deletion to be complete
+			err = wait.PollImmediate(1*time.Second, 10*time.Second, func() (bool, error) {
```
mangelajo: 10 seconds would be enough?, I've seen resources taking longer to delete, but may be resources which have callbacks ... seems short to me anyway

tpantelis: If we bump the timeout then I think we should periodically print a message "_Waiting for deletion of previous Submariner resource %q to complete_" - like every 5 sec.

tpantelis (approved):

dfarrell07 (approved):

PR #1104, opened 2021-02-22
Specify defaults when unmarshalling

skitt:
... instead of requiring a call to SetDefaults().


tpantelis (approved):

tpantelis:
Nice  👍

mangelajo (approved):

PR #1109, opened 2021-02-24
Deploy CRDs only as required by components

### LINKED ISSUE #1098: Skip SubmarinerCRs install for lighthouse only deployments
Now we have the option to install Lighthouse only without any Submariner components for use cases where user only wants Service Discovery from Lighthouse but not cross cluster connectivity from Submariner. But it still installs gateways, clusters and endpoint CRDs. This breaks logic in lighthouse that looks for gateways CR to see if submariner is providing connectivity or not.

skitt:
We no longer need the Submariner CRDs all the time; only when
connectivity is required. Service-discovery-only deployments shouldn’t
have the Submariner CRDs.

The operator however needs the CRDs for resources it manages; this
includes service discovery, so that CRD is added when setting up the
operator.

This patch distinguishes deployments by components, and only sets up
the relevant CRDs. It includes a stub Globalnet component which
shouldn’t be used as such but will be fleshed out later.

Fixes: #1098


pkg/subctl/datafile/datafile.go
```diff
@@ -160,8 +161,11 @@ func (data *SubctlData) getAndCheckBrokerAdministratorConfig(private bool) (*res
 	if err != nil {
 		return config, err
 	}
-	// The point of the broker client is to access Submariner data, we know we should able to list clusters
+	// We attempt to access a Submariner object; if we can’t find this, we know we still connected
```
tpantelis: ?

```suggestion
	// We attempt to access a Submariner object; if we can’t find this, we know we can still connect
```

pkg/broker/ensure.go
```diff
@@ -22,32 +22,44 @@ import (
...
 
-		_, err = lighthouse.Ensure(crdCreator, lighthouse.BrokerCluster)
-		if err != nil {
-			return fmt.Errorf("error setting up the lighthouse requirements: %s", err)
+		for i := range componentArr {
+			switch componentArr[i] {
+			case components.Connectivity:
+				err = engine.Ensure(crdCreator)
+				if err != nil {
+					return fmt.Errorf("error setting up the engine requirements: %s", err)
```
tpantelis:
```suggestion
					return fmt.Errorf("error setting up the connectivity requirements: %s", err)
```

pkg/broker/ensure.go
```diff
@@ -22,32 +22,44 @@ import (
...
+				}
+			case components.ServiceDiscovery:
+				_, err = lighthouse.Ensure(crdCreator, lighthouse.BrokerCluster)
+				if err != nil {
+					return fmt.Errorf("error setting up the lighthouse requirements: %s", err)
```
tpantelis:
```suggestion
					return fmt.Errorf("error setting up the service discovery requirements: %s", err)
```

pkg/broker/ensure.go
```diff
@@ -22,32 +22,44 @@ import (
...
+				}
+			case components.Globalnet:
+				// Globalnet needs the Lighthouse CRDs too
+				_, err = lighthouse.Ensure(crdCreator, lighthouse.BrokerCluster)
+				if err != nil {
+					return fmt.Errorf("error setting up the lighthouse requirements: %s", err)
```
tpantelis:
```suggestion
					return fmt.Errorf("error setting up the globalnet requirements: %s", err)
```

tpantelis (changes requested):

vthapar (approved):

nyechiel:
@skitt this one needs some love :)

mangelajo (approved):

tpantelis (approved):

tpantelis:
The unit test job keeps failing - make fails with exit status 2 but I don't see a reason for it - all the tests pass.

dfarrell07:
> The unit test job keeps failing - make fails with exit status 2 but I don't see a reason for it - all the tests pass.

I also don't see a PR here or in Shipyard that could have caused this. I don't see similar failures in other repos.

It seems to have started just a few hours ago.

https://github.com/submariner-io/submariner-operator/actions/workflows/unit.yml

tpantelis:
> > The unit test job keeps failing - make fails with exit status 2 but I don't see a reason for it - all the tests pass.
> 
> I also don't see a PR here or in Shipyard that could have caused this. I don't see similar failures in other repos.
> 
> It seems to have started just a few hours ago.
> 
> https://github.com/submariner-io/submariner-operator/actions/workflows/unit.yml

yeah it mysteriously started. One difference between the failing runs and the successful runs prior is that the former have this near the end just prior to failing:

```
[submariner-operator]$ uname
make: *** [/opt/shipyard/Makefile.inc:100: unit] Error 1
```

Not sure of the significance or where that is coming from.

dfarrell07 (approved):

PR #1115, opened 2021-03-01
Ignore resources pending deletion

skitt:
If a resource is deleted with a grace period, we will continue
receiving reconciliation requests for it, but we should ignore them.

### DIFF
--- a/controllers/submariner/broker_controller.go
+++ b/controllers/submariner/broker_controller.go
@@ -64,6 +64,11 @@ func (r *BrokerReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) 
 		return reconcile.Result{}, err
 	}
 
+	if instance.ObjectMeta.DeletionTimestamp != nil {
+		// Graceful deletion has been requested, ignore the object
+		return reconcile.Result{}, nil
+	}
+
 	// Broker CRDs
 	crdUpdater := crdutils.NewFromControllerClient(r.Client)
 	err = gateway.Ensure(crdUpdater)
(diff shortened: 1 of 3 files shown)


controllers/submariner/broker_controller.go
```diff
@@ -63,6 +63,10 @@ func (r *BrokerReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error)
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	if instance.ObjectMeta.DeletionTimestamp != nil {
+		// Graceful deletion has been requested, ignore the object
+		return reconcile.Result{}, nil
+	}
```
tpantelis, line 70: nit - newlines for better readability.

```suggestion

	if instance.ObjectMeta.DeletionTimestamp != nil {
		// Graceful deletion has been requested, ignore the object
		return reconcile.Result{}, nil
	}
	
```

controllers/submariner/submariner_controller.go
```diff
@@ -120,6 +120,10 @@ func (r *SubmarinerReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	if instance.ObjectMeta.DeletionTimestamp != nil {
+		// Graceful deletion has been requested, ignore the object
+		return reconcile.Result{}, nil
+	}
```
tpantelis, line 127:
```suggestion

	if instance.ObjectMeta.DeletionTimestamp != nil {
		// Graceful deletion has been requested, ignore the object
		return reconcile.Result{}, nil
	}
```

tpantelis (approved):

dfarrell07 (approved):

PR #1128, opened 2021-03-04
feat(subctl): check version mismatch

### LINKED ISSUE #1048: subctl - version mismatch notification
A notification to the user that tells if `subctl` version supports the managed `submariner` version.

**Why is this needed**:
There might be a mismatch between the cli and the submariner API.

SteveMattar:
In case subctl has a lower version than submariner it could lead to a mismatch
against the API. We should notify the user and exit the program.

Closes #1048


mangelajo:
I suggest we also add a version field to the datafile.

SteveMattar:
I get an `Unauthorized` error message when trying to access the clientset.

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,27 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+	submariner, err := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if submariner != nil && submariner.Spec.Version != "" {
+		subctlVer, _ := goversion.NewVersion(version.Version)
+		submarinerVer, _ := goversion.NewVersion(submariner.Spec.Version)
+
+		if subctlVer != nil && submarinerVer != nil && subctlVer.LessThan(submarinerVer) {
```
dfarrell07, line 250: Should we only fail when the subctl version is less than the Submariner version? I thought we required them to be the same.

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,27 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+			return goerrors.New("subctl " + version.Version + " is too old. Please upgrade your version.")
```
dfarrell07: Maybe we should show the detected Submariner version here too.

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,27 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+
+	clientSet, err := submarinerclientset.NewForConfig(config)
+	exitOnError("Unable to get the Submariner config", err)
+
+	submariner, err := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if submariner != nil && submariner.Spec.Version != "" {
```
dfarrell07, line 246: Should we react (error/warning) if we don't find the expected resource/versions, instead of basically doing the same path as the a successful match? Same applies to the conditional below.

pkg/subctl/cmd/root.go
SteveMattar: The CLI should allow backward compatibility. We could define at most 3 minor versions but for now we don't have such policy.

mangelajo: No, then it's impossible to do an upgrade for example. 

But that's a good point.
We should probably test over a limited version skew, and only support that.

For example

subctl version N is expected to work with submariner version N-1, and we test that
for commands (show, etc...).

We at least do it for the join/deploy broker commands on the upgrade job. That job is probably a good opportunity to test the show command on that combination.

pkg/subctl/cmd/root.go
mangelajo: +1

pkg/subctl/cmd/root.go
mangelajo: We shouldn't I believe, otherwise you can't deploy-broker, or join a cluster :-)

The show command is ready for submariner not being installed.

mangelajo (changes requested):

pkg/subctl/cmd/root.go
SteveMattar: I'm not sure if `Spec.Version` is mandatory.

pkg/subctl/cmd/root.go
dfarrell07: > We could define at most 3 minor versions but for now we don't have such policy.

Okay makes sense, we can do that in a future PR.

dfarrell07 (approved):

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,25 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+
+	submariner, _ := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
+
+	if submariner != nil && submariner.Spec.Version != "" {
+		subctlVer, _ := goversion.NewVersion(version.Version)
+		submarinerVer, _ := goversion.NewVersion(submariner.Spec.Version)
+
+		if subctlVer != nil && submarinerVer != nil && subctlVer.LessThan(submarinerVer) {
+			return goerrors.New("Detected submariner version: " + submariner.Spec.Version +
+				". subctl " + version.Version + " is too old. Please upgrade your CLI version.")
```
tpantelis:
```suggestion
			return goerrors.New(fmt.Sprintf("The subctl version %q is older than the deployed Submariner version %q. Please upgrade your subctl version.", submariner.Spec.Version ))
```

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,25 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+func checkVersionMismatch(cmd *cobra.Command, args []string) error {
+	config, err := getRestConfig(kubeConfig, kubeContext)
+	exitOnError("The provided kubeconfig is invalid", err)
+
+	clientSet, err := submarinerclientset.NewForConfig(config)
+	exitOnError("Unable to get the Submariner config", err)
+
+	submariner, _ := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
+
+	if submariner != nil && submariner.Spec.Version != "" {
```
tpantelis: We should be consistent and fail on error as we do above:

```suggestion
	submariner, err := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
	
        if apiErrors.IsNotFound(err) {
	    return nil
         }
	
         exitOnError("Error retrieving the Submariner resource", err)

	if submariner.Spec.Version != "" {
```

Note: `import apiErrors "k8s.io/apimachinery/pkg/api/errors"`

tpantelis (changes requested):

pkg/subctl/cmd/root.go
```diff
@@ -234,3 +240,31 @@ var nodeLabelBackoff wait.Backoff = wait.Backoff{
...
+
+	clientSet, err := submarinerclientset.NewForConfig(config)
+	exitOnError("Unable to get the Submariner config", err)
+
+	submariner, err := clientSet.SubmarinerV1alpha1().Submariners(OperatorNamespace).Get(submarinercr.SubmarinerName, metav1.GetOptions{})
+
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	exitOnError("Error retrieving the Submariner resource", err)
```
tpantelis: There's a `getSubmarinerResource` function you can use.

```suggestion
	submariner := getSubmarinerResource(config)

	if submariner == nil {
	    return nil
	}
```

SteveMattar: I tried to use it but the function exits in case of an error

tpantelis: yes but you're also using `exitOnError` on L244 and L247.

SteveMattar: Oh, I've done my tests before this part was added to `getSubmarinerResource`
```
if errors.IsNotFound(err) {
		return nil
}
```
It should work fine now thanks to @sridhargaddam +1

tpantelis (approved):

sridhargaddam (approved):
LGTM, thanks @SteveMattar

dfarrell07 (approved):

skitt (approved):

dfarrell07:
Can you re-review @mangelajo? :smiley: :pray:

mangelajo (approved):

PR #1135, opened 2021-03-05
Change CreateOrUpdate functions to use admiral

tpantelis:


dfarrell07 (approved):
It seems the unit tests have a false failure?

tpantelis:
> It seems the unit tests have a false failure?

Not sure what's going on with the unit tests. They all succeed but the  process still fails with exit status 2 - I can't tell where that's coming from. This is happening in other PRs too.

I ran it locally and it succeeded.

pkg/subctl/operator/brokercr/ensure.go
```diff
@@ -38,15 +39,20 @@ func Ensure(config *rest.Config, namespace string, brokerSpec submariner.BrokerS
...
 		return err
 	}
 
-	client := dynClient.Resource(schema.GroupVersionResource{
-		Group:    submariner.SchemeGroupVersion.Group,
-		Version:  submariner.SchemeGroupVersion.Version,
-		Resource: "brokers"}).Namespace(namespace)
-
-	return util.CreateAnew(client, brokerCR, nil)
+	return util.CreateAnew(&resource.InterfaceFuncs{
```
mangelajo, line 47: CreateAnew? :) it isn't very descriptive, a client? :)

mangelajo: ah, create a new object, and you provide all the functions.

mangelajo: just opinion, but wouldn't it be more "humanized" to say:

util.CreateANew(&broker, .....) ?

mangelajo:
Why did we change it from dynamic client to clientset? the code looked cleaner with the dynamic

pkg/subctl/operator/brokercr/ensure.go
tpantelis: Not sure what you mean by "humanized"...  so you propose to capitalize the N? `Anew` is one word. Also this function is in admiral so renaming it would be outside the scope of this PR.

tpantelis:
> Why did we change it from dynamic client to clientset? the code looked cleaner with the dynamic



> Why did we change it from dynamic client to clientset? the code looked cleaner with the dynamic

Originally it did use clientsets but was changed to dynamic when initially changed to use admiral. But then the admiral API was made generic to not require dynamic. We can keep it dynamic here if you want - I'm fine either way.

pkg/subctl/operator/brokercr/ensure.go
tpantelis: @mangelajo :point_up:

skitt (approved):

mangelajo (approved):

PR #1159, opened 2021-03-10
subctl: Add subcmd to gather data

### LINKED ISSUE #1152: Add subctl command to gather data
Add a new subctl command to gather logs and data

**Why is this needed**:
Sub task of Epic https://github.com/submariner-io/submariner-operator/issues/1125

vthapar:
Fixes #1152

### DIFF
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -21,6 +21,8 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/components"
 )
 
 var (
@@ -28,16 +30,28 @@ var (
 	gatherModule string
 )
 
+const (
+	Logs      = "logs"
+	Resources = "resources"
+)
+
 var gatherModuleFlags = map[string]bool{
-	"connectivity": false,
-	"discovery":    false,
-	"broker":       false,
-	"operator":     false,
+	components.Connectivity:     false,
+	components.ServiceDiscovery: false,
+	components.Broker:           false,
+	components.Operator:         false,
 }
 
 var gatherTypeFlags = map[string]bool{
 	"logs":      false,
 	"resources": false,
+}
+
+var gatherFuncs = map[string]func(*cli.Status, string) error{
+	components.Connectivity:     gatherConnectivity,
+	components.ServiceDiscovery: gatherDiscovery,
+	components.Broker:           gatherBroker,
+	components.Operator:         gatherOperator,
 }
 
 func init() {
@@ -47,15 +61,18 @@ func init() {
 }
 
 func addGatherFlags(gatherCmd *cobra.Command) {
-	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","), "comma-separated list of data types to gather")
-	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma-separated list of components for which to gather data")
+	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","),
+		"comma-separated list of data types to gather")
+	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","),
+		"comma-separated list of components for which to gather data")
 }
 
 var gatherCmd = &cobra.Command{
-	Use:   "gather <kubeConfig1>",
+	Use:   "gather <kubeConfig>",
 	Short: "Gather troubleshooting data from a cluster",
-	Long: fmt.Sprintf("This command gathers data from a submariner cluster for troubleshooting. Data gathered" +
-		"can be selected by component (%v) and type (%v). Default is to capture all data.", strings.Join(getAllModuleKeys(), ","), strings.Join(getAllTypeKeys(), ",")),
+	Long: fmt.Sprintf("This command gathers data from a submariner cluster for troubleshooting. The data gathered "+
+		"can be selected by component (%v) and type (%v). Default is to capture all data.",
+		strings.Join(getAllModuleKeys(), ","), strings.Join(getAllTypeKeys(), ",")),
 	Run: func(cmd *cobra.Command, args []string) {
 		gatherData()
 	},
@@ -65,22 +82,57 @@ func gatherData() {
 	err := checkGatherArguments()
 	exitOnError("Invalid arguments", err)
 
-	fmt.Printf("Gathering following data for module(s): ")
 	for module, ok := range gatherModuleFlags {
 		if ok {
-			fmt.Printf("%s ", module)
+			for dataType, ok := range gatherTypeFlags {
+				if ok {
+					status := cli.NewStatus()
+					status.Start(fmt.Sprintf("Gathering %s %s...", module, dataType))
+					status.End(cli.CheckForError(gatherFuncs[module](status, dataType)))
+				}
+			}
 		}
 	}
-	for dataType, ok := range gatherTypeFlags {
-		if ok {
-			fmt.Printf("\n  * %s ", dataType)
-		}
+}
+
+func gatherConnectivity(status *cli.Status, dataType string) error {
+	switch dataType {
+	case Logs:
+		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+	case Resources:
+		status.QueueWarningMessage("Gather Connectivity Resources not implemented yet")
 	}
-	fmt.Println()
+	return nil
+}
+
+func gatherDiscovery(status *cli.Status, dataType string) error {
+	switch dataType {
+	case Logs:
+		status.QueueWarningMessage("Gather ServiceDiscovery Logs not implemented yet")
+	case Resources:
+		status.QueueWarningMessage("Gather ServiceDiscovery Resources not implemented yet")
+	}
+	return nil
+}
+
+func gatherBroker(status *cli.Status, dataType string) error {
+	if dataType == Resources {
+		status.QueueWarningMessage("Gather Broker Resources not implemented yet")
+	}
+	return nil
+}
+
+func gatherOperator(status *cli.Status, dataType string) error {
+	switch dataType {
+	case Logs:
+		status.QueueWarningMessage("Gather Operator Logs not implemented yet")
+	case Resources:
+		status.QueueWarningMessage("Gather Operator Resources not implemented yet")
+	}
+	return nil
 }
 
 func checkGatherArguments() error {
-
 	gatherTypeList := strings.Split(gatherType, ",")
 	for _, arg := range gatherTypeList {
 		if _, found := gatherTypeFlags[arg]; !found {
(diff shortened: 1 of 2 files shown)


pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func init() {
+	addKubeconfigFlag(gatherCmd)
+	addGatherFlags(gatherCmd)
+	rootCmd.AddCommand(gatherCmd)
+}
+
+func addGatherFlags(gatherCmd *cobra.Command) {
+	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","), "comma separated data type to be gathered")
```
```suggestion
	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","), "comma-separated list of data types to gather")
```

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,127 @@
...
+	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma separated modules to be gathered")
```
```suggestion
	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma- separated list of components for which to gather data")
```

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+var gatherCmd = &cobra.Command{
+	Use:   "gather <kubeConfig1>",
+	Short: "Gather troubleshooting data from a cluster",
+	Long: `This command gathers data from a submariner cluter for troubleshooting. Data gathered
+can be selected by module [connectivity, discovery, broker, operator] and type
+ [logs, resources]. Default is to capture all types of data for all modules.`,
```
```suggestion
	Long: fmt.Sprintf("This command gathers data from a submariner cluster for troubleshooting. Data gathered
can be selected by component (%v) and type (%v). Default is to capture all data.", strings.Join(getAllModuleKeys(), ","), strings.Join(getAllTypeKeys(), ",")),
```

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,127 @@
...
+	for dataType, ok := range gatherTypeFlags {
+		if ok {
+			fmt.Printf("\n  * %s ", dataType)
+		}
+	}
+	fmt.Println()
+}
+
+func checkGatherArguments() error {
+	if gatherType != "" {
```
tpantelis: nit: don't need to check this

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,127 @@
...
+		gatherTypeList := strings.Split(gatherType, ",")
+		for _, arg := range gatherTypeList {
+			if _, found := gatherTypeFlags[arg]; !found {
+				return fmt.Errorf("%s is not a supported type", arg)
+			}
+			gatherTypeFlags[arg] = true
+		}
+	}
+
+	if gatherModule != "" {
```
tpantelis: ...nor here

tpantelis (changes requested):

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	_ "github.com/submariner-io/lighthouse/test/e2e/discovery"
+	_ "github.com/submariner-io/lighthouse/test/e2e/framework"
+	_ "github.com/submariner-io/submariner/test/e2e/dataplane"
+	_ "github.com/submariner-io/submariner/test/e2e/redundancy"
```
mangelajo: why do we need those?

mangelajo (changes requested):

pkg/subctl/cmd/gather.go
vthapar: We don't. I missed removing these from my scratch code. Expected IDE to auto remove these, looks like one of my plugins broken again.

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,120 @@
...
+
+func init() {
+	addKubeconfigFlag(gatherCmd)
+	addGatherFlags(gatherCmd)
+	rootCmd.AddCommand(gatherCmd)
+}
+
+func addGatherFlags(gatherCmd *cobra.Command) {
+	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","), "comma-separated list of data types to gather")
+	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma- separated list of components for which to gather data")
```
skitt:
```suggestion
	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma-separated list of components for which to gather data")
```

skitt (changes requested):
Just a minor tweak...

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,120 @@
...
+	rootCmd.AddCommand(gatherCmd)
+}
+
+func addGatherFlags(gatherCmd *cobra.Command) {
+	gatherCmd.Flags().StringVar(&gatherType, "type", strings.Join(getAllTypeKeys(), ","), "comma-separated list of data types to gather")
+	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","), "comma-separated list of components for which to gather data")
+}
+
+var gatherCmd = &cobra.Command{
+	Use:   "gather <kubeConfig1>",
```
mangelajo: Sorry, I didn't notice this before. 

In other modules, except for verify (which needs multiple kubeconfigs),
we use just the --kubeconfig.

I think that in some cases customers don't even know where the kubeconfig could be (for example after oc login)

Should we change to --kubeconfig instead? 
may be even accept --namespace for when submariner is not in the default namespace?

mangelajo (changes requested):
Sorry, a couple of things I realized late.

pkg/subctl/cmd/gather.go
vthapar: > Sorry, I didn't notice this before.
> 
> In other modules, except for verify (which needs multiple kubeconfigs),
> we use just the --kubeconfig.
> 
> I think that in some cases customers don't even know where the kubeconfig could be (for example after oc login)
> 
> Should we change to --kubeconfig instead?
> may be even accept --namespace for when submariner is not in the default namespace?

Later we will want to add support to pull data from multiple clusters. So thought let us use `kubeConfig1` so that later on we can add more to it. As for namespace, don't we store it in Submariner CR? What if we read it off there and don't leave user any room for error?

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	Logs = "logs"
+	Resources = "resources"
+)
+var gatherModuleFlags = map[string]bool{
+	components.Connectivity:     false,
+	components.ServiceDiscovery: false,
+	components.Broker:           false,
+	components.Operator:         false,
```
skitt, line 42: Wouldn’t it make sense to distinguish Globalnet too, since it has its own agent?

vthapar: > Wouldn’t it make sense to distinguish Globalnet too, since it has its own agent?

It can fall under connectivity, we can easily detect if globalnet is enabled and get those logs too.

skitt (approved):

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+var gatherTypeFlags = map[string]bool{
+	"logs":      false,
+	"resources": false,
+}
+
+var gatherFuncs = map[string]interface{}{
```
```suggestion
var gatherFuncs = map[string]func(*cli.Status, string) error{
```

pkg/subctl/components/consts.go
```diff
@@ -20,4 +20,6 @@ const (
 	Connectivity     = "connectivity"
 	ServiceDiscovery = "service-discovery"
 	Globalnet        = "globalnet"
+	Broker           = "broker"
+	Operator         = "Operator"
```
tpantelis: For consistency as we don't capitalize the others.

```suggestion
	Operator         = "operator"
```

dfarrell07 (approved):

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,170 @@
...
+		if ok {
+			for dataType, ok := range gatherTypeFlags {
+				if ok {
+					err = gatherFuncs[module].(func(*cli.Status, string) error)(status, dataType)
+				}
+			}
+		}
+	}
+	status.End(cli.CheckForError(err))
+	exitOnError("Error deploying the operator", err)
```
tpantelis: I think it would be better to create a status spinner per module/type. This way we can pass/fail each one.

```suggestion
	for module, ok := range gatherModuleFlags {
		if ok {
			for dataType, ok := range gatherTypeFlags {
				if ok {
					status := cli.NewStatus()
					status.Start(fmt.Sprintf("Gathering %s %s...", module, dataType))
					status.End(cli.CheckForError(gatherFuncs[module](status, dataType)))
				}
			}
		}
	}
```

tpantelis (changes requested):

pkg/subctl/cmd/gather.go
vthapar: Yeah, I was in two minds on what to go for. Went with single one coz not sure we want to use a common `cli.Status` across all. Note that it is declared as a global in `cmd`. But that needs some rework, and can revisit if decide to go that way.

tpantelis: You're using a local `Status` here either way. When I was playing around with it, it seemed better to me to have a spinner/pass/fail marker next to each one.

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,174 @@
...
+		gatherData()
+	},
+}
+
+func gatherData() {
+	err := checkGatherArguments()
+	exitOnError("Invalid arguments", err)
+
+	status := cli.NewStatus()
+	status.Start("Gathering Data...")
```
tpantelis: This is left-over.

```suggestion
```

pkg/subctl/cmd/gather.go
```diff
@@ -0,0 +1,174 @@
...
+		"comma-separated list of data types to gather")
+	gatherCmd.Flags().StringVar(&gatherModule, "module", strings.Join(getAllModuleKeys(), ","),
+		"comma-separated list of components for which to gather data")
+}
+
+var gatherCmd = &cobra.Command{
+	Use:   "gather <kubeConfig>",
+	Short: "Gather troubleshooting data from a cluster",
+	Long: fmt.Sprintf("This command gathers data from a submariner cluster for troubleshooting. Data gathered"+
+		"can be selected by component (%v) and type (%v). Default is to capture all data.",
```
tpantelis: Sorry meant to note this earlier.

```suggestion
	Long: fmt.Sprintf("This command gathers data from a submariner cluster for troubleshooting. The data gathered "+
		"can be selected by component (%v) and type (%v). Default is to capture all data.",
```

pkg/subctl/cmd/gather.go
skitt: Good point, I agree having separate spinners would be better.

pkg/subctl/cmd/gather.go
vthapar: @mangelajo Changed to `--kubeconfig` for now. Will revisit when we support multiple clusters.

vthapar:
@tpantelis @mangelajo Once ready, recommend squash and merge. Almost all changes are in single file. Can reuse description of initial commit.

tpantelis (approved):

tpantelis (approved):

Jaanki (approved):

mangelajo (approved):

PR #1165, opened 2021-03-11
Add validate connections command to subctl

### LINKED ISSUE #1145: subctl validate: Verify if the gateway connections are properly established.
As part of [Epic](https://github.com/submariner-io/submariner-operator/issues/1126) this issue is to track verification of Gateway connection status on the local cluster.

maayanf24:
Verify if the gateway connections are properly established,
or Alert the user if an issue was found.

Fixes issue #1145

### DIFF
--- a/pkg/subctl/cmd/validate_connections.go
+++ b/pkg/subctl/cmd/validate_connections.go
@@ -40,7 +40,7 @@ func validateConnections(cmd *cobra.Command, args []string) {
 	exitOnError("Error getting REST config for cluster", err)
 
 	for _, item := range configs {
-		message := fmt.Sprintf("Validating Connections in %q", item.clusterName)
+		message := fmt.Sprintf("Validating connections in cluster %q", item.clusterName)
 		status.Start(message)
 		fmt.Println()
 		submariner := getSubmarinerResource(item.config)
@@ -53,23 +53,25 @@ func validateConnections(cmd *cobra.Command, args []string) {
 
 		gateways := submariner.Status.Gateways
 		if gateways == nil {
-			message = "There are no gateways detected in cluster"
-			status.QueueFailureMessage(message)
+			message = "There are no gateways detected"
+			status.QueueWarningMessage(message)
 			status.End(cli.Failure)
 			continue
 		}
 
 		allConnectionsEstablished := true
 		for _, gateway := range *gateways {
-			for _, connection := range gateway.Connections {
-				if connection.Status == submv1.Connecting {
-					message = fmt.Sprintf("Connection to cluster %q is still trying to connect", connection.Endpoint.ClusterID)
-					status.QueueFailureMessage(message)
-					allConnectionsEstablished = false
-				} else if connection.Status == submv1.ConnectionError {
-					message = fmt.Sprintf("Connection to cluster %q is not established", connection.Endpoint.ClusterID)
-					status.QueueFailureMessage(message)
-					allConnectionsEstablished = false
+			if gateway.HAStatus == submv1.HAStatusActive {
+				for _, connection := range gateway.Connections {
+					if connection.Status == submv1.Connecting {
+						message = fmt.Sprintf("Connection to cluster %q is in progress", connection.Endpoint.ClusterID)
+						status.QueueFailureMessage(message)
+						allConnectionsEstablished = false
+					} else if connection.Status == submv1.ConnectionError {
+						message = fmt.Sprintf("Connection to cluster %q is not established", connection.Endpoint.ClusterID)
+						status.QueueFailureMessage(message)
+						allConnectionsEstablished = false
+					}
 				}
 			}
 		}
@@ -79,7 +81,7 @@ func validateConnections(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		message = "All Gateways connections are established"
+		message = "All connections are established"
 		status.QueueSuccessMessage(message)
 		status.End(cli.Success)
 	}
--- a/pkg/subctl/cmd/validate_connections.go
+++ b/pkg/subctl/cmd/validate_connections.go
@@ -21,6 +21,9 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+
+	v1opts "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var validateConnectionsCmd = &cobra.Command{
@@ -51,7 +54,13 @@ func validateConnections(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		gateways := submariner.Status.Gateways
+		submarinerClient, err := submarinerv1.NewForConfig(item.config)
+		exitOnError("Unable to get the Submariner client", err)
+
+		gateways, err := submarinerClient.SubmarinerV1().Gateways(OperatorNamespace).
+			List(v1opts.ListOptions{})
+		exitOnError("Unable to get the Gateways client", err)
+
 		if gateways == nil {
 			message = "There are no gateways detected"
 			status.QueueWarningMessage(message)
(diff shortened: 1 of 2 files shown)


maayanf24:
This PR depends on https://github.com/submariner-io/submariner-operator/pull/1139

dfarrell07 (approved):
Looks great, thanks @maayanf24!

aswinsuryan (approved):

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,83 @@
...
+	exitOnError("Error getting REST config for cluster", err)
+
+	for _, item := range configs {
+		message := fmt.Sprintf("Validating Connections in %q", item.clusterName)
+		status.Start(message)
+		fmt.Println()
+		submariner := getSubmarinerResource(item.config)
+
+		if submariner == nil {
+			fmt.Printf("%v for cluster %q\n", submMissingMessage, item.clusterName)
```
sridhargaddam: Instead of fmt.Printf, shall we use the following
```
status.QueueWarningMessage(submMissingMessage)
status.End(cli.Success)
```

sridhargaddam:
Thanks @maayanf24. Its looking good, just a small comment.

sridhargaddam (approved):

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+func init() {
+	validateCmd.AddCommand(validateConnectionsCmd)
+}
+
+func validateConnections(cmd *cobra.Command, args []string) {
+	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
+	exitOnError("Error getting REST config for cluster", err)
+
+	for _, item := range configs {
+		message := fmt.Sprintf("Validating Connections in %q", item.clusterName)
```
tpantelis:
```suggestion
		message := fmt.Sprintf("Validating connections in cluster %q", item.clusterName)
```

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+
+		if submariner == nil {
+			status.QueueWarningMessage(submMissingMessage)
+			status.End(cli.Success)
+			continue
+		}
+
+		gateways := submariner.Status.Gateways
+		if gateways == nil {
+			message = "There are no gateways detected in cluster"
```
tpantelis: We print the cluster name in the opening message above.

```suggestion
			message = "There are no gateways detected"
```

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+			status.QueueFailureMessage(message)
+			status.End(cli.Failure)
```
tpantelis, line 58: Should this be a failure?

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+			continue
+		}
+
+		allConnectionsEstablished := true
+		for _, gateway := range *gateways {
+			for _, connection := range gateway.Connections {
+				if connection.Status == submv1.Connecting {
+					message = fmt.Sprintf("Connection to cluster %q is still trying to connect", connection.Endpoint.ClusterID)
+					status.QueueFailureMessage(message)
```
tpantelis, line 70: This doesn't seem like it should be a failure, ie it might connect - we just don't know. Perhaps warning?

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+			status.QueueFailureMessage(message)
+			status.End(cli.Failure)
+			continue
+		}
+
+		allConnectionsEstablished := true
+		for _, gateway := range *gateways {
+			for _, connection := range gateway.Connections {
+				if connection.Status == submv1.Connecting {
+					message = fmt.Sprintf("Connection to cluster %q is still trying to connect", connection.Endpoint.ClusterID)
```
tpantelis:
```suggestion
					message = fmt.Sprintf("Connection to cluster %q is in progress", connection.Endpoint.ClusterID)
```

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+		gateways := submariner.Status.Gateways
+		if gateways == nil {
+			message = "There are no gateways detected in cluster"
+			status.QueueFailureMessage(message)
+			status.End(cli.Failure)
+			continue
+		}
+
+		allConnectionsEstablished := true
+		for _, gateway := range *gateways {
```
tpantelis: I think we should make sure it's the/an active gateway (`HAStatus == HAStatusActive`).

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+				}
+			}
+		}
+
+		if !allConnectionsEstablished {
+			status.End(cli.Failure)
+			continue
+		}
+
+		message = "All Gateways connections are established"
```
tpantelis: L43 doesn't mention `gateway` so we should be consistent. Or mention `gateway` in both places but I don't think it's necessary.

```suggestion
		message = "All connections are established"
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_connections.go
mangelajo: It should not be a failure, may be just a warning. Since we are supporting discovery without connectivity.

We could detect, if "ServiceDiscovery" is enabled alone, then Warning. If connectivity + service discovery enabled = Failure


Let's leave a warning for now, and in the future we can do something more advanced like what I was describing,

pkg/subctl/cmd/validate_connections.go
mangelajo: Correct. We could have passive gateways which have 0 connections.

pkg/subctl/cmd/validate_connections.go
mangelajo: It should not be a failure, may be just a warning. Since we are supporting discovery without connectivity.

We could detect, if "ServiceDiscovery" is enabled alone, then Warning. If connectivity + service discovery enabled = Failure


Let's leave a warning for now, and in the future we can do something more advanced like what I was describing,

pkg/subctl/cmd/validate_connections.go
mangelajo: In this case, I think it's more like an error, your cluster connectivity really doesn't work until that's ready.

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,86 @@
...
+
+func validateConnections(cmd *cobra.Command, args []string) {
+	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
+	exitOnError("Error getting REST config for cluster", err)
+
+	for _, item := range configs {
+		message := fmt.Sprintf("Validating Connections in %q", item.clusterName)
+		status.Start(message)
+		fmt.Println()
+		submariner := getSubmarinerResource(item.config)
```
mangelajo, line 46: One thing that worries me about this is that if the submariner-operator is failing or stopped, the gateways will not be up-to date, and the data we provide will be out-dated or erroneous.

I still think we should stop grabbing the gateways from the submariner resources, and instead by Gateways.List()
 
@skitt  ^ 

Sorry to keep insisting in this, I'm known to be stubborn sometimes :-P , but I have faced many times where te operator is failing and it seems like it's not connected, but it is... or it seems like it's disconnected but it's not, we will have more complex escalations because of this.

mangelajo (changes requested):

pkg/subctl/cmd/validate_connections.go
skitt: I agree, same as for metrics it seems preferable to go to the source (and I’ve probably changed my mind on this, ISTR advocating for going through the `Submariner` resource in the past).

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,88 @@
...
+		if gateways == nil {
+			message = "There are no gateways detected"
+			status.QueueWarningMessage(message)
+			status.End(cli.Failure)
+			continue
+		}
+
+		allConnectionsEstablished := true
+		for _, gateway := range *gateways {
+			if gateway.HAStatus == submv1.HAStatusActive {
```
tpantelis: minor thing but we can save some indentation:

```suggestion
			if gateway.HAStatus != submv1.HAStatusActive {
			    continue
			}
			...
```

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,98 @@
...
+	submarinerv1 "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+
+	v1opts "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var validateConnectionsCmd = &cobra.Command{
+	Use:   "connections",
+	Short: "Validate the gateways connections",
+	Long: "This command validates if the connections of the gateways are all " +
+		"established",
```
tpantelis:
```suggestion
	Short: "Validate the gateway connections",
	Long: "This command checks that the gateway connections are all established",
```

pkg/subctl/cmd/validate_connections.go
```diff
@@ -0,0 +1,98 @@
...
+			status.End(cli.Success)
+			continue
+		}
+
+		submarinerClient, err := submarinerv1.NewForConfig(item.config)
+		exitOnError("Unable to get the Submariner client", err)
+
+		gateways, err := submarinerClient.SubmarinerV1().Gateways(OperatorNamespace).
+			List(v1opts.ListOptions{})
+		exitOnError("Unable to get the Gateways client", err)
```
tpantelis: Since we may be processing multiple clusters, perhaps we should `QueueFailureMessage` for these cases instead of failing fast.

tpantelis (changes requested):

tpantelis (approved):

mangelajo (approved):

Issue #1143, opened 2021-03-09
Capture logs from submariner-related pods


Jaanki:
**What would you like to be added**:
Log collection mechanism for below pods

-  submariner-operator pods
- all submariner/connectivity pods
- all service discovery pods

**Why is this needed**:
Sub task of Epic #1125

Jaanki:
https://github.com/submariner-io/submariner-operator/pull/1167

PR #1167, opened 2021-03-16
Collect submariner pod logs

Jaanki:
Collect the logs for pods related to submariner
and write those to a file at a location
[pod-type]/logs/[podname].log


pkg/discovery/network/pods.go
```diff
@@ -49,10 +50,10 @@ func findPodCommandParameter(clientSet kubernetes.Interface, labelSelector, para
 	return "", nil
 }
 
-func findPod(clientSet kubernetes.Interface, labelSelector string) (*v1.Pod, error) {
+func FindPod(clientSet kubernetes.Interface, labelSelector string) (*v1.PodList, error) {
 	pods, err := clientSet.CoreV1().Pods("").List(v1meta.ListOptions{
 		LabelSelector: labelSelector,
-		Limit:         1,
+		//Limit:         1,
```
Jaanki: @mangelajo Are these changes and the relevant changes throughout the code ok to you or do they break any of your logic?

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,63 @@
...
+	"path/filepath"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
+)
+
+func getGatewayPodsDetails(clientSet kubernetes.Interface) {
```
Jaanki: The same logic will be replicated for lighthouse, routeagent and operator pods once this is finalised.

Jaanki: This will be invoked from subctl gather command that @vthapar is working on PR #1159

pkg/discovery/network/pods.go
Jaanki: I can change the func name to FindAllPods or FindPods

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,63 @@
+package cmd
```
vthapar: Create a folder `subctl/cmd/gather` and rename this file to `connectivity.go` Change pkg name to `gather`. We'll have one file for each of the modules in this folder/pkg.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if err != nil {
+		fmt.Println("error getting pods")
+	}
+	for _, pod := range pods.Items {
+		getPodLogs(clientSet, pod.Name)
+	}
+}
+
+func getPodLogs(clientset kubernetes.Interface, podname string) {
```
vthapar: Make this return error. Take a look at commit 2 on #1159 to get an idea.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	defer logstream.Close()
+
+	logs := new(bytes.Buffer)
+	_, err = io.Copy(logs, logstream)
+	if err != nil {
+		fmt.Errorf("error in copy information from logs to buf %s", err)
+	}
+
+	dirname := filepath.Join(strings.Split(podname, "-")[1], "logs")
```
vthapar: Do we need multiple directories? Why not just name files as `connectivity_<podname>.log`

vthapar (changes requested):

pkg/subctl/cmd/logs.go
Jaanki: my idea to structure it as
```
lighthouse/logs/pod1.log
lighthouse/logs/pod2.log
routeagent/logs/pod1.log
```
and so on for other pods
and similarly for other details
`lighthouse/details/pod1.txt`
and so on

pkg/discovery/network/pods.go
tpantelis: I suggest `FindPods`.

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -34,8 +34,9 @@ const (
 )
 
 func discoverOvnKubernetesNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, error) {
-	ovnDBPod, err := findPod(clientSet, "name=ovnkube-db")
+	ovnDBPods, err := FindPod(clientSet, "name=ovnkube-db")
 
+	ovnDBPod := &ovnDBPods.Items[0]
 	if err != nil || ovnDBPod == nil {
```
tpantelis: This is will panic if no pod is found, ie list is empty.

```suggestion
	if err != nil || len(ovnDBPods.Items) == 0 {
```

pkg/discovery/network/pods.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 func findPodCommandParameter(clientSet kubernetes.Interface, labelSelector, parameter string) (string, error) {
-	pod, err := findPod(clientSet, labelSelector)
+	pods, err := FindPod(clientSet, labelSelector)
 
+	pod := &pods.Items[0]
```
tpantelis: ...same here

pkg/discovery/network/weavenet.go
```diff
@@ -21,8 +21,9 @@ import (
 )
 
 func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, error) {
-	weaveNetPod, err := findPod(clientSet, "name=weave-net")
+	weaveNetPods, err := FindPod(clientSet, "name=weave-net")
 
+	weaveNetPod := &weaveNetPods.Items[0]
```
tpantelis: ...and here

tpantelis (changes requested):

pkg/discovery/network/pods.go
```diff
@@ -50,18 +50,26 @@ func findPodCommandParameter(clientSet kubernetes.Interface, labelSelector, para
...
-		LabelSelector: labelSelector,
-		Limit:         1,
-	})
+	listOptions := v1meta.ListOptions{
+		LabelSelector : labelSelector,
+		Limit         : 1,
+	}
+	pods, err := FindPods(clientSet, listOptions)
+
+    return &pods.Items[0], err
```
tpantelis: This would panic if no pods found.

pkg/discovery/network/pods.go
```diff
@@ -50,18 +50,26 @@ func findPodCommandParameter(clientSet kubernetes.Interface, labelSelector, para
...
+		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", options.LabelSelector)
 	}
 
 	if len(pods.Items) == 0 {
 		return nil, nil
 	}
 
-	return &pods.Items[0], nil
-}
+	return pods, nil
```
tpantelis: This really can be reduced to:

```
pods, err := clientSet.CoreV1().Pods("").List(options)
return pods, errors.WithMessagef(err, "error listing Pods by label selector %q", options.LabelSelector)
```

As such, this function doesn't really add any value as users can just call `clientSet.CoreV1().Pods("").List(options)` themselves. So I suggest to do the latter.

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +84,30 @@ func gatherData() {
...
-func gatherConnectivity(status *cli.Status, dataType string) error {
+func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string) error {
 	switch dataType {
 	case Logs:
-		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+         logscollected, err := getConnectivityPodsLogs(clientSet)
+         if logscollected {
+         	status.QueueSuccessMessage("Successfully gathered connectivity logs")
+         } else {
+         	status.QueueFailureMessage(fmt.Sprintf("error gathering connectivity logs", err))
```
```suggestion
         	status.QueueFailureMessage(fmt.Sprintf("Error gathering connectivity logs", err))
```

pkg/subctl/cmd/logs.go
vthapar: I'd prefer to avoid having nested directories. I was thinking something on lines of 

```
gather-<cluster-id>-<timestamp>/
  connectivity-logs-<podname1>.log
  connectivity-logs-<podname2>.log
  service-discovery-logs-<podname1>.log
  broker-resources-gateways.json
```
Avoids unnecessary nested directories and one look at file name you know what the contents are. Instead of submariner or lighthouse, use module names as used in gather command [connectivity, service-discovery etc.]

@mangelajo @skitt @tpantelis @dfarrell07 Appreciate your inputs on this. Let us use this PR to decide on what directory layout and filenames should look like.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
+)
+
+func getConnectivityPodsLogs(clientSet kubernetes.Interface) (bool, error){
```
tpantelis: I don't see the purpose of returning a `bool` along with `error`. It seems the `bool` is to indicate the logs were successfully collected but a nil `error` also indicates that. So I'd just return `error`.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	pods, err := network.FindPods(clientSet, v1.ListOptions{LabelSelector: "app=submariner-gateway"})
+
+	if err != nil {
+		fmt.Println("error getting pods")
```
tpantelis: Remove

```suggestion
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, pod := range pods.Items {
+		_, err := getPodLogs(clientSet, pod.Name)
+		if err != nil {
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+func getPodLogs(clientset kubernetes.Interface, podname string) (bool, error){
```
tpantelis: Same here re: `bool` return.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	podLogOpts := corev1.PodLogOptions{}
+	logrequest := clientset.CoreV1().Pods(SubmarinerNamespace).GetLogs(podname, &podLogOpts)
+	logstream, err := logrequest.Stream()
+	if err != nil {
+		fmt.Errorf("error opening log stream %s", err)
+		return false, err
```
tpantelis:
```suggestion
		return errors.WithMessagef(err, "error opening log stream for pod %q", podname)
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return true, nil
+}
+
+func convertStreamToString(logstream io.ReadCloser) (string, error) {
+	logs := new(bytes.Buffer)
+	_, err := io.Copy(logs, logstream)
+	if err != nil {
+		fmt.Errorf("error in copy information from logs to buf %s", err)
```
tpantelis: We return the err so no need to log it.

```suggestion
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	podLogOpts := corev1.PodLogOptions{}
+	logrequest := clientset.CoreV1().Pods(SubmarinerNamespace).GetLogs(podname, &podLogOpts)
+	logstream, err := logrequest.Stream()
+	if err != nil {
+		fmt.Errorf("error opening log stream %s", err)
+		return false, err
+	}
+	defer logstream.Close()
+
+	logs, err := convertStreamToString(logstream)
```
tpantelis:
```suggestion
	logs, err := convertStreamToString(logstream)
	if err != nil {
		return errors.WithMessagef(err, "error processing log stream for pod %q", podname)
	}
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return logs.String(), nil
+}
+
+func writeToFile(podname string, data string) error {
+	dirname := filepath.Join(strings.Split(podname, "-")[1], "logs")
+	if _, err := os.Stat(dirname); os.IsNotExist(err) {
+		os.MkdirAll(dirname, 0777)
+	}
+	extension := ".logs"
```
tpantelis:
```suggestion
	extension := ".log"
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	logFile := filepath.Join(dirname, podname)
+
+	f, err := os.OpenFile(logFile+extension, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0777)
+	if err != nil {
+		fmt.Println("error opening file", err)
```
tpantelis:
```suggestion
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+		return err
+	}
+	defer f.Close()
+	_, err = f.WriteString(data)
+	if err != nil {
+		fmt.Errorf("error writing to file", err)
```
tpantelis:
```suggestion
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		fmt.Errorf("error opening log stream %s", err)
+		return false, err
+	}
+	defer logstream.Close()
+
+	logs, err := convertStreamToString(logstream)
+	err = writeToFile(podname, logs)
+	if err != nil {
+		return false, err
```
tpantelis:
```suggestion
		return errors.WithMessagef(err, "error persisting log stream for pod %q", podname)
```

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
+)
+
+func getConnectivityPodsLogs(clientSet kubernetes.Interface) (bool, error){
+	pods, err := network.FindPods(clientSet, v1.ListOptions{LabelSelector: "app=submariner-gateway"})
```
tpantelis: Don't we also want to collect the route-agent logs? There's also globalnet and the network plugin syncer.

pkg/subctl/cmd/logs.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if err != nil {
+		fmt.Println("error getting pods")
+		return false, err
+	}
+	for _, pod := range pods.Items {
+		_, err := getPodLogs(clientSet, pod.Name)
```
tpantelis: I think it would be useful to `QueueSuccessMessage`/`QueueFailureMessage` for each pod. For success, include the pod name and output file name, eg "_Log for pod %q successfully persisted to file %q_".

pkg/subctl/cmd/logs.go
tpantelis: yeah I agree it would be better to put all the logs in one directory. The main reason for collecting the logs would likely be to tar them up and send to support so one directory would make that easier. Also for troubleshooting we'd want all the logs atomically captured and stored together. We also may want to allow the user to specify the dir or we generate one with a timestamp, ie `submariner-<timestamp>`.  For the file names, I lean towards just the pod names:

```
submariner-gateway-<cluster-id>.log
submariner-route-agent-<cluster-id>.log
...
```


The module name instructs subctl which pods to collect but I don't think we need to reflect that in the file name.

tpantelis (changes requested):

pkg/subctl/cmd/logs.go
Jaanki: THe logic would be same for those pods. I will add those pods once this is confirmed as the logic will get repeated

pkg/subctl/cmd/logs.go
Jaanki: How about this
`logs/<cluster-id>-<podname>.log`

tpantelis: yeah that file name pattern sounds fine. But as described above, we should timestamp the directory - I suggest `submariner-<timestamp>`.

dfarrell07 (approved):

pkg/subctl/cmd/logs.go
vthapar: > How about this
> `logs/<cluster-id>-<podname>.log`

Don't create separate directory for logs and other resources. We'll be collecting resources in yaml or json format, so log files can be distinguished by `*.log`

Jaanki: With the latest push when log collection for all modules is implemented, running `subctl gather` on all clusters might create multiple dirs with different timestamp becasue there is a possibility that the time would have changed mid way and hence some of the logs would be written to new dir.

Here's the proof while running it locally
I ran `subctl gather` and had 2 dirs created

`submariner-20210323231719
submariner-20210323231720`

because the `second` changed.

cluster1 logs got divided in between these 2 dirs

```
ls submariner-20210323231719/
**cluster1-submariner-gateway-qnspm.log**
cluster2-submariner-routeagent-jzrmg.log 
cluster3-submariner-routeagent-8l5sx.log
cluster2-submariner-gateway-jn79f.log     
cluster2-submariner-routeagent-twnmm.log  
cluster3-submariner-routeagent-kl9rr.log
cluster2-submariner-routeagent-552mm.log  
cluster3-submariner-gateway-2mz29.log     
cluster3-submariner-routeagent-mgq9v.log

ls submariner-20210323231720
**cluster1-submariner-routeagent-mtf5h.log  
cluster1-submariner-routeagent-p8fqf.log** 
```

Jaanki: The output would be like this

```
$ ./../submariner-operator/bin/subctl gather
Gathering information from cluster2
 ✓ Gathering operator resources
 ⚠ Gather Operator Resources not implemented yet
 ✓ Gathering operator logs
 ✗ Gathering Operator logs not successful error getting logs for pod "submariner-lighthouse-agent-d6bcd4855-gftvq": error opening log stream for pod "submariner-lighthouse-agent-d6bcd4855-gftvq": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-d6bcd4855-gftvq" is waiting to start: image can't be pulled
 ✓ Gathering connectivity logs 
 ✓ Successfully collected Gateway pods logs
 ✓ Successfully collected Routeagent pods logs
 ✓ Gathering connectivity resources
 ⚠ Gather Connectivity Resources not implemented yet
 ✓ Gathering service-discovery logs
 ✗ Gathering Service Discovery logs not successful error getting logs for pod "submariner-lighthouse-agent-d6bcd4855-gftvq": error opening log stream for pod "submariner-lighthouse-agent-d6bcd4855-gftvq": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-d6bcd4855-gftvq" is waiting to start: image can't be pulled
 ✓ Gathering service-discovery resources
 ⚠ Gather ServiceDiscovery Resources not implemented yet
 ✓ Gathering broker logs
 ⚠ No logs to gather on Broker
 ✓ Gathering broker resources
 ⚠ Gather Broker Resources not implemented yet

Gathering information from cluster3
 ✓ Gathering connectivity logs 
 ✓ Successfully collected Gateway pods logs
 ✓ Successfully collected Routeagent pods logs
 ✓ Gathering connectivity resources
 ⚠ Gather Connectivity Resources not implemented yet
 ✓ Gathering service-discovery logs
 ✗ Gathering Service Discovery logs not successful error getting logs for pod "submariner-lighthouse-agent-5ddc466879-flqdn": error opening log stream for pod "submariner-lighthouse-agent-5ddc466879-flqdn": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-5ddc466879-flqdn" is waiting to start: trying and failing to pull image
 ✓ Gathering service-discovery resources
 ⚠ Gather ServiceDiscovery Resources not implemented yet
 ✓ Gathering broker logs
 ⚠ No logs to gather on Broker
 ✓ Gathering broker resources
 ⚠ Gather Broker Resources not implemented yet
 ✓ Gathering operator logs
 ✗ Gathering Operator logs not successful error getting logs for pod "submariner-lighthouse-agent-5ddc466879-flqdn": error opening log stream for pod "submariner-lighthouse-agent-5ddc466879-flqdn": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-5ddc466879-flqdn" is waiting to start: trying and failing to pull image
 ✓ Gathering operator resources
 ⚠ Gather Operator Resources not implemented yet

Gathering information from cluster1
 ✓ Gathering connectivity logs 
 ✓ Successfully collected Gateway pods logs
 ✓ Successfully collected Routeagent pods logs
 ✓ Gathering connectivity resources
 ⚠ Gather Connectivity Resources not implemented yet
 ✓ Gathering service-discovery logs
 ✗ Gathering Service Discovery logs not successful error getting logs for pod "submariner-lighthouse-agent-649c9965db-bvc4m": error opening log stream for pod "submariner-lighthouse-agent-649c9965db-bvc4m": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-649c9965db-bvc4m" is waiting to start: trying and failing to pull image
 ✓ Gathering service-discovery resources
 ⚠ Gather ServiceDiscovery Resources not implemented yet
 ✓ Gathering broker logs
 ⚠ No logs to gather on Broker
 ✓ Gathering broker resources
 ⚠ Gather Broker Resources not implemented yet
 ✓ Gathering operator logs
 ✗ Gathering Operator logs not successful error getting logs for pod "submariner-lighthouse-agent-649c9965db-bvc4m": error opening log stream for pod "submariner-lighthouse-agent-649c9965db-bvc4m": container "submariner-lighthouse-agent" in pod "submariner-lighthouse-agent-649c9965db-bvc4m" is waiting to start: trying and failing to pull image
 ✓ Gathering operator resources
 ⚠ Gather Operator Resources not implemented yet
```

tpantelis: The idea is that you put all files from all modules from all clusters under the same timestamped dir which you would create at the beginning of the subclt cmd.

pkg/discovery/network/pods.go
```diff
@@ -50,18 +50,29 @@ func findPodCommandParameter(clientSet kubernetes.Interface, labelSelector, para
...
-		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", labelSelector)
+		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", options.LabelSelector)
 	}
 
 	if len(pods.Items) == 0 {
 		return nil, nil
 	}
 
-	return &pods.Items[0], nil
+	return pods, nil
```
tpantelis: I mentioned this in a previous comment: 

This really can be reduced to:

```
pods, err := clientSet.CoreV1().Pods("").List(options)
return pods, errors.WithMessagef(err, "error listing Pods by label selector %q", options.LabelSelector)
```

As such, this function doesn't really add any value as users can just call `clientSet.CoreV1().Pods("").List(options)` themselves. So I suggest to remove this function and change `findPod` back to the way it was..

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
-					status := cli.NewStatus()
-					status.Start(fmt.Sprintf("Gathering %s %s...", module, dataType))
-					status.End(cli.CheckForError(gatherFuncs[module](status, dataType)))
+	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
+	exitOnError("Error getting REST config", err)
+	for _, item := range configs {
+		clientSet, err := kubernetes.NewForConfig(item.config)
+		exitOnError("error getting clientset", err)
+		clustername := item.clusterName
+		fmt.Println("Gathering information from", clustername)
```
tpantelis: Since we're using a `Status` for output, we shouldn't use `Println`. We can remove this line since we start a new `Status` for each module/type below.

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
+				for dataType, ok := range gatherTypeFlags {
+					if ok {
+						status := cli.NewStatus()
+						status.Start(fmt.Sprintf("Gathering %s %s", module, dataType))
+						status.End(cli.CheckForError(gatherFuncs[module](status, clientSet, dataType, clustername)))
+					}
 				}
 			}
 		}
+		fmt.Println("")
```
```suggestion
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
+		clientSet, err := kubernetes.NewForConfig(item.config)
+		exitOnError("error getting clientset", err)
+		clustername := item.clusterName
+		fmt.Println("Gathering information from", clustername)
+		for module, ok := range gatherModuleFlags {
+			if ok {
+				for dataType, ok := range gatherTypeFlags {
+					if ok {
+						status := cli.NewStatus()
+						status.Start(fmt.Sprintf("Gathering %s %s", module, dataType))
```
```suggestion
						status.Start(fmt.Sprintf("Gathering %s %s from cluster %q", module, dataType, item.clusterName))
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
-				if ok {
-					status := cli.NewStatus()
-					status.Start(fmt.Sprintf("Gathering %s %s...", module, dataType))
-					status.End(cli.CheckForError(gatherFuncs[module](status, dataType)))
+	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
+	exitOnError("Error getting REST config", err)
+	for _, item := range configs {
+		clientSet, err := kubernetes.NewForConfig(item.config)
+		exitOnError("error getting clientset", err)
+		clustername := item.clusterName
```
tpantelis: Just use `item.clusterName`

```suggestion
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
+		fmt.Println("")
 	}
 }
 
-func gatherConnectivity(status *cli.Status, dataType string) error {
+func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string, clustername string) error {
 	switch dataType {
 	case Logs:
-		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+		err := gather.GetGatewayPodsLogs(clientSet, clustername)
```
tpantelis: `PodsLogs` and "_pods logs_" sound awkward. Also we use the term _gather_ elsewhere so let's be consistent here. This applies to all other `Get*PodsLogs` functions.

```suggestion
		err := gather.GatherGatewayPodLogs(clientSet, clustername)
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
+					}
 				}
 			}
 		}
+		fmt.Println("")
 	}
 }
 
-func gatherConnectivity(status *cli.Status, dataType string) error {
+func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string, clustername string) error {
```
tpantelis: I suggest to use a struct called `gatherInfo` to contain the params. I suspect we may need to pass more params.

```suggestion
func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string, clusterName string) error {
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
 	switch dataType {
 	case Logs:
-		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+		err := gather.GetGatewayPodsLogs(clientSet, clustername)
+		if err != nil {
+			status.QueueFailureMessage(fmt.Sprintf("Gathering Gateway pods logs not successful %s", err))
+		} else {
+			status.QueueSuccessMessage("Successfully collected Gateway pods logs")
```
tpantelis: This applies to all other occurrences of "_pods logs_" as well. Also let's be consistent and use the term _gather_ everywhere. 

```suggestion
			status.QueueSuccessMessage("Successfully gathered Gateway pod logs")
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
 }
 
-func gatherConnectivity(status *cli.Status, dataType string) error {
+func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string, clustername string) error {
 	switch dataType {
 	case Logs:
-		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+		err := gather.GetGatewayPodsLogs(clientSet, clustername)
+		if err != nil {
+			status.QueueFailureMessage(fmt.Sprintf("Gathering Gateway pods logs not successful %s", err))
```
tpantelis: Similar elsewhere...

```suggestion
			status.QueueFailureMessage(fmt.Sprintf("Failed to gather Gateway pod logs: %s", err))
```

pkg/subctl/cmd/gather.go
```diff
@@ -119,10 +147,10 @@ func gatherDiscovery(status *cli.Status, dataType string) error {
 	return nil
 }
 
-func gatherBroker(status *cli.Status, dataType string) error {
+func gatherBroker(status *cli.Status, clientSet kubernetes.Interface, dataType string, clustername string) error {
 	switch dataType {
 	case Logs:
-		status.QueueSuccessMessage("No logs to gather on Broker")
+		status.QueueWarningMessage("No logs to gather on Broker")
```
tpantelis: This doesn't warrant a warning. In fact we should really skip this altogether. I suggest to have the gather functions return a `*cli.Result` instead of `error`. The caller on L101 above only uses the returned error to select `cli.Success` or `cli.Failure` anyway. If the gather function returns nil then don't call `status.End` - this will result in no output.

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
+)
+
+const (
+	gatewayPods = "app=submariner-gateway"
+	routeagentPods = "app=submariner-routeagent"
+)
+
+func GetGatewayPodsLogs(clientSet kubernetes.Interface, clustername string) error {
```
tpantelis: Same in other functions as well. In general, please follow the camel-case naming convention when a var has multiple words.  

```suggestion
func GetGatewayPodsLogs(clientSet kubernetes.Interface, clusterName string) error {
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+		err := getPodLogs(clientSet, pod.Name, clustername)
+		if err != nil {
+			return errors.WithMessagef(err, "error getting logs for pod %q", pod.Name)
+		}
+	}
+	return nil
+}
+
+func getPodLogs(clientset kubernetes.Interface, podname string, clustername string) error {
+	logrequest := clientset.CoreV1().Pods("submariner-operator").GetLogs(podname, &corev1.PodLogOptions{})
```
tpantelis: Let's avoid hard-coding the namespace. We can get it from the `Pod` instance.

```suggestion
func getPodLogs(clientset kubernetes.Interface, pod *corev1.Pod, clustername string) error {
	logrequest := clientset.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{})
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func processLogStream(logrequest *rest.Request, podname string) (string, error) {
+	logstream, err := logrequest.Stream()
+	if err != nil {
+		return "", errors.WithMessagef(err, "error opening log stream for pod %q", podname)
```
tpantelis: The caller outputs the pod name.

```suggestion
		return "", errors.WithMessage(err, "error opening log stream")
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	defer logstream.Close()
+
+	logs := new(bytes.Buffer)
+	_, err = io.Copy(logs, logstream)
+	if err != nil {
+		return "", errors.WithMessagef(err, "error converting log stream to string for pod %q", podname)
```
tpantelis:
```suggestion
		return "", errors.WithMessage(err, "error copying the log stream")
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+		err := os.MkdirAll(dirname, 0777)
+		if err != nil {
+			return errors.WithMessagef(err, "error creating directory %s", dirname)
+		}
+	}
+
+	filename := filepath.Join(dirname, clustername + "-" + podname + ".log")
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0777)
+	if err != nil {
+		return errors.WithMessagef(err, "error opening file %s for pod %s", filename, podname)
```
tpantelis:
```suggestion
		return errors.WithMessagef(err, "error opening file %s", filename)
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	defer f.Close()
+	_, err = f.WriteString(data)
+	if err != nil {
+		return errors.WithMessagef(err, "error writing to file %s for pod %s", filename, podname)
```
tpantelis:
```suggestion
		return errors.WithMessagef(err, "error writing to file %s", filename)
```

tpantelis (changes requested):

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +85,43 @@ func gatherData() {
...
-	for module, ok := range gatherModuleFlags {
-		if ok {
-			for dataType, ok := range gatherTypeFlags {
-				if ok {
-					status := cli.NewStatus()
-					status.Start(fmt.Sprintf("Gathering %s %s...", module, dataType))
-					status.End(cli.CheckForError(gatherFuncs[module](status, dataType)))
+	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
+	exitOnError("Error getting REST config", err)
+	for _, item := range configs {
```
vthapar: Please don't overload this patch with code for gathering from multiple clusters. That change would require its own discussion that is not within scope of this change. In this PR focus should be on gathering submariner logs, filenames, directory structures etc.

vthapar (changes requested):

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
+)
+
+const (
+	gatewayPods = "app=submariner-gateway"
```
vthapar:
```suggestion
	gatewayPodLabel = "app=submariner-gateway"
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	routeagentPods = "app=submariner-routeagent"
```
vthapar:
```suggestion
	routeagentPodLabel = "app=submariner-routeagent"
```

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
+/*
```
vthapar, line 1: @skitt Does `subctl/cmd/gather/*` make sense or you think it should be `subctl/gather/*`

vthapar (changes requested):

pkg/subctl/cmd/gather.go
Jaanki: DOne

pkg/subctl/cmd/gather/connectivity.go
mangelajo: subctl/cmd/gather sounds fine to me

pkg/subctl/cmd/gather.go
mangelajo: We have the operator logs on broker, right?

pkg/subctl/cmd/gather/connectivity.go
skitt: I vote for `subctl/cmd/gather` too.

pkg/discovery/network/pods.go
mangelajo: It's not exactly the same, pods wouldn't be nil on return then I think.

While on this case the function returns nil,nil

@tpantelis I think that in this case is better not to change the contract of the function since it's used in other places. (network discovery for example)

skitt: @mangelajo as I understand it, you’re both asking for the same thing — to leave `findPod` as-is. `FindPods` can’t be used anywhere else because it’s new to this PR; there might be other similar functions elsewhere.

pkg/subctl/cmd/gather.go
skitt: Yes, we do now.

pkg/discovery/network/pods.go
tpantelis: yes leave `findPod` as-is. I also assert we don't even need `FindPods` - it's basically a thin wrapper for `clientSet.CoreV1().Pods("").List(options)` so I don't think adds any value,

pkg/subctl/cmd/gather.go
tpantelis: OK. I still think it makes more sense to return `*cli.Result` as I outlined above.

pkg/discovery/network/pods.go
Jaanki: FindPods lists ALL pods in the given ns. findPod lists just 1 pod `(Limit: 1)`.
pods in FindPods will be nil when there are no pods matching a label.

The idea to add FindPods is to be able to list pods with varying ListOptions whenever needed.
Pods can still be nil when there are no pods matching the mentioned label.

Jaanki: Adjusting findPod to my requirements will need changes in all findPod (removing the `Limit: 1` and returning the whole list and not just pods.Item[0] which in turn will need changes in all the places that calls findPod.

tpantelis: My point is that the caller can just call `clientSet.CoreV1().Pods("").List(options)`. `FindPods` adds no value on top of that.  And we wouldn't want to return nil as `List` does not (unless maybe if there's an err returned).

pkg/subctl/cmd/gather.go
Jaanki: Println was to print an empty line in between each cluster output. Like this:
```
Gathering information from cluster1
Logs collected
                                                   ------  added by Println
Gathering information from cluster2
Logs collected
```

But since, I am removing logic to collect logs from all clusters but just 1, this is removed.

tpantelis: Either way we should use the `Status` for all output

pkg/discovery/network/pods.go
tpantelis: > Adjusting findPod to my requirements will need changes in all findPod (removing the `Limit: 1` and returning the whole list and not just pods.Item[0] which in turn will need changes in all the places that calls findPod.

right that's why we're saying to leave `findPod` as is.

pkg/subctl/cmd/gather.go
mangelajo: I think it's fine to leave, see at the example output where there is a prefix "Gathering information from ..." before all the tick statuses. So there's a separation between the cluster items.

pkg/subctl/cmd/gather.go
mangelajo: I prefer the separation between the clusters, it's easier to read. But anyway this is just opinion. We can probably merge, see how it looks and we can shape the aesthetics/UI of it later :)

pkg/subctl/cmd/gather.go
mangelajo: Can we leave that to later improvement? I agree though.

pkg/subctl/cmd/gather.go
mangelajo: We need to gather the operator logs on the broker, and if that's handled somewhere else, then no warning.

mangelajo:
```suggestion
```

pkg/subctl/cmd/gather/connectivity.go
mangelajo: I see it marked as resolved but not changed.

mangelajo (changes requested):

pkg/subctl/cmd/gather.go
tpantelis: yeah good point.

pkg/subctl/cmd/gather/connectivity.go
Jaanki: all submariner related pods are in submariner-operator ns. There are multiple instances in the code where the ns is hard-coded as a constant. I will do the same.

tpantelis: yes it is so you should at least use the constant but we shouldn't perpetuate that especially since we have the actual namespace in hand and don't need to hard-code. I believe we have a future use case to customize it.

pkg/subctl/cmd/gather/connectivity.go
tpantelis: Re: the function name, the package name is `gather` so we can just name the function `GatewayPodLogs` as it will be referenced as `gather.GatewayPodLogs`. Same with the other functions

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -0,0 +1,120 @@
...
+	dirname := "submariner-" + time.Now().Format("20060102150405") //submariner-YYYYMMDDHHMMSS
+	if _, err := os.Stat(dirname); os.IsNotExist(err) {
+		err := os.MkdirAll(dirname, 0777)
+		if err != nil {
+			return errors.WithMessagef(err, "error creating directory %s", dirname)
+		}
+	}
+
+	filename := filepath.Join(dirname, clustername + "-" + podname + ".log")
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0777)
```
tpantelis:
```suggestion
	f, err := os.Create(filename)
```

pkg/discovery/network/pods.go
Jaanki: Keeping findpod as is and removing FindPods

pkg/subctl/cmd/gather.go
Jaanki: Will address this while adding log collection from Broker logic

pkg/subctl/cmd/gather/connectivity.go
Jaanki: I mark comments resolved when I have resolved them locally but not necessarily have pushed the changes.

pkg/subctl/cmd/gather/connectivity.go
Jaanki: Create truncates the data while OpenFile will append the data to the existing file.

pkg/subctl/cmd/gather.go
Jaanki: Yes. I will add that logic once this PR is finalised as it would be just be copy pasting.

pkg/subctl/cmd/gather/utils.go
```diff
@@ -0,0 +1,80 @@
...
+
+func findPods(clientSet kubernetes.Interface, options v1meta.ListOptions) (*corev1.PodList, error) {
+	pods, err := clientSet.CoreV1().Pods("").List(options)
+
+	if err != nil {
+		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", options.LabelSelector)
+	}
+
+	if len(pods.Items) == 0 {
+		return nil, errors.New("No pods found matching the Labelselector")
```
Jaanki: This gives below output when getting gateway and routeagent pods on Broker
```

✓ Gathering connectivity logs
 ✗ Failed to gather Gateway pod logs: No pods found matching the Labelselector
 ✗ Failed to gather Routeagent pod logs: No pods found matching the Labelselector
```

pkg/subctl/cmd/gather/connectivity.go
tpantelis: right we shouldn't want to append although it shouldn't already exist unless we allowed the dir to be specified.  Also I think we want mode 0700 (read, write, & execute only for owner) not 0777 as the latter gives read/write/execute access to everyone.

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +87,57 @@ func gatherData() {
...
+	exitOnError("error getting clientset", err)
+
+	rawconfig, err := config.RawConfig()
+	exitOnError("error getting Raw config", err)
+
+	clustername := getClusterNameFromContext(rawconfig, "")
+
+	dirname := "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
+	if _, err := os.Stat(dirname); os.IsNotExist(err) {
+		err := os.MkdirAll(dirname, 0777)
```
```suggestion
		err := os.MkdirAll(dirname, 0700)
```
```suggestion
		err := os.MkdirAll(dirname, 0777)
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +87,57 @@ func gatherData() {
...
+		if err != nil {
+			exitOnError(fmt.Sprintf("error creating directory %s", dirname), err)
```
```suggestion
			exitOnError(fmt.Sprintf("Error creating directory %s", dirname), err)
```
```suggestion
			exitOnError(fmt.Sprintf("error creating directory %s", dirname), err)
```

pkg/subctl/cmd/gather.go
```diff
@@ -82,23 +87,57 @@ func gatherData() {
...
 }
 
-func gatherConnectivity(status *cli.Status, dataType string) error {
+func gatherConnectivity(status *cli.Status, clientSet kubernetes.Interface, dataType string, params gather.GatherParams) error {
 	switch dataType {
 	case Logs:
-		status.QueueWarningMessage("Gather Connectivity Logs not implemented yet")
+		err := gather.GatherGatewayPodLogs(clientSet, params)
+		if err != nil {
+			status.QueueFailureMessage(fmt.Sprintf("Failed to gather Gateway pod logs: %s", err))
```
tpantelis: This should fail the `Status` on error.

pkg/subctl/cmd/gather/utils.go
tpantelis: The Status should be marked as failure.

tpantelis (changes requested):
I pushed a new [commit](https://github.com/submariner-io/submariner-operator/pull/1167/commits/39d36545cf3de4878cbc11140ace7893b7628957) to address further comments and refactor it a bit to accommodate subsequent PRs for resource gathering.

tpantelis (approved):

vthapar (approved):

skitt (approved):

mangelajo (approved):

Jaanki:
https://github.com/submariner-io/submariner-operator/issues/1143

PR #1169, opened 2021-03-16
Enable PR dependency check jobs

Jaanki:
This PR adds a `Dependent issue workflow` that will check for dependencies in PR.

### DIFF
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -1,0 +1,36 @@
+---
+name: PR Dependencies
+
+on:
+  issues:
+    types:
+      - opened
+      - reopened
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+  pull_request:
+    types:
+      - opened
+      - reopened
+  schedule:
+    - cron: '0 0 * * *' # check daily
+
+jobs:
+  check:
+    name: Check Dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # The label to use to mark dependent issues
+          label: dependent
+
+          # Enable checking for dependencies in issues.
+          check_issues: on
+
+          # A comma-separated list of keywords to mark dependency.
+          keywords: depends on, Depends on


tpantelis:
What is this PR for?

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          # (Optional) The label to use to mark dependent issues
+          label: dependent
+
+          # (Optional) Enable checking for dependencies in issues. Enable by
+          # setting the value to "on". Default "off"
+          check_issues: on
+
+          # (Optional) A comma-separated list of keywords. Default
+          # "depends on, blocked by"
```
dfarrell07:
```suggestion
          # (Optional) A comma-separated list of keywords.
```

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: z0al/dependent-issues@v1
+        env:
+          # (Required) The token to use to make API calls to GitHub.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # (Optional) The label to use to mark dependent issues
+          label: dependent
+
+          # (Optional) Enable checking for dependencies in issues. Enable by
+          # setting the value to "on". Default "off"
```
dfarrell07:
```suggestion
          # Enable checking for dependencies in issues. 
```

Reading this and looking at the main docs, I'm still not sure what this does. 😅

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@v1
+        env:
+          # (Required) The token to use to make API calls to GitHub.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # (Optional) The label to use to mark dependent issues
```
dfarrell07:
```suggestion
          # The label to use to mark dependent issues
```

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+  schedule:
+    - cron: '0 * * * *' # check hourly
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: z0al/dependent-issues@v1
+        env:
+          # (Required) The token to use to make API calls to GitHub.
```
dfarrell07:
```suggestion
```

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+  schedule:
+    - cron: '0 * * * *' # check hourly
+
+jobs:
+  check:
```
dfarrell07, line 21: Can we give this a human-friendly name for the PR UI?

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: Dependent Issues
+
+on:
```
dfarrell07, line 4: This job is going to run *a lot*. Every time an issue or PR is changed at all, and every hour. Do we really think this ability to mark dependencies is important enough to justify starting a container and running these scans likely at least order of 10k, likely order of 100k, times per year?

dfarrell07 (changes requested):
The job is currently failing with "Error: Resource not accessible by integration".

.github/workflows/dependent-issues.yml
Jaanki: This job will run on those PRs that have a "Depends on/ depends on" in comment section and not on all PRs. It not only marks the dependencies, it lets you about the dependency so that a PR is not merged before its dependent PR is merged.

.github/workflows/dependent-issues.yml
mangelajo: example: https://github.com/submariner-io/submariner/blob/devel/.github/workflows/e2e.yml#L9

.github/workflows/dependent-issues.yml
mangelajo: It's probably very short lived, let's see how it goes.

mangelajo:
Let's update the commit to explain better what's being done, instead of "Depends on PR", may be

"Enable PR dependency check jobs"

.github/workflows/dependent-issues.yml
dfarrell07: It has to scan every issue and PR at least.

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+  schedule:
+    - cron: '0 * * * *' # check hourly
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
```
mangelajo, line 23:
```suggestion
    name: check-dependent-PRs
    runs-on: ubuntu-latest
```

.github/workflows/dependent-issues.yml
dfarrell07: I think Janki modified the Issue/PR event triggers to only run on open/re-open, which would reduce the number of runs dramatically. That seems like a decent compromise to me.

mangelajo: That would make sense, may be then also daily?

I guess the periodic is just a fall back to fix any lost event due to CI issues.

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+  pull_request:
+    types:
+      - opened
+      - reopened
+  schedule:
+    - cron: '0 * * * *' # check hourly
```
mangelajo:
```suggestion
    - cron: '0 0 * * *' # check daily
```

Based on @dfarrell07 comments  ^ but checking daily.

.github/workflows/dependent-issues.yml
mangelajo: @dfarrell07 see line 22 now, does that seem ok?

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  check:
+    name: check-dependent-PRs
```
dfarrell07: These names should be written for display on the GH UI, with spaces and such. The job name will frequently be listed under the workflow name, so they can build on each other.

One example:

https://github.com/submariner-io/submariner/actions

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,36 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # The label to use to mark dependent issues
+          label: dependent
+
+          # Enable checking for dependencies in issues.
+          check_issues: on
+
+          # A comma-separated list of keywords to mark dependency.
+          keywords: depends on, Depends on
```
dfarrell07, line 36: Will this work for something like:

> Depends on: `<PR>`

And will this not trigger for something like:

> Does this (sic) depends on `<Issue>`?

.github/workflows/dependent-issues.yml
dfarrell07: I think daily is good, especially while we're kinda piloting this.

Jaanki: Yes, I modified the issue/PR trigger on open and re-open. But I feel daily is still too long a time gap. However, I want to get this merged ASAP. So going with daily checks for now. We can always edit it later.

.github/workflows/dependent-issues.yml
Jaanki: Commented as above and quoting again

Yes, I modified the issue/PR trigger on open and re-open. But I feel daily is still too long a time gap. However, I want to get this merged ASAP. So going with daily checks for now. We can always edit it later.

dfarrell07: Agreed, we can increase the check frequency as we use the feature more.

.github/workflows/dependent-issues.yml
dfarrell07: Much better, thanks @Jaanki!

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,36 @@
+---
+name: Dependency Checks
```
dfarrell07:
```suggestion
name: PR Dependencies
```

Maybe? To be clear that this isn't meant to group the many other types of dependency-related checks we do?

.github/workflows/dependent-issues.yml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+  schedule:
+    - cron: '0 0 * * *' # check daily
+
+jobs:
+  check:
+    name: PR Dependency
```
dfarrell07:
```suggestion
    name: Update PR Dependency Labeling
```

Maybe?

.github/workflows/dependent-issues.yml
Jaanki: yes. When you comment

```Depends on: <PR>```
or
```depends on: <PR>```

File: .github/workflows/dependent-issues.yml (hunk 11 above)
Comment (Jaanki) [reply .github/workflows/dependent-issues.yml]: this along with L22 on GH UI will be 

```Dependency Checks/PR Dependency```

COmbined it makes more sense

Comment (dfarrell07) [reply .github/workflows/dependent-issues.yml]: We should likely also rename the file to match the workflow name.

File: .github/workflows/dependent-issues.yml (hunk 12 above)
Comment (Jaanki) [reply .github/workflows/dependent-issues.yml]: It is not just updating the labeling. Its checking for dependency among PRs

Comment (dfarrell07) [reply .github/workflows/dependent-issues.yml]: But when it finds dependencies, doesn't it maintain a list in comments (basically labeling)?

File: .github/workflows/dependent-issues.yml (hunk 11 above)
Comment (dfarrell07) [reply .github/workflows/dependent-issues.yml]: > this along with L22 on GH UI will be
> 
> `Dependency Checks/PR Dependency`
> 
> COmbined it makes more sense

But in some places, like the main GH Actions UI, it will only show the workflow name until you drill down. This workflow should be much more narrowly scoped than about "dependency checks" in general, it should be about PR dependencies.

File: .github/workflows/dependent-issues.yml (hunk 12 above)
Comment (Jaanki) [reply .github/workflows/dependent-issues.yml]: Not exactly.Initially we add the comment saying that PR1 is dependent on PR2. So this dependency job on PR1 will keep on failing until PR2 is merged. Once PR2 is merged, this job will notify that PR2 is merged and will pass for PR1. So we can then safely merge PR1 as well

**Review by mangelajo — APPROVED** [review]
Works for me

File: .github/workflows/dependent-issues.yml (hunk 12 above)
Comment (dfarrell07) [reply .github/workflows/dependent-issues.yml]: Right, so "labeling" is not a great name. How about something like "Check Dependencies", paired with the PR-naming context of the Workflow?

```suggestion
    name: Check Dependencies
```

dfarrell07 (approved):
LGTM, thanks @Jaanki!

PR #1172, opened 2021-03-18
Allow user to set name of CoreDNS custom ConfigMap

### LINKED ISSUE #1166: Allow to set the name of the custom coredns configmap

The ability to set the name of the custom configmap to coredns.

**Why is this needed**:

Similar to #1054, in our environment we can provide custom configmaps to coredns in kube-system but the name of the configmap must be coredns-extra-configs. Currently we have to deploy this manually:

```
apiVersion: v1
data:
  Corefile: |
    clusterset.local:53 {
        forward . <<lighthouse-coredns-clusterip>>
    }
kind: ConfigMap
metadata:
  name: coredns-extra-configs
  namespace: kube-system
```

It would be nice to have this deployed by the operator. Would also really aprecciate to have this backported to 0.8.1.

vthapar:
Currently we use `coredns-custom` as only possible name for custom
CoreDNS configMap. Not all deployments use this name, so give users
an option to configure this at the time of join.

Fixes #1166


controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -173,12 +172,14 @@ func (r *ServiceDiscoveryReconciler) Reconcile(request reconcile.Request) (recon
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	err = updateDNSCustomConfigMap(r.client, r.k8sClientSet, instance, reqLogger)
-	if errors.IsNotFound(err) {
+	if instance.Spec.CoreDNSCustomConfigName != "" {
```
skitt: This changes the default behaviour, doesn’t it? Currently we look for `coredns-custom`, update that if it’s present, or call `updateDNSConfigMap()` if it isn’t; with this patch we’ll only do that if the custom config name is set. Does that matter?

mangelajo:
Is there a way, by looking at the system coredns file to find the include of the configmap and doing it automatically instead?

mangelajo (changes requested):
see my comment for auto detection, I believe it should be possible to find the included coredns config looking at the parent system file.

vthapar:
> see my comment for auto detection, I believe it should be possible to find the included coredns config looking at the parent system file.

It is possible, but there can be multiple custom configs.

controllers/servicediscovery/servicediscovery_controller.go
vthapar: Yes, I wanted to check with Jakob before making the change. If custom config is set, we will create if not already present.

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -55,6 +55,7 @@ type SubmarinerSpec struct {
 	Debug                    bool   `json:"debug"`
 	NatEnabled               bool   `json:"natEnabled"`
 	ServiceDiscoveryEnabled  bool   `json:"serviceDiscoveryEnabled,omitempty"`
+	CoreDNSCustomConfigName  string `json:"coreDNSCustomConfigName,omitempty"`
```
aswinsuryan: We may need to get the namespace as well as core-dns could be running in a different namespace like the Openshift one.

mangelajo:
> > see my comment for auto detection, I believe it should be possible to find the included coredns config looking at the parent system file.
> 
> It is possible, but there can be multiple custom configs.

Yeah, but I think it would be not that bad to figure out which one is a top level include and use that one? 

Anyway, that could be additional to this work, for just in case we are not able to detect your current work could be the fallback.

mangelajo (approved):

apis/submariner/v1alpha1/submariner_types.go
mangelajo: Oh that's true

mangelajo: let's make it a separate struct so configname & namespace are grouped.

```
coreDnsCustomConfig:
     configMap: 
     namespace: 

```
?

mangelajo (changes requested):
see the comments, sorry ':D

vthapar:
> > > see my comment for auto detection, I believe it should be possible to find the included coredns config looking at the parent system file.
> > 
> > 
> > It is possible, but there can be multiple custom configs.
> 
> Yeah, but I think it would be not that bad to figure out which one is a top level include and use that one?
> 
> Anyway, that could be additional to this work, for just in case we are not able to detect your current work could be the fallback.

Top level may or may not be right one. It varies from deployment to deployment, so having user pass sounds like a good idea, for now at least. We can revisit this once we have more experience and better understanding of AKS deployments.

apis/submariner/v1alpha1/submariner_types.go
vthapar: @mangelajo @aswinsuryan Yeah, I was planning to do this in a later patch, but would be good to setup the CR in a way that it is easy to add more fields to it. Any ideas on what else can go here? Wondering if we can restructure this in a generic way that works for Openshift and we don't have to do Openshift specific stuff in `servicediscovery_controller.go`. We can detect Openshift in `subctl` and fill in these values in CR. Or do we address it later when we start adding support for different providers and deployments?

Brendon had a suggestion in slack that we should just make user pass something like `provider=AKS` and do rest of customization on our on. It is on our roadmap, outside scope of this PR though.

mangelajo: > @mangelajo @aswinsuryan Yeah, I was planning to do this in a later patch, but would be good to setup the CR in a way that it is easy to add more fields to it. Any ideas on what else can go here? Wondering if we can restructure this in a generic way that works for Openshift and we don't have to do Openshift specific stuff in `servicediscovery_controller.go`. We can detect Openshift in `subctl` and fill in these values in CR. Or do we address it later when we start adding support for different providers and deployments?
>

We did auto detection of stuff in subctl and ended up moving it to the operator itself, there are users of the operator that won't use subctl (ACM for example) we don't want them to need to replicate such logic.

So IMHO auto detection should come first, but we can offer an override from here.
 
> Brendon had a suggestion in slack that we should just make user pass something like `provider=AKS` and do rest of customization on our on. It is on our roadmap, outside scope of this PR though.

I think we can just add heuristics as we discover them to handle the right custom config map. Probably the first one generic should be parsing the main coredns file.

But we should not put any auto discovery logic in subctl, it should be on the operator.

I think your patch is fine, but let's include the namespace, we can make more auto-detections later.

mangelajo:
I was waiting on the namespace parameter @vthapar does that make sense?

vthapar:
> I was waiting on the namespace parameter @vthapar does that make sense?

Just added. Ran into some issues testing hence the delay :)

aswinsuryan (approved):

mangelajo (approved):

tpantelis (approved):

pkg/subctl/cmd/join.go
```diff
@@ -130,6 +132,10 @@ func addJoinFlags(cmd *cobra.Command) {
 		"maximum number of packets lost before the connection is marked as down")
 	cmd.Flags().BoolVar(&globalnetEnabled, "globalnet", true,
 		"enable/disable Globalnet for this cluster")
+	cmd.Flags().StringVar(&corednsCustomConfigMap, "coredns-custom-configmap", "",
+		"Name of the custom CoreDNS configmap to configure forwarding to lighthouse")
+	cmd.Flags().StringVar(&corednsCustomNamespace, "coredns-custom-namespace", "",
+		"Name of the namespace in which coredns is running")
```
skitt: I’m wondering if it would make sense to combine these, *e.g.* `namespace/configmap` (I don’t know if there’s a K8s convention for this)...

vthapar: That would make `namespace` a necessary parameter. Or are you suggesting something like `myconfigmap` or `myns\myconfigmap`

skitt: I’m suggesting the latter, `namespace/` would be optional. It avoids having to deal with two separate parameters...

mangelajo (approved):

vthapar:
Sorry for repeated commits. Looks like my local setup wasn't picking up image with latest changes and I ended up testing with stale images a couple times. Had to prune system to get it to work.

skitt (approved):

PR #1175, opened 2021-03-21
Bundle 0.9.0

SteveMattar:
fix(Makefile): bundle target
update kustomize and bundle configurations
create bundle 0.9.0


pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -98,7 +98,7 @@ func main() {
 	//    type: string`
 
 	re := regexp.MustCompile("`([^`]*)`")
-	reNS := regexp.MustCompile(`(?s)\s*namespace:\s*placeholder\s*`)
+	reNS := regexp.MustCompile(`(?s)\s*\$(SA_NAMESPACE):\s*placeholder\s*`)
```
mangelajo: I think this is not the same.

The previous regex was looking for 
namespace: placeholder


Now you are looking for 

$(SA_NAMESPACE): placeholder

This needs testing, but what about:

```suggestion
	reNS := regexp.MustCompile(`(?s)\s*namespace:\s*\$(SA_NAMESPACE)\s*`)
```

mangelajo (changes requested):

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
SteveMattar: Thanks for paying attention :+1:

dfarrell07 (approved):

pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go
```diff
@@ -98,7 +98,7 @@ func main() {
 	//    type: string`
 
 	re := regexp.MustCompile("`([^`]*)`")
-	reNS := regexp.MustCompile(`(?s)\s*namespace:\s*placeholder\s*`)
+	reNS := regexp.MustCompile(`(?s)\s*namespace:\s*\$(SA_NAMESPACE)\s*`)
```
skitt: I’ll check but I wonder whether this matches — going by the start of the regex, parentheses are the grouping operator, and need to be escaped to match literal parentheses.

SteveMattar: yup, I forgot to escape the parentheses as well

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -4,7 +4,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: submariner-operator
```
mangelajo, line 7: Whats this deployment in system namespace for?

config/default/manager_webhook_patch.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: controller-manager-webhook
```
mangelajo, line 5: shouldn't this be submariner-operator-webhook then?

mangelajo (changes requested):
A few questions about the webhooks

config/default/manager_auth_proxy_patch.yaml
SteveMattar: this name is just a placeholder and can be replaced when running kustomize. However, this yaml file is not included - see `config/default/kustomization.yaml`

config/default/manager_webhook_patch.yaml
SteveMattar: Same here this webhook is not part of our build. we are not using webhooks at this moment, maybe in the future.

nyechiel:
@mangelajo can you review again?

mangelajo (approved):
Thanks for the explanations Steve!

Issue #1144, opened 2021-03-09
subctl validate: Verify stale endpoints and overlapping issues in a cluster


sridhargaddam:
**What would you like to be added**:
As part of [Epic](https://github.com/submariner-io/submariner-operator/issues/1126) we want to support verification of ClusterCIDRs (both Pod and Service CIDRs) and report any errors if noticed. 

This can be done by reading the `endpoint` objects from submariner-namespace and checking if there is any overlap when vanilla submariner is used. Also, in the same context, it would be better to validate if there are any stale endpoints.

PR #1173, opened 2021-03-18
Validate pods status

maayanf24:
Validate that all submariner pods status is running, and pods are not
constantly restarting

Fix issue #1151

Signed-off-by: Maayan Friedman <maafried@redhat.com>

PR #1176, opened 2021-03-22
Validate if the clusters have overlapping CIDRs

sridhargaddam:
This PR validates the endpoint objects in the cluster and verifies if
they have overlapping CIDRs. The implementation works for both vanilla
Submariner as well as Globalnet deployments as we check the endpoint objects.
Along with overlapping check, it also verifies if there are any stale
endpoint objects in the cluster.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/1144

### DIFF
--- a/pkg/subctl/cmd/validate_deployment.go
+++ b/pkg/subctl/cmd/validate_deployment.go
@@ -17,46 +17,129 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
-	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
+	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	smClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner/pkg/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
+
+	"github.com/submariner-io/submariner-operator/apis/submariner/v1alpha1"
+	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 )
 
 var validatePodsCmd = &cobra.Command{
-	Use:   "pods",
-	Short: "Validate the submariner pods",
-	Long:  "This command checks that all the submariner pods are running",
-	Run:   validatePods,
+	Use:   "deployment",
+	Short: "Validate the submariner deployment",
+	Long: "This command checks that the submariner components are properly deployed and running" +
+		" with no overlapping CIDRs.",
+	Run: validateSubmarinerDeployment,
 }
 
 func init() {
 	validateCmd.AddCommand(validatePodsCmd)
 }
 
-func validatePods(cmd *cobra.Command, args []string) {
+func validateSubmarinerDeployment(cmd *cobra.Command, args []string) {
 	configs, err := getMultipleRestConfigs(kubeConfig, kubeContext)
 	exitOnError("error getting REST config for cluster", err)
 
 	for _, item := range configs {
-		message := fmt.Sprintf("Validating submariner pods in %q", item.clusterName)
-		status.Start(message)
-		fmt.Println()
-		checkPods(item.config, OperatorNamespace)
+		submariner := getSubmarinerResource(item.config)
+		if submariner == nil {
+			status.QueueWarningMessage(submMissingMessage)
+			status.End(cli.Success)
+			return
+		}
+
+		checkPods(item, submariner, OperatorNamespace)
+		checkOverlappingCIDRs(item, submariner)
 	}
 }
 
-func checkPods(config *rest.Config, operatorNamespace string) {
-	submariner := getSubmarinerResource(config)
-	if submariner == nil {
-		status.QueueWarningMessage(submMissingMessage)
-		status.End(cli.Success)
-		return
+func checkOverlappingCIDRs(item restConfig, submariner *v1alpha1.Submariner) {
+	var message string
+	submarinerClient, err := smClientset.NewForConfig(item.config)
+	exitOnError("Unable to get the Submariner client", err)
+
+	localClusterName := submariner.Status.ClusterID
+	endpointList, err := submarinerClient.SubmarinerV1().Endpoints(submariner.Namespace).List(metav1.ListOptions{})
+	if err != nil {
+		exitOnError("Error while listing the endpoints", err)
 	}
 
-	kubeClientSet, err := kubernetes.NewForConfig(config)
+	status.Start("Verifying if cluster CIDRs overlap")
+
+	var localEndpoint submarinerv1.Endpoint
+	overlappingClusters := make(map[string]submarinerv1.Endpoint)
+	for i, source := range endpointList.Items {
+		if localClusterName == source.Spec.ClusterID {
+			localEndpoint = source
+		}
+
+		for _, dest := range endpointList.Items[i+1:] {
+			// Currently we dont support multiple endpoints in a cluster, hence return an error.
+			// When the corresponding support is added, this check needs to be updated.
+			if source.Spec.ClusterID == dest.Spec.ClusterID {
+				message = fmt.Sprintf("Looks like some stale endpoints are present for cluster %q",
+					source.Spec.ClusterID)
+				status.QueueFailureMessage(message)
+				status.End(cli.Failure)
+				break
+			}
+
+			for _, subnet := range dest.Spec.Subnets {
+				overlap, err := util.IsOverlappingCIDR(source.Spec.Subnets, subnet)
+				if err != nil {
+					// Ideally this case will never hit, as the subnets are valid CIDRs
+					message = fmt.Sprintf("Error parsing the CIDR %s in cluster %q", err, dest.Spec.ClusterID)
+					status.QueueFailureMessage(message)
+					status.End(cli.Failure)
+					break
+				}
+
+				if overlap {
+					overlappingClusters[source.Spec.ClusterID] = source
+					overlappingClusters[dest.Spec.ClusterID] = dest
+				}
+			}
+		}
+	}
+
+	if len(overlappingClusters) > 0 {
+		if _, exists := overlappingClusters[localClusterName]; exists {
+			delete(overlappingClusters, localClusterName)
+			message = fmt.Sprintf("localCluster %q with CIDRs %q overlaps with clusters %#v",
+				localClusterName, localEndpoint.Spec.Subnets, overlappingClusters)
+		} else {
+			message = fmt.Sprintf("localCluster %q does not have overlapping CIDRs with other member clusters. "+
+				"However, the following submariner member clusters have overlapping CIDRs %#v", localClusterName,
+				getClusterDetails(overlappingClusters))
+		}
+		status.QueueFailureMessage(message)
+		status.End(cli.Failure)
+	}
+
+	status.QueueSuccessMessage("Clusters do not have overlapping CIDRs")
+	status.End(cli.Success)
+}
+
+func getClusterDetails(overlappingClusters map[string]submarinerv1.Endpoint) string {
+	output := []string{}
+	for _, cluster := range overlappingClusters {
+		output = append(output, fmt.Sprintf("clusterID: %s, subnets: %s", cluster.Spec.ClusterID, cluster.Spec.Subnets))
+	}
+	return strings.Join(output, " ; ")
+}
+
+func checkPods(item restConfig, submariner *v1alpha1.Submariner, operatorNamespace string) {
+	message := fmt.Sprintf("Validating submariner pods in %q", item.clusterName)
+	status.Start(message)
+	fmt.Println()
+
+	kubeClientSet, err := kubernetes.NewForConfig(item.config)
 
 	if err != nil {
 		exitOnError("error creating Kubernetes client", err)
@@ -89,7 +172,7 @@ func checkPods(config *rest.Config, operatorNamespace string) {
 		}
 	}
 
-	message := "All Submariner pods are up and running"
+	message = "All Submariner pods are up and running"
 	status.QueueSuccessMessage(message)
 	status.End(cli.Success)
 }
--- a/pkg/subctl/cmd/validate_deployment.go
+++ b/pkg/subctl/cmd/validate_deployment.go
@@ -17,10 +17,8 @@ package cmd
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/spf13/cobra"
-	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	smClientset "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner/pkg/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


mangelajo (approved):
I'm not sure if we are counting on that, but could it be possible to add another validation that will look for the worker node ips, and see if they overlap with those cidrs as we know it'd be problematic too

sridhargaddam:
> I'm not sure if we are counting on that, but could it be possible to add another validation that will look for the worker node ips, and see if they overlap with those cidrs as we know it'd be problematic too

Good point @mangelajo. I will add that check in a follow on PR. I hope it's fine :)

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,128 @@ package cmd
...
 )
 
 var validatePodsCmd = &cobra.Command{
-	Use:   "pods",
-	Short: "Validate the submariner pods",
-	Long:  "This command checks that all the submariner pods are running",
-	Run:   validatePods,
+	Use:   "deployment",
+	Short: "Validate the submariner deployment",
+	Long:  "This command checks if there are any issues related to overlapping CIDRs and the state of submariner pods.",
```
maayanf24: Mentioning here the check of overlapping CIDRs is too specific.
I think a more generic description would be better, as suggested in PR https://github.com/submariner-io/submariner-operator/pull/1173

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,128 @@ package cmd
...
+func getClusterDetails(overlappingClusters map[string]submarinerv1.Endpoint) string {
+	output := []string{}
+	for _, cluster := range overlappingClusters {
+		output = append(output, fmt.Sprintf("clusterID: %s, subnets: %s", cluster.Spec.ClusterID, cluster.Spec.Subnets))
+	}
+	return strings.Join(output, " ; ")
+}
+
+func checkPods(item restConfig, submariner *v1alpha1.Submariner, operatorNamespace string) {
+	message := fmt.Sprintf("* Validating submariner pods in %q", item.clusterName)
```
maayanf24: In other validation commands the message is without `*` prefix

maayanf24 (changes requested):

pkg/subctl/cmd/validate_deployment.go
sridhargaddam: It's part of a Long description that is expected to be a bit descriptive, isn't it? 
Also, in the other PR it includes the following description - "This command checks that the submariner components are properly deployed and running" . When overlapping CIDRs are present, Submariner Pods will continue to run, and only when users look into the logs of the Route-agent Pod, will know that there is some issue. So, I felt its good to include it as part of the description.

Anyways, this is a subjective thing. Let's hear from others and if the majority feel that overlapping CIDRs should be removed from the description, I'll update the PR accordingly.

pkg/subctl/cmd/validate_deployment.go
sridhargaddam: Yes, I know ;-)

I added it looking at `subctl join ...` where beginning of some operation (aka validation) is shown with some dot 

```
2021-03-22T15:35:52.6449258Z [cluster1]  • Discovering network details  ...
2021-03-22T15:35:52.6449880Z [cluster1]     Discovered network details:
2021-03-22T15:35:52.6450610Z [cluster1]         Network plugin:  generic
2021-03-22T15:35:52.6451177Z [cluster1]         Service CIDRs:   [100.0.0.0/16]
2021-03-22T15:35:52.6451675Z [cluster1]         Cluster CIDRs:   [10.0.0.0/16]
2021-03-22T15:35:52.6452442Z [cluster1]  ✓ Discovering network details
2021-03-22T15:35:52.6453226Z [cluster1]  • Discovering multi cluster details  ...
2021-03-22T15:35:52.6454078Z [cluster1]  • Validating Globalnet configurations  ...
2021-03-22T15:35:52.6454960Z [cluster1]  ✓ Validating Globalnet configurations
2021-03-22T15:35:52.6455736Z [cluster1]  • Assigning Globalnet IPs  ...
2021-03-22T15:35:52.6456461Z [cluster1]  ✓ Assigning Globalnet IPs
2021-03-22T15:35:52.6457157Z [cluster1]  ✓ GlobalCIDR is: 169.254.1.0/24
2021-03-22T15:35:52.6457903Z [cluster1]  ✓ Discovering multi cluster details
2021-03-22T15:35:52.6458709Z [cluster1]  • Deploying the Submariner operator  ...
2021-03-22T15:35:52.6459521Z [cluster1]  ✓ Deploying the Submariner operator
2021-03-22T15:35:52.6460392Z [cluster1]  ✓ Created Lighthouse service accounts and roles
2021-03-22T15:35:52.6461487Z [cluster1]  • Creating SA for cluster  ...
2021-03-22T15:35:52.6464075Z [cluster1]  ✓ Creating SA for cluster
2021-03-22T15:35:52.6464793Z [cluster1]  • Deploying Submariner  ...
2021-03-22T15:35:52.6465511Z [cluster1]  ✓ Deploying Submariner
2021-03-22T15:35:52.6466220Z [cluster1]  ✓ Submariner is up and running
```

If people like it, my idea was to change all the other commands. But anyway, I can modify if required, no issues.

pkg/subctl/cmd/validate_deployment.go
sridhargaddam: Okay, my suggestion is - "This command checks that the submariner components are properly deployed, running and there are no overlapping CIDRs"

@tpantelis @dfarrell07 any suggestions?

dfarrell07: That sounds good to me @sridhargaddam :+1: 

(I'd just add a comma after running)

dfarrell07 (approved):

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,129 @@ package cmd
...
 
 var validatePodsCmd = &cobra.Command{
-	Use:   "pods",
-	Short: "Validate the submariner pods",
-	Long:  "This command checks that all the submariner pods are running",
-	Run:   validatePods,
+	Use:   "deployment",
+	Short: "Validate the submariner deployment",
+	Long: "This command checks that the submariner components are properly deployed, running," +
+		" and there are no overlapping CIDRs.",
```
tpantelis:
```suggestion
	Long: "This command checks that the submariner components are properly deployed and running with no overlapping CIDRs.",
```

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,129 @@ package cmd
...
+			message = fmt.Sprintf("localCluster %q does not have overlapping CIDRs with other member clusters. "+
+				"However, the following submariner member clusters have overlapping CIDRs %#v", localClusterName,
+				getClusterDetails(overlappingClusters))
+		}
+		status.QueueFailureMessage(message)
+		status.End(cli.Failure)
+	}
+
+	status.QueueSuccessMessage("Clusters do not have overlapping CIDRs")
+	status.End(cli.Success)
```
tpantelis, line 108: I think we can simplify this code by checking `Clusters` instead of `Endpoints`:

```
        localClusterName := submariner.Status.ClusterID
	
	localCluster, err := submarinerClient.SubmarinerV1().Clusters(submariner.Namespace).Get(localClusterName, metav1.GetOptions{})
	exitOnError("Error getting local Cluster", err)

	localCIDRs := append(localCluster.Spec.ClusterCIDR, localCluster.Spec.ServiceCIDR...)
	status.Start(fmt.Sprintf("Validating CIDRs %v for cluster %q", localCIDRs, localClusterName))

	clusters, err := submarinerClient.SubmarinerV1().Clusters(submariner.Namespace).List(metav1.ListOptions{})
	exitOnError("Error listing Clusters", err)

	result := cli.Success
	for _, c := range clusters.Items {
		if c.Name == localClusterName {
			continue
		}

		for _, cidr := range append(c.Spec.ClusterCIDR, c.Spec.ServiceCIDR...) {
			overlap, err := util.IsOverlappingCIDR(localCIDRs, cidr)
			exitOnError(fmt.Sprintf("Error processing CIDR %Q in cluster %q", cidr, c.Name), err)

			if overlap {
				status.QueueFailureMessage(fmt.Sprintf("CIDR %q in cluster %q overlaps", cidr, c.Name))
				result = cli.Failure
			}
		}
	}

	if result == cli.Success {
		status.QueueSuccessMessage("CIDRs do not overlap with any other cluster")
	}

	status.End(result)
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_deployment.go
sridhargaddam: Thank you @tpantelis. But this code is missing checks for Globalnet CIDRs, Overlapping CIDRs in other clusters, stale endpoints etc. 

With endpoints we have certain advantages as the endpoint.Spec.Subnets is populated based on Globalnet/Vanilla Submariner - so we don't have to explicitly check Globalnet again. Also, an endpoint plays an important role for tunnels, so checking for stale endpoints will help validate any corner cases etc.

pkg/subctl/cmd/validate_deployment.go
maayanf24: Sounds good to me. Thanks @sridhargaddam

maayanf24 (approved):

skitt (approved):

pkg/subctl/cmd/validate_deployment.go
tpantelis: To cover GN, we can add: `localCIDRs := append(localCIDRs, localCluster.Spec.Global)`. Both the `Cluster` and `Endpoint`  are populated from the env vars via the `SubmarinerSpecification` - the former has the CIDRs in 3 fields and the latter in one.

The purpose of this code is to check for overlapping CIDRs, not stale endpoints, so perhaps we're overloading it and stale endpoints should be another check. 

Overlapping CIDRs in other clusters would be checked when processing those clusters. This code is in the context of a single cluster so I think it should just check that cluster. If every cluster checks every other cluster then we'll have repeated output for overlaps. If we want to check all clusters in one sweep then perhaps we should read from the broker.

sridhargaddam: > To cover GN, we can add: `localCIDRs := append(localCIDRs, localCluster.Spec.Global)`. 

If we keep adding such checks then there will not be much difference between the implementations ;-) 
And on the other hand, AFAICT a stale endpoint has more effect on the cluster functionality compared to a stale cluster object.

> Both the `Cluster` and `Endpoint` are populated from the env vars via the `SubmarinerSpecification` - the former has the CIDRs in 3 fields and the latter in one.

Yes. But it's the latter (aka endpoint.Spec.Subnets) that plays an important role during Tunnel configuration. 

> The purpose of this code is to check for overlapping CIDRs, not stale endpoints, so perhaps we're overloading it and stale endpoints should be another check.

This is basically a validation tool, to point out any issues with the Submariner deployment. The command here is `subctl validate deployment` - so if we validate stale endpoints while checking for overlapping CIDRs, we are optimizing the checks instead of creating new sub-commands. IMHO, we should also look at UI POV to minimize the number of sub-commands instead of creating additional sub-commands to `subctl validate`.

> Overlapping CIDRs in other clusters would be checked when processing those clusters. This code is in the 
> context of a single cluster so I think it should just check that cluster. If every cluster checks every other cluster
> then we'll have repeated output for overlaps. If we want to check all clusters in one sweep then perhaps we 
> should read from the broker.

AFAICT in most cases, users will run this command against a single cluster. Based on my experience supporting users, overlapping clusters is one of the common issues I have seen. If we provide warning about Overlapping Clusters even when it overlaps across other clusters, it saves a lot of time/debugging-effort for our users instead of expecting them to run the `subctl validate deployment` command on each and every cluster to ultimately know that there is some issue with Overlapping CIDRs.

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,129 @@ package cmd
...
+	for _, cluster := range overlappingClusters {
+		output = append(output, fmt.Sprintf("clusterID: %s, subnets: %s", cluster.Spec.ClusterID, cluster.Spec.Subnets))
+	}
+	return strings.Join(output, " ; ")
+}
+
+func checkPods(item restConfig, submariner *v1alpha1.Submariner, operatorNamespace string) {
+	message := fmt.Sprintf("Validating submariner pods in %q", item.clusterName)
+	status.Start(message)
+	fmt.Println()
```
tpantelis:
```suggestion
```

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -17,46 +17,129 @@ package cmd
...
-		message := fmt.Sprintf("Validating submariner pods in %q", item.clusterName)
-		status.Start(message)
-		fmt.Println()
-		checkPods(item.config, OperatorNamespace)
+		submariner := getSubmarinerResource(item.config)
+		if submariner == nil {
+			status.QueueWarningMessage(submMissingMessage)
+			status.End(cli.Success)
+			return
+		}
```
tpantelis: We need to start the status in order to end it if submariner is missing.

```suggestion
                status.Start(fmt.Sprintf("Retrieving Submariner resource from %q", item.clusterName))
                 
		submariner := getSubmarinerResource(item.config)
		if submariner == nil {
			status.QueueWarningMessage(submMissingMessage)
			status.End(cli.Success)
			continue
		}
		
		status.End(cli.Success)
```

pkg/subctl/cmd/validate_deployment.go
tpantelis: ok good points and I agree using the endpoints is better. My main concern is the complexity of the algorithm - it looks to be n^3 and it will parse CIDRs multiple times. This may be an issue with larger deployments but we can always optimize later. 

Also the overlap messages print a lot of info (entire Endpoint objects) which is a bit hard to read/decipher. And it's not entirely correct: if local "_cluster1_" overlaps with _cluster2_ and "_cluster3_" overlaps with "_cluster4_", then it will hit the code path that prints "_localCluster %q with CIDRs %q overlaps with clusters %#v_" which will include "_cluster3_"  and "_cluster4_" which isn't correct. Since we're checking for overlap across all clusters anyway, I suggest to simply print specific info for each overlap found and not treat the local cluster any different. This simplifies the code.

```
        status.Start("Checking if cluster CIDRs overlap")

	endpointList, err := submarinerClient.SubmarinerV1().Endpoints(submariner.Namespace).List(metav1.ListOptions{})

	if err != nil {
		status.QueueFailureMessage(fmt.Sprintf("Error listing the Submariner endpoints in cluster %q", localClusterName))
		status.End(cli.Failure)
		return
	}

	for i, source := range endpointList.Items {
		for _, dest := range endpointList.Items[i+1:] {
			// Currently we dont support multiple endpoints in a cluster, hence return an error.
			// When the corresponding support is added, this check needs to be updated.
			if source.Spec.ClusterID == dest.Spec.ClusterID {
				status.QueueFailureMessage(fmt.Sprintf("Found multiple Submariner endpoints (%q and %q) in cluster %q",
					source.Name, dest.Name, source.Spec.ClusterID))
				continue
			}

			for _, subnet := range dest.Spec.Subnets {
				overlap, err := util.IsOverlappingCIDR(source.Spec.Subnets, subnet)
				if err != nil {
					// Ideally this case will never hit, as the subnets are valid CIDRs
					status.QueueFailureMessage(fmt.Sprintf("Error parsing CIDR in cluster %q: %s", dest.Spec.ClusterID, err))
					continue
				}

				if overlap {
					status.QueueFailureMessage(fmt.Sprintf("CIDR %q in cluster %q overlaps with cluster %q (CIDRs: %v)",
						subnet, dest.Spec.ClusterID, source.Spec.ClusterID, source.Spec.Subnets))
				}
			}
		}
	}

	if status.HasFailureMessages() {
		status.End(cli.Failure)
		return
	}

	status.QueueSuccessMessage("Clusters do not have overlapping CIDRs")
	status.End(cli.Success)
```

Note: I added a new method to `Status`:

```
func (s *Status) HasFailureMessages() bool {
	return len(s.failureQueue) > 0
}
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_deployment.go
sridhargaddam: Ok, will update it. Thanks.

sridhargaddam: Sample output when overlapping CIDRs are present:

```
[sgaddam@localhost submariner-operator]$ ./bin/subctl validate deployment 
 ✓ Retrieving Submariner resource from "cluster2"
 ✓ Validating submariner pods in "cluster2"
 ✓ All Submariner pods are up and running
 ✗ Checking if cluster CIDRs overlap
 ✗ CIDR "10.1.0.0/16" in cluster "cluster1" overlaps with cluster "cluster-test" (CIDRs: [100.6.0.0/16 10.1.0.0/16])
```

Sample output in success scenario:
```
[sgaddam@localhost submariner-operator]$ ./bin/subctl validate deployment 
 ✓ Retrieving Submariner resource from "cluster2"
 ✓ Validating submariner pods in "cluster2"
 ✓ All Submariner pods are up and running
 ✓ Checking if cluster CIDRs overlap
 ✓ Clusters do not have overlapping CIDRs
[sgaddam@localhost submariner-operator]$
```

tpantelis (approved):

PR #1185, opened 2021-03-24
WIP: Reading Calico IPPools

sridhargaddam:


mangelajo (changes requested):
@sridhargaddam we are introducing a lot of dependencies (potentially conflicting) from calico. 

Can we use the dynamic client instead?

See an example
https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/openshift4.go#L43

sridhargaddam:
> @sridhargaddam we are introducing a lot of dependencies (potentially conflicting) from calico.
> 
> Can we use the dynamic client instead?
> 
> See an example
> https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/openshift4.go#L43

Thanks for the pointers @mangelajo. 

@maayanf24 ^^

Issue #1125, opened 2021-03-03
Epic: Automated log and data collection


vthapar:
This will track tasks needed to fully automate collection of data and logs to troubleshoot any submariner issues in deployments.

- [x] Add `subctl` sub-command to gather logs and data from a specific cluster #1152 
- [x] Add an option to gather data sub-command to gather data from all clusters #1218 
- [x] Generate tarball with all the gathered data
- [ ] Modify gather data sub-command to selectively collect logs and data based on related failures
- [x] Capture logs from submariner-operator pods #1143 
- [x] Capture logs from all submariner/connectivity pods #1143
- [x] Capture logs from all service discovery pods #1143
- [x] Capture all data present on broker #1154
- [x] Capture data from operator and related config CRs #1155 
- [x] Capture data from submariner/connectivity resources like Clusters, Gateways, configmaps etc. #1156 
- [x] Capture data from service discovery resources like ServiceExports/Imports, Endpointslices, configmaps etc. #1158 
- [x] Capture data specific to CNI/CableDriver specific #1157 
- [ ] Add `subctl` sub-command to change debug verbosity level for specific pods before gathering data
- [x] Add e2e tests for the new `subctl` gather command #1194
- [x] Add log file to include additional info of note during the collection process, eg how many resources of a type were collected.

Jaanki:
I am taking tasks 5,6 and 7 to collect logs from submariner and related pods. Raised a issue https://github.com/submariner-io/submariner-operator/issues/1143

Jaanki:
https://github.com/submariner-io/submariner-operator/pull/1188

PR #1187, opened 2021-03-25
Gather resources from Broker using generated Broker kubeconfig

### LINKED ISSUE #1154: Capture troubleshooting data from broker
Collect configuration and other data on broker for troubleshooting

**Why is this needed**:
Sub task of Epic #1125

Jaanki:
Depends on #1196 

Closes: #1154 
Epic: #1125 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.10.0
 	github.com/spf13/cobra v1.1.1
-	github.com/submariner-io/admiral v0.9.0-m1.0.20210326161940-f9d67998ec93
+	github.com/submariner-io/admiral v0.9.0-m1.0.20210330170028-a5e4008ebf84
 	github.com/submariner-io/lighthouse v0.9.0-m1
 	github.com/submariner-io/shipyard v0.9.0-m1
 	github.com/submariner-io/submariner v0.9.0-m1.0.20210329154437-ecf33cb91f20
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -24,8 +24,6 @@ import (
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/kubernetes"
 
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd/gather"
@@ -55,7 +53,7 @@ var gatherTypeFlags = map[string]bool{
 	"resources": false,
 }
 
-var gatherFuncs = map[string]func(string, *gather.Info) bool{
+var gatherFuncs = map[string]func(string, gather.Info) bool{
 	components.Connectivity:     gatherConnectivity,
 	components.ServiceDiscovery: gatherDiscovery,
 	components.Broker:           gatherBroker,
@@ -117,17 +115,14 @@ func gatherData() {
 
 	fmt.Printf("Gathering information from cluster %q\n", clusterName)
 
-	info := &gather.Info{
+	info := gather.Info{
 		RestConfig:  restConfig,
 		ClusterName: clusterName,
 		DirName:     directory,
 	}
 
-	info.ClientSet, err = kubernetes.NewForConfig(restConfig)
-	exitOnError("Error creating k8s client set", err)
-
-	info.DynClient, err = dynamic.NewForConfig(restConfig)
-	exitOnError("Error creating dynamic client", err)
+	info.DynClient, info.ClientSet, err = getClients(restConfig)
+	exitOnError("Error getting client %s", err)
 
 	for module, ok := range gatherModuleFlags {
 		if ok {
@@ -145,7 +140,7 @@ func gatherData() {
 	}
 }
 
-func gatherConnectivity(dataType string, info *gather.Info) bool {
+func gatherConnectivity(dataType string, info gather.Info) bool {
 	switch dataType {
 	case Logs:
 		err := gather.GatewayPodLogs(info)
@@ -178,7 +173,7 @@ func gatherConnectivity(dataType string, info *gather.Info) bool {
 	return true
 }
 
-func gatherDiscovery(dataType string, info *gather.Info) bool {
+func gatherDiscovery(dataType string, info gather.Info) bool {
 	switch dataType {
 	case Logs:
 		err := gather.ServiceDiscoveryPodLogs(info)
@@ -203,14 +198,29 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
 	return true
 }
 
-func gatherBroker(dataType string, info *gather.Info) bool {
+func gatherBroker(dataType string, info gather.Info) bool {
 	switch dataType {
-	case Logs:
-		info.Status.QueueWarningMessage("No logs to gather on Broker")
 	case Resources:
-		_, _ = getBrokerRestConfig(info.RestConfig)
+		brokerRestConfig, brokerNamespace, err := getBrokerRestConfigAndNamespace(info.RestConfig)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting the broker's rest config: %s", err))
+			return true
+		}
 
-		info.Status.QueueWarningMessage("Gather Broker Resources not implemented yet")
+		info.DynClient, info.ClientSet, err = getClients(brokerRestConfig)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting the broker client %s", err))
+			return true
+		}
+
+		info.RestConfig = brokerRestConfig
+		info.ClusterName = "broker"
+
+		// The broker's ClusterRole used by member clusters only allows the below resources to be queried
+		gather.Endpoints(info, brokerNamespace)
+		gather.Clusters(info, brokerNamespace)
+		gather.EndpointSlices(info, brokerNamespace)
+		gather.ServiceImports(info, brokerNamespace)
 	default:
 		return false
 	}
@@ -218,7 +228,7 @@ func gatherBroker(dataType string, info *gather.Info) bool {
 	return true
 }
 
-func gatherOperator(dataType string, info *gather.Info) bool {
+func gatherOperator(dataType string, info gather.Info) bool {
 	switch dataType {
 	case Logs:
 		info.Status.QueueWarningMessage("Gather Operator Logs not implemented yet")
--- a/pkg/subctl/cmd/resource.go
+++ b/pkg/subctl/cmd/resource.go
@@ -101,25 +101,25 @@ func getGatewaysResource(config *rest.Config) *submarinerv1.GatewayList {
 	return gateways
 }
 
-func getBrokerRestConfig(localConfig *rest.Config) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespace(localConfig *rest.Config) (*rest.Config, string, error) {
 	submarinerClient, err := subOperatorClientset.NewForConfig(localConfig)
 	if err != nil {
-		return nil, errors.WithMessage(err, "error getting submariner client")
+		return nil, "", errors.WithMessage(err, "error getting submariner client")
 	}
 
-	brokerConfig, err := getBrokerRestConfigFromSubmariner(submarinerClient)
+	brokerConfig, brokerNamespace, err := getBrokerRestConfigAndNamespaceFromSubmariner(submarinerClient)
 	if apierrors.IsNotFound(err) {
-		return getBrokerRestConfigFromServiceDisc(submarinerClient)
+		return getBrokerRestConfigAndNamespaceFromServiceDisc(submarinerClient)
 	}
 
-	return brokerConfig, err
+	return brokerConfig, brokerNamespace, err
 }
 
-func getBrokerRestConfigFromSubmariner(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespaceFromSubmariner(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, string, error) {
 	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).
 		Get(submarinercr.SubmarinerName, v1opts.GetOptions{})
 	if err != nil {
-		return nil, errors.WithMessage(err, "error obtaining the Submariner resource")
+		return nil, "", errors.WithMessage(err, "error obtaining the Submariner resource")
 	}
 
 	// Try to authorize against the submariner Cluster resource as we know the CRD should exist and the credentials
@@ -129,16 +129,16 @@ func getBrokerRestConfigFromSubmariner(submarinerClient *subOperatorClientset.Cl
 			Group:    submarinerv1.SchemeGroupVersion.Group,
 			Version:  submarinerv1.SchemeGroupVersion.Version,
 			Resource: "clusters",
-		})
+		}, submariner.Spec.BrokerK8sRemoteNamespace)
 
-	return restConfig, err
+	return restConfig, submariner.Spec.BrokerK8sRemoteNamespace, err
 }
 
-func getBrokerRestConfigFromServiceDisc(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespaceFromServiceDisc(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, string, error) {
 	serviceDisc, err := submarinerClient.SubmarinerV1alpha1().ServiceDiscoveries(OperatorNamespace).
 		Get(names.ServiceDiscoveryCrName, v1opts.GetOptions{})
 	if err != nil {
-		return nil, errors.WithMessage(err, "error obtaining the ServiceDiscovery resource")
+		return nil, "", errors.WithMessage(err, "error obtaining the ServiceDiscovery resource")
 	}
 
 	// Try to authorize against the ServiceImport resource as we know the CRD should exist and the credentials
@@ -148,7 +148,7 @@ func getBrokerRestConfigFromServiceDisc(submarinerClient *subOperatorClientset.C
 			Group:    "multicluster.x-k8s.io",
 			Version:  "v1alpha1",
 			Resource: "serviceimports",
-		})
+		}, serviceDisc.Spec.BrokerK8sRemoteNamespace)
 
-	return restConfig, err
+	return restConfig, serviceDisc.Spec.BrokerK8sRemoteNamespace, err
 }
--- a/scripts/kind-e2e/lib_subctl_gather_test.sh
+++ b/scripts/kind-e2e/lib_subctl_gather_test.sh
@@ -42,6 +42,9 @@ function test_subctl_gather() {
 
   validate_pod_log_files $subm_ns '-l component=submariner-lighthouse'
   validate_pod_log_files kube-system '-l k8s-app=kube-dns'
+
+  # Broker
+  with_context "$broker" validate_broker_resources
 }
 
 function validate_pod_log_files() {
@@ -66,6 +69,11 @@ function validate_resource_files() {
   local resource=$2
   local kind=$3
   local selector=$4
+  local cluster_name=$5
+
+  if [[ $cluster_name == "" ]]; then
+    cluster_name=${cluster}
+  fi
 
   names=$(kubectl get $resource --namespace=$ns $selector -o=jsonpath='{.items..metadata.name}')
   read -ra names_array <<< "$names"
@@ -73,7 +81,7 @@ function validate_resource_files() {
   short_res=$(echo $resource | awk -F. '{ print $1 }')
 
   for name in "${names_array[@]}"; do
-    file=$out_dir/${cluster}_${short_res}_${ns}_${name}.yaml
+    file=$out_dir/${cluster_name}_${short_res}_${ns}_${name}.yaml
     cat $file
 
     kind_count=$(grep "kind: $kind$" $file | wc -l)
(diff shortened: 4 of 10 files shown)


pkg/subctl/cmd/gather/broker.go
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+*/
+
+package gather
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const brokerNamespace = "submariner-k8s-broker"
```
mangelajo: The broker isn't on a fixed place, 

You can probably benefit from this here (to get the rest config to the cluster), and also from the submariner object you can get the namespace of the broker (see https://github.com/submariner-io/submariner-operator/blob/2efe168a8015624ec875c6d363e781c59639d039/apis/submariner/v1alpha1/submariner_types.go#L41 ) 

https://github.com/submariner-io/submariner-operator/pull/1190/files

mangelajo (changes requested):

pkg/subctl/cmd/resource.go
```diff
@@ -110,3 +126,70 @@ func getBrokerRestConfig(submariner *v1alpha1.Submariner) (*rest.Config, error)
...
+
+func IsUnknownAuthorityError(err error) bool {
+	if urlError, ok := err.(*url.Error); ok {
+		if _, ok := urlError.Unwrap().(x509.UnknownAuthorityError); ok {
+			return true
+		}
+	}
+
+	return false
+}
```
Jaanki: Would be removed and implementation from admiral will be used.

pkg/subctl/cmd/resource.go
```diff
@@ -110,3 +126,70 @@ func getBrokerRestConfig(submariner *v1alpha1.Submariner) (*rest.Config, error)
...
+	}, nil
+}
+
+func IsAuthorizedFor(restConfig *rest.Config, gvr schema.GroupVersionResource) (bool, error) {
+	client, err := dynamic.NewForConfig(restConfig)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = client.Resource(gvr).Namespace(broker.SubmarinerBrokerNamespace).Get("any", metav1.GetOptions{})
```
Jaanki: added ns here

Jaanki:
Depends on #1196

Jaanki:
depends on #1196

pkg/subctl/cmd/gather.go
```diff
@@ -190,13 +186,34 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+		gather.Endpoints(info, SubmarinerNamespace)
+
+		gather.Clusters(info, brokerNamespace)
+
+		//gather.EndpointSlices(info, brokerNamespace)
+
+		//gather.ServiceImports(info, brokerNamespace)
 
-		info.Status.QueueWarningMessage("Gather Broker Resources not implemented yet")
+		//gather.ServiceExports(info, brokerNamespace)
```
Jaanki: Function implementation in PR #1196

Jaanki: Is the ns "brokerNamespace" correct? I don't see any of these resources listed on my Broker.

pkg/subctl/cmd/gather/types.go
```diff
@@ -27,8 +27,12 @@ type Info struct {
...
-	ClientSet   kubernetes.Interface
-	DynClient   dynamic.Interface
 	ClusterName string
 	DirName     string
+	Clients     Clients
+}
+
+type Clients struct {
+	ClientSet   kubernetes.Interface
+	DynClient   dynamic.Interface
```
tpantelis: I don't think adding a nested struct here adds any value. In fact it makes references longer: `info.Clients.ClientSet` vs `info.ClientSet`.

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 
 // Pinned to kubernetes-1.17.0
 replace (
+	github.com/submariner-io/admiral v0.9.0-m1.0.20210326161940-f9d67998ec93 => github.com/submariner-io/admiral v0.9.0-m1.0.20210330145336-e9efa4fe64e5
```
tpantelis: You don't need to do this - just change the version on L22. I just put in `devel` and let the build fill in the actual version.

pkg/subctl/cmd/gather.go
```diff
@@ -190,13 +186,34 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
 }
 
 func gatherBroker(dataType string, info *gather.Info) bool {
+	brokerRestConfig, err := getBrokerRestConfig(info.Submariner)
+	exitOnError("Error getting Broker rest config", err)
```
tpantelis: `QueueFailureMessage` and return.

pkg/subctl/cmd/gather.go
```diff
@@ -129,11 +126,10 @@ func gatherData() {
 		DirName:     directory,
 	}
 
-	info.ClientSet, err = kubernetes.NewForConfig(restConfig)
-	exitOnError("Error creating k8s client set", err)
+	dynClient, clientSet, err := getClients(restConfig)
```
```suggestion
	info.DynClient, info.ClientSet, err = getClients(restConfig)
```

pkg/subctl/cmd/gather.go
```diff
@@ -190,13 +186,34 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
 }
 
 func gatherBroker(dataType string, info *gather.Info) bool {
+	brokerRestConfig, err := getBrokerRestConfig(info.Submariner)
+	exitOnError("Error getting Broker rest config", err)
+
+	dynClient, clientSet, err := getClients(brokerRestConfig)
+	exitOnError("Error getting client %s", err)
```
tpantelis: ...same here

pkg/subctl/cmd/gather.go
```diff
@@ -190,13 +186,34 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+
+	info = &gather.Info{
+		Clients:     gather.Clients{ClientSet:clientSet, DynClient:dynClient},
+		RestConfig:  brokerRestConfig,
+		ClusterName: "broker",
+	}
+
```
tpantelis: The code above should be moved to L207 below as it doesn't apply to `Logs`.

pkg/subctl/cmd/gather.go
```diff
@@ -190,13 +186,34 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+	exitOnError("Error getting Broker rest config", err)
+
+	dynClient, clientSet, err := getClients(brokerRestConfig)
+	exitOnError("Error getting client %s", err)
+
+	info = &gather.Info{
+		Clients:     gather.Clients{ClientSet:clientSet, DynClient:dynClient},
+		RestConfig:  brokerRestConfig,
+		ClusterName: "broker",
+	}
```
tpantelis: You should make a complete copy of `info`,  ie `brokerInfo := *info`. However I suggest just making this implicit in the function signature, ie pass `info` by value rather than pointer so each invocation gets its own copy and can modify it as they see fit:  

```suggestion
	info.RestConfig = brokerRestConfig
	info.ClusterName: = "broker"
	
	info.DynClient, info.ClientSet, err = getClients(brokerRestConfig)
	...
```

pkg/subctl/cmd/gather.go
tpantelis: `ServiceExports` are not present on the broker.

tpantelis (changes requested):

pkg/subctl/cmd/gather/types.go
Jaanki: I wanted to separate clients. Anyways, will remove this

pkg/subctl/cmd/gather.go
```diff
@@ -106,7 +104,6 @@ func gatherData() {
 	submariner := getSubmarinerResource(restConfig)
 	if submariner == nil {
 		fmt.Println(submMissingMessage)
-		return
```
tpantelis: This change isn't necessary - this code was removed by https://github.com/submariner-io/submariner-operator/pull/1202 anyway.

pkg/subctl/cmd/gather.go
```diff
@@ -192,11 +186,32 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+		info.Status.QueueSuccessMessage("No pods accessible. No logs to collect on Broker")
 	case Resources:
-		_, _ = getBrokerRestConfig(info.Submariner)
+		brokerRestConfig, err := getBrokerRestConfig(info.Submariner)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting Broker rest config %s", err))
+			return true
+		}
+
+		info.DynClient, info.ClientSet, err = getClients(brokerRestConfig)
```
tpantelis, line 210: `info` is passed by reference so you can't modify unless you restore its previous state on return. But, as I mentioned earlier, you can simply pass by value instead, ie

`func gatherBroker(dataType string, info gather.Info) bool {`

pkg/subctl/cmd/gather.go
```diff
@@ -192,11 +186,32 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting client %s", err))
+			return true
+		}
+
+		info.RestConfig = brokerRestConfig
+		info.ClusterName = "broker"
+		brokerNamespace := info.Submariner.Spec.BrokerK8sRemoteNamespace
+
+		// NewBrokerClusterRole allows only below resources to be queried
+		gather.Endpoints(info, SubmarinerNamespace)
```
```suggestion
		gather.Endpoints(info, brokerNamespace)
```

pkg/subctl/cmd/gather/types.go
```diff
@@ -24,11 +24,11 @@ import (
 )
 
 type Info struct {
+	ClientSet   kubernetes.Interface
+	DynClient   dynamic.Interface
```
tpantelis: You essentially didn't change anything here so please put the fields back where they were to avoid a diff.

tpantelis (changes requested):
You also need to rebase.

pkg/subctl/cmd/gather.go
Jaanki: Not addressed this comment. Will do in coming update

pkg/subctl/cmd/gather.go
```diff
@@ -193,11 +188,31 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting client %s", err))
+			return true
+		}
+
+		info.RestConfig = brokerRestConfig
+		info.ClusterName = "broker"
+
+		// NewBrokerClusterRole allows only below resources to be queried
+		gather.Endpoints(&info, brokerNamespace)
```
Jaanki, line 209:
```suggestion
		gather.Endpoints(info, brokerNamespace)
```

pkg/subctl/cmd/gather.go
```diff
@@ -193,11 +188,31 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+
+		gather.Clusters(&info, brokerNamespace)
```
Jaanki, line 211:
```suggestion
		gather.Clusters(info, brokerNamespace)
```

pkg/subctl/cmd/gather.go
Jaanki: I guess this states otherwise https://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/

tpantelis: Actually that article indicates it may be cheaper to pass structs by value. In other languages it's typically more efficient to pass by pointer (Java only allows that). Regardless, passing by pointer does allow the recipient to modify the struct such that its effects are observed on return.  This may not be desirable and is not in this case. You are modifying fields of `info` which is subsequently passed to the next gather function and which sees `ClusterName` as "broker" and `DynClient` pointing to the broker cluster.  You can see this effect in the E2E tests. Noticed they all failed, eg [this one](https://github.com/submariner-io/submariner-operator/pull/1187/checks?check_run_id=2239780706). After **Gathering broker resources**, it goes onto **Gathering operator resources** but that fails: "_Failed to gather submariners: error listing "submariners": submariners.submariner.io is forbidden: User "system:serviceaccount:submariner-k8s-broker:cluster-cluster2" cannot list resource "submariners" in API group "submariner.io" in the namespace "submariner-operator"_".  So the resource isn't written to the file and the subsequent verification fails. This is b/c `info.DynClient` is now erroneously pointing to the broker cluster which doesn't have the CRD.

So you must make a copy of `info`, modify and use the copy and leave the original unmodified, hence my suggestion. Passing by value accomplishes this and inherently allows each gather function to safely modify the struct if need be w/o unwanted side-effects.

tpantelis:
You should also add e2e tests for the broker resources to _scripts/kind-e2e/lib_subctl_gather_test.sh_.

pkg/subctl/cmd/gather.go
Jaanki: yes. Its the `after gather broker` calls that are affected. I realised this much later after I had logged off for the day. I initially thought the concern is about the config changes `before gather broker` is called

pkg/subctl/cmd/gather.go
```diff
@@ -203,22 +198,39 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
 	return true
 }
 
-func gatherBroker(dataType string, info *gather.Info) bool {
+func gatherBroker(dataType string, info gather.Info) bool {
 	switch dataType {
 	case Logs:
-		info.Status.QueueWarningMessage("No logs to gather on Broker")
+		info.Status.QueueSuccessMessage("No pods accessible. No logs to collect on Broker")
```
tpantelis: This isn't exactly true - we do deploy an operator pod on the broker. I think we should just remove these 2 lines so nothing is outputted - we can add it back if we decide to gather the pod logs.

pkg/subctl/cmd/gather.go
```diff
@@ -203,22 +198,39 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
-		_, _ = getBrokerRestConfig(info.RestConfig)
+		brokerRestConfig, brokerNamespace, err := getBrokerRestConfigAndNamespace(info.RestConfig)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting Broker rest config or namespace %s", err))
+			return true
+		}
+
+		info.DynClient, info.ClientSet, err = getClients(brokerRestConfig)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting client %s", err))
```
```suggestion
			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting Broker client: %s", err))
```

pkg/subctl/cmd/gather.go
```diff
@@ -203,22 +198,39 @@ func gatherDiscovery(dataType string, info *gather.Info) bool {
...
+func gatherBroker(dataType string, info gather.Info) bool {
 	switch dataType {
 	case Logs:
-		info.Status.QueueWarningMessage("No logs to gather on Broker")
+		info.Status.QueueSuccessMessage("No pods accessible. No logs to collect on Broker")
 	case Resources:
-		_, _ = getBrokerRestConfig(info.RestConfig)
+		brokerRestConfig, brokerNamespace, err := getBrokerRestConfigAndNamespace(info.RestConfig)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting Broker rest config or namespace %s", err))
```
```suggestion
			info.Status.QueueFailureMessage(fmt.Sprintf("Error getting Broker rest config: %s", err))
```

scripts/kind-e2e/lib_subctl_gather_test.sh
```diff
@@ -90,3 +90,12 @@ function validate_resource_files() {
...
 }
 
+function validate_broker_resources() {
+
+  ${DAPPER_SOURCE}/bin/subctl gather --kubeconfig ${KUBECONFIGS_DIR}/kind-config-${cluster} --dir $out_dir
+
+  validate_resource_files $SUBMARINER_BROKER_NS 'endpoints.submariner.io' 'Endpoint'
+  validate_resource_files $SUBMARINER_BROKER_NS 'clusters.submariner.io' 'Cluster'
+  validate_resource_files all 'serviceimports.multicluster.x-k8s.io' 'ServiceImport'
+  validate_resource_files all 'endpointslices.discovery.k8s.io' 'EndpointSlice'
```
tpantelis: They're synced to the broker NS.

```suggestion
  validate_resource_files $SUBMARINER_BROKER_NS 'serviceimports.multicluster.x-k8s.io' 'ServiceImport'
  validate_resource_files $SUBMARINER_BROKER_NS 'endpointslices.discovery.k8s.io' 'EndpointSlice'
```

scripts/kind-e2e/lib_subctl_gather_test.sh
```diff
@@ -90,3 +90,12 @@ function validate_resource_files() {
...
+}
```
tpantelis, line 106: We don't need to run `subctl gather` again specifically for the broker.  You should be able to do:

```
function validate_broker_resources() {
  validate_resource_files $SUBMARINER_BROKER_NS 'endpoints.submariner.io' 'Endpoint'
  validate_resource_files $SUBMARINER_BROKER_NS 'clusters.submariner.io' 'Cluster'
  validate_resource_files $SUBMARINER_BROKER_NS 'serviceimports.multicluster.x-k8s.io' 'ServiceImport'
  validate_resource_files $SUBMARINER_BROKER_NS 'endpointslices.discovery.k8s.io' 'EndpointSlice'
}

function test_subctl_gather() {
  ...
  # Broker
  with_context "$broker" validate_broker_resources
}
```

Also the E2E failed in the broker validation b/c it uses "cluster1" in the file names instead of "broker". We can modify `validate_resource_files` to take an additional optional arg:

```
function validate_resource_files() {
  ...
  local cluster_name=$5

  if [[ $cluster_name == "" ]]; then
    cluster_name=${cluster}
  fi
  ...
  for name in "${names_array[@]}"; do
    file=$out_dir/${cluster_name}_${short_res}_${ns}_${name}.yaml
    ...
  done
}
```

Then to call it

`validate_resource_files $SUBMARINER_BROKER_NS 'endpoints.submariner.io' 'Endpoint' '' 'broker'`

pkg/subctl/cmd/resource.go
```diff
@@ -128,16 +128,16 @@ func getBrokerRestConfigFromSubmariner(submarinerClient *subOperatorClientset.Cl
...
+	return restConfig, submariner.Spec.BrokerK8sRemoteNamespace, err
 }
 
-func getBrokerRestConfigFromServiceDisc(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespaceFromServiceDisc(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, string, error) {
 	serviceDisc, err := submarinerClient.SubmarinerV1alpha1().ServiceDiscoveries(OperatorNamespace).
 		Get(names.ServiceDiscoveryCrName, v1opts.GetOptions{})
 	if err != nil {
-		return nil, errors.WithMessage(err, "error obtaining the ServiceDiscovery resource")
+		return nil, "", errors.WithMessage(err, "error obtaining the ServiceDiscovery resource")
```
maayanf24, line 141:
```suggestion
		return nil, "", errors.WithMessage(err, "Error obtaining the ServiceDiscovery resource")
```

tpantelis (changes requested):

pkg/subctl/cmd/resource.go
```diff
@@ -100,25 +100,25 @@ func getGatewaysResource(config *rest.Config) *submarinerv1.GatewayList {
...
+	return brokerConfig, brokerNamespace, err
 }
 
-func getBrokerRestConfigFromSubmariner(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespaceFromSubmariner(submarinerClient *subOperatorClientset.Clientset) (*rest.Config, string, error) {
 	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(OperatorNamespace).
 		Get(submarinercr.SubmarinerName, v1opts.GetOptions{})
 	if err != nil {
-		return nil, errors.WithMessage(err, "error obtaining the Submariner resource")
+		return nil, "", errors.WithMessage(err, "error obtaining the Submariner resource")
```
maayanf24, line 122:
```suggestion
		return nil, "", errors.WithMessage(err, "Error obtaining the Submariner resource")
```

pkg/subctl/cmd/resource.go
```diff
@@ -100,25 +100,25 @@ func getGatewaysResource(config *rest.Config) *submarinerv1.GatewayList {
 	return gateways
 }
 
-func getBrokerRestConfig(localConfig *rest.Config) (*rest.Config, error) {
+func getBrokerRestConfigAndNamespace(localConfig *rest.Config) (*rest.Config, string, error) {
 	submarinerClient, err := subOperatorClientset.NewForConfig(localConfig)
 	if err != nil {
-		return nil, errors.WithMessage(err, "error getting submariner client")
+		return nil, "", errors.WithMessage(err, "error getting submariner client")
```
maayanf24, line 107:
```suggestion
		return nil, "", errors.WithMessage(err, "Error getting submariner client")
```

maayanf24 (changes requested):

pkg/subctl/cmd/resource.go
tpantelis: Actually `error` strings shouldn't be [capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

maayanf24: But we have them in subctl, so it should be uniformed

tpantelis: Which places? I'm glancing thru some of the code and don't see capitalized error strings. There may be some but they should be changed.

maayanf24 (approved):

tpantelis (approved):

PR #1190, opened 2021-03-26
Add function to create broker REST config

tpantelis:
... from the `Submariner` resource Spec fields.

### DIFF
--- a/pkg/subctl/cmd/resource.go
+++ b/pkg/subctl/cmd/resource.go
@@ -18,14 +18,16 @@ package cmd
 import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1opts "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/submariner-operator/apis/submariner/v1alpha1"
-	submarinerclientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
+	subOperatorClientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/submarinercr"
-	submarinerclientv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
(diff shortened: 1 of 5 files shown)


pkg/subctl/cmd/resource.go
```diff
@@ -95,3 +97,13 @@ func getGatewaysResource(config *rest.Config) *submarinerclientv1.GatewayList {
 
 	return gateways
 }
+
+func getBrokerRestConfig(submariner *v1alpha1.Submariner) (*rest.Config, error) {
+	tlsClientConfig := rest.TLSClientConfig{}
+
```
tpantelis: @skitt Similar code in admiral does this:

```
          if useTokenCAForEndpoint {
                caDecoded, err := base64.StdEncoding.DecodeString(brokerSpec.Ca)
                if err != nil {
                        return nil, "", fmt.Errorf("error decoding CA data: %v", err)
                }
                tlsClientConfig.CAData = caDecoded
        }
```

and the caller first tries w/o the token then with it if the first try fails. I'm not really clear on this. Do we need to do the same here or is this OK?

mangelajo: Probably we need to do the same.

Some endpoints work just with the CA, some with both, I believe this was due to a difference in rancher endpoints. But I can't recall the exact details now.

We could not do it, but in that case, can you leave a comment/link here so we have a reference if problems arise we can investigate & document the issue better.

mangelajo (changes requested):
Just a comment

mangelajo (approved):
I'm changing my vote to +1 with those comments (we can follow up), as this is probably necessary for @Jaanki's work.

pkg/subctl/cmd/resource.go
tpantelis: Will do. Waiting on @skitt to weight in. If we need to do exactly what admiral is doing it would make sense to reuse that but it would need an admiral patch we well.

skitt: Yes, I can’t remember the details right now but IIRC we had to do this for OpenShift on AWS as well in some setups.

tpantelis: So should we just always use the CA? Are there cases it would not work with the CA?

skitt: We need the check, see https://github.com/submariner-io/submariner/commit/dc68b2c3c97ab5cfcf89c7a2364495fd5806dffb and https://github.com/submariner-io/submariner/issues/649 for details.

tpantelis: OK - https://github.com/submariner-io/admiral/pull/199

dfarrell07 (approved):

PR #1186, opened 2021-03-25
Allow operator logging to be configured

skitt:
... in a similar fashion to other pods.

Depends on #1180.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

PR #1192, opened 2021-03-26
Allow operator logging to be configured

skitt:
... in a similar fashion to other pods.


skitt:
This is the same as #1186 but with the correct target.

dfarrell07 (approved):

mangelajo:
> This is the same as #1186 but with the correct target.

Sorry about merging to wrong target, I'm thinking about adding something to submariner-bot: reporting as "non-passing" CI for PRs not on top of devel, so we can't merge them, and when finally rebased to devel, report as passed.

mangelajo (approved):

skitt:
> > This is the same as #1186 but with the correct target.
> 
> Sorry about merging to wrong target, I'm thinking about adding something to submariner-bot: reporting as "non-passing" CI for PRs not on top of devel, so we can't merge them, and when finally rebased to devel, report as passed.

A GHA can do this: https://github.com/submariner-io/shipyard/pull/506

Issue #1148, opened 2021-03-09
subctl validate: Verify if firewall configuration allows ingress TCP 8080 on the Gateway nodes.


sridhargaddam:
**What would you like to be added**:
As part of [Epic](https://github.com/submariner-io/submariner-operator/issues/1126) this issue is to verify if firewall configuration allows TCP/8080 on the Gateway nodes where metrics are supported.

PR #1195, opened 2021-03-28
Validate firewall configuration for prometheus metrics

sridhargaddam:
This PR validates if the firewall configuration in the cluster allows
TCP/8080 traffic on the Gateway nodes. Validation is done by spawning
a tcpdump sniffer pod on the Gateway node and a client pod on one of
the non-Gateway nodes. The clientPod tries to send some tcp traffic
to the Gateway node while tcpdump is executed in the sniffer pod.
Finally, when both the Pods finish their execution, the Pod output is
compared with expected values and appropriate message is displayed to
the user.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/1148

### DIFF
--- a/pkg/subctl/cmd/validate_fw_metrics.go
+++ b/pkg/subctl/cmd/validate_fw_metrics.go
@@ -29,7 +29,7 @@ const (
 
 var validateFirewallMetricsCmd = &cobra.Command{
 	Use:   "metrics",
-	Short: "Validate if firewall allows metrics on the Gateway nodes.",
+	Short: "Validate firewall access to metrics.",
 	Long:  "This command checks whether firewall configuration allows metrics to be read from the Gateway nodes.",
 	Run:   validateFirewallMetricsConfig,
 }
@@ -99,5 +99,5 @@ func validateFirewallMetricsConfigWithinCluster(item restConfig) {
 		return
 	}
 
-	status.QueueSuccessMessage("The firewall configuration for prometheus metrics is working fine.")
+	status.QueueSuccessMessage("Prometheus metrics can be retrieved from gateway nodes.")
 }
--- a/pkg/subctl/cmd/validate_fw_metrics.go
+++ b/pkg/subctl/cmd/validate_fw_metrics.go
@@ -1,0 +1,103 @@
+/*
+© 2021 Red Hat, Inc. and others.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


pkg/subctl/cmd/validate_fw_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	// Verify that tcpdump output (i.e, from snifferPod) contains the HostIP of clientPod
+	if !strings.Contains(sPod.PodOutput, cPod.Pod.Status.HostIP) {
+		message := fmt.Sprintf("The tcpdump output from the sniffer pod does not contain the"+
+			" client pod HostIP. Please check that your firewall configuration allows TCP/8080 traffic"+
+			" on the %q node.", sPod.Pod.Spec.NodeName)
+		status.QueueFailureMessage(message)
+		return
+	}
+
+	status.QueueSuccessMessage("The firewall configuration for prometheus metrics is working fine.")
```
skitt:
```suggestion
	status.QueueSuccessMessage("Prometheus metrics can be retrieved from gateway nodes.")
```

I’m a bit late to the party on this series, but we’re not really checking the firewall — we don’t know that there is one — we’re checking connectivity.

sridhargaddam: That's true @skitt. We are checking if the metrics port (8080) on the Gateway node is accessible. This is an indirect way of ensuring that even if the firewall is enabled, its properly configured to allow the port. I'll update the comment. Thanks for reviewing.

mangelajo (approved):
looks good, we should add E2E for it.

pkg/subctl/cmd/validate_fw_metrics.go
dfarrell07: Does this mean that the overall command should be less about the (potential) firewall?

pkg/subctl/cmd/validate_fw_metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	TCPSniffMetricsCommand = "tcpdump -ln -c 5 -i any tcp and src port 9898 and dst port 8080 and 'tcp[tcpflags] == tcp-syn'"
+)
+
+var validateFirewallMetricsCmd = &cobra.Command{
+	Use:   "metrics",
+	Short: "Validate if firewall allows metrics on the Gateway nodes.",
```
tpantelis:
```suggestion
	Short: "Validate firewall access to metrics",
```

tpantelis (changes requested):
Looks good - just one minor thing.

tpantelis (approved):

dfarrell07 (approved):

maayanf24 (approved):

PR #1200, opened 2021-03-30
Add additional parameters for IPSec

mangelajo:
to subctl join:
--preferred-server
--force-udp-encaps

which are passed down to the submariner-gateway as:

- CE_IPSEC_PREFERREDSERVER
- CE_IPSEC_FORCEENCAPS

### DIFF
--- a/apis/submariner/v1alpha1/submariner_types.go
+++ b/apis/submariner/v1alpha1/submariner_types.go
@@ -52,6 +52,8 @@ type SubmarinerSpec struct {
 	CeIPSecIKEPort           int                  `json:"ceIPSecIKEPort,omitempty"`
 	CeIPSecNATTPort          int                  `json:"ceIPSecNATTPort,omitempty"`
 	CeIPSecDebug             bool                 `json:"ceIPSecDebug"`
+	CeIPSecPreferredServer   bool                 `json:"ceIPSecPreferredServer,omitempty"`
+	CeIPSecForceUDPEncaps    bool                 `json:"ceIPSecForceUDPEncaps,omitempty"`
 	Debug                    bool                 `json:"debug"`
 	NatEnabled               bool                 `json:"natEnabled"`
 	ServiceDiscoveryEnabled  bool                 `json:"serviceDiscoveryEnabled,omitempty"`
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -58,6 +58,8 @@ var (
 	imageVersion                  string
 	nattPort                      int
 	ikePort                       int
+	preferredServer               bool
+	forceUDPEncaps                bool
 	colorCodes                    string
 	natTraversal                  bool
 	disableNat                    bool
@@ -95,6 +97,11 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&natTraversal, "natt", true, "enable NAT traversal for IPsec")
 	cmd.Flags().BoolVar(&disableNat, "disable-nat", false, "disable NAT for IPsec")
 	err := cmd.Flags().MarkDeprecated("disable-nat", "please use --natt=false instead")
+
+	cmd.Flags().BoolVar(&preferredServer, "preferred-server", false,
+		"enable this cluster as a preferred server for dataplane connections")
+	cmd.Flags().BoolVar(&forceUDPEncaps, "force-udp-encaps", false, "force UDP encapsulation for IPSec")
+
 	// Errors here are fatal programming errors
 	exitOnError("deprecation error", err)
 	cmd.Flags().BoolVar(&ipsecDebug, "ipsec-debug", false, "enable IPsec debugging (verbose logging)")
@@ -481,6 +488,8 @@ func populateSubmarinerSpec(subctlData *datafile.SubctlData, netconfig globalnet
 		CeIPSecNATTPort:          nattPort,
 		CeIPSecIKEPort:           ikePort,
 		CeIPSecDebug:             ipsecDebug,
+		CeIPSecForceUDPEncaps:    forceUDPEncaps,
+		CeIPSecPreferredServer:   preferredServer,
 		CeIPSecPSK:               base64.StdEncoding.EncodeToString(subctlData.IPSecPSK.Data["psk"]),
 		BrokerK8sCA:              base64.StdEncoding.EncodeToString(subctlData.ClientToken.Data["ca.crt"]),
 		BrokerK8sRemoteNamespace: string(subctlData.ClientToken.Data["namespace"]),
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -61,7 +61,9 @@ issues:
         - lll
         - stylecheck
       path: "pkg/subctl/operator/common/embeddedyamls/yamls.go"
-
+    - linters:
+        - maligned
+      path: "apis/submariner/v1alpha1/submariner_types.go"
     # BrokerK8sApiServer parameter is used by other projects, like ACM,
     # so not changing it to BrokerK8sAPIServer as suggested by stylecheck
     - linters:
--- a/config/crd/bases/submariner.io_submariners.yaml
+++ b/config/crd/bases/submariner.io_submariners.yaml
@@ -50,12 +50,16 @@ spec:
                 type: string
               ceIPSecDebug:
                 type: boolean
+              ceIPSecForceUDPEncaps:
+                type: boolean
               ceIPSecIKEPort:
                 type: integer
               ceIPSecNATTPort:
                 type: integer
               ceIPSecPSK:
                 type: string
+              ceIPSecPreferredServer:
+                type: boolean
               clusterCIDR:
                 type: string
               clusterID:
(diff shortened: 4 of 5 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -95,6 +97,11 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&natTraversal, "natt", true, "enable NAT traversal for IPsec")
 	cmd.Flags().BoolVar(&disableNat, "disable-nat", false, "disable NAT for IPsec")
 	err := cmd.Flags().MarkDeprecated("disable-nat", "please use --natt=false instead")
+
+	cmd.Flags().BoolVar(&preferredServer, "preferred-server", false,
+		"enable this cluster as a preferred server for dataplane connections")
+	cmd.Flags().BoolVar(&forceUDPEncaps, "force-udp-encaps", false, "force UDP encapsulation for IPSec")
```
sridhargaddam, line 103: If NatDiscovery figures out that there is no NAT between the clusters (so ideally they can communicate using ESP) and user inputs `--force-udp-encaps`, does it cause IPsec to use UDP Encapsulation?

pkg/subctl/cmd/join.go
```diff
@@ -95,6 +97,11 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&natTraversal, "natt", true, "enable NAT traversal for IPsec")
 	cmd.Flags().BoolVar(&disableNat, "disable-nat", false, "disable NAT for IPsec")
 	err := cmd.Flags().MarkDeprecated("disable-nat", "please use --natt=false instead")
+
+	cmd.Flags().BoolVar(&preferredServer, "preferred-server", false,
```
sridhargaddam, line 101: In a simple two-cluster setup, if the user supplies `preferred-server` for both of them, how do we handle the scenario?

pkg/subctl/cmd/join.go
mangelajo: yes, it forces to always use encapsulation, even if there is no nat

That's the purpose :-)

pkg/subctl/cmd/join.go
mangelajo: yes, I'm coding that part :)

```
package libreswan

import (
	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
	"github.com/submariner-io/submariner/pkg/types"
	"k8s.io/klog"
)

type operationMode int

const (
	operationModeLegacy operationMode = iota
	operationModeServer
	operationModeClient
)

func (i *libreswan) calculateOperationMode(remoteEndpoint types.SubmarinerEndpoint) operationMode {
	defaultValue := false
	leftPreferred, err := i.localEndpoint.Spec.GetBackendBool(v1.PreferredServerConfig, &defaultValue)
	if err != nil {
		klog.Errorf("Error parsing local endpoint config: %s", err)
	}

	rightPreferred, err := remoteEndpoint.Spec.GetBackendBool(v1.PreferredServerConfig, nil)
	if err != nil {
		klog.Errorf("Error parsing remote endpoint config %q: %s", remoteEndpoint.Spec.CableName, err)
	}

	if rightPreferred == nil {
		return operationModeLegacy
	}

	if *leftPreferred && !*rightPreferred {
		return operationModeServer
	} else if *rightPreferred && !*leftPreferred {
		return operationModeClient
	}

	// At this point we didn't have enough info to decide, we break ties based on the cable name
	if i.localEndpoint.Spec.CableName > remoteEndpoint.Spec.CableName {
		return operationModeServer
	}

	return operationModeClient
}

```

pkg/subctl/cmd/join.go
sridhargaddam: Okay, thanks for the clarification.

sridhargaddam (approved):

pkg/subctl/cmd/join.go
mangelajo: May be if none is preferred we can stick to the legacyMode... we could call it something else "bidirectional" or something

pkg/subctl/cmd/join.go
```diff
@@ -95,6 +97,11 @@ func addJoinFlags(cmd *cobra.Command) {
...
+		"enable this cluster as a preferred server for dataplane connections")
```
dfarrell07, line 102: Maybe we can say something about what it means to be a prefered server? What impact does this have, and/or why would I use this?

mangelajo: I thought about that but the other explanations are kind of short .  I was planning to write a more detailed explanation on the docs.

apis/submariner/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -22,7 +22,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
```
mangelajo: ha, my IDE keeps modifying that file import I wonder why

dfarrell07 (approved):

tpantelis (approved):

sridhargaddam (approved):

mangelajo:
Checking a solution for the linter suggested by @dfarrell07

tpantelis (approved):

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -52,6 +52,8 @@ type SubmarinerSpec struct {
 	CeIPSecIKEPort           int                  `json:"ceIPSecIKEPort,omitempty"`
 	CeIPSecNATTPort          int                  `json:"ceIPSecNATTPort,omitempty"`
 	CeIPSecDebug             bool                 `json:"ceIPSecDebug"`
+	CeIPSecPreferredServer   *bool                `json:"ceIPSecPreferredServer,omitempty"`
+	CeIPSecForceUDPEncaps    *bool                `json:"ceIPSecForceUDPEncaps,omitempty"`
```
skitt: Why are these pointers? They can be plain `bool` AFAICT (false when absent).

mangelajo: to allow the "omitEmpty", otherwise we can omit them, or can we?

mangelajo: Otherwise they will get added to https://github.com/submariner-io/submariner-operator/pull/1200/files#diff-3bf1b6c02ff8069cea7b4c42751006ce5baafde160d68e6272782f726e9c738bL111

tpantelis: good point.

tpantelis: Do we need to omit if empty?

skitt: `bool` values can be `omitempty` without being pointers, they are then omitted if false (see https://golang.org/pkg/encoding/json/#Marshal).

mangelajo: I wanted to avoid issues on upgrade of the operator for example, but may be there is no such issue?

skitt: This would also fix the alignment linting error ;-).

mangelajo: For example the ACM team using the operator alone will need to start providing those new fields, but they are optional. I believe it's better to keep with omitEmpty / non-required

tpantelis: There shouldn't be. On deserialization from previous version, if the field is missing it will be false, exactly if it was omitted if empty. Adding a field is fine for backwards compat.

skitt: No, they wouldn’t. Absent values would be unmarshalled as false. (I’m building to check, just to make sure.)

skitt: So `make manifests` just finished with

```suggestion
	CeIPSecPreferredServer   bool                 `json:"ceIPSecPreferredServer,omitempty"`
	CeIPSecForceUDPEncaps    bool                 `json:"ceIPSecForceUDPEncaps,omitempty"`
```

and the CRDs stay the same: the values aren’t required, and old CRs can still be unmarshalled.

skitt: > No, they wouldn’t. Absent values would be unmarshalled as false. (I’m building to check, just to make sure.)

This was in reply to Miguel, not Tom ;-).

tpantelis: > > No, they wouldn’t. Absent values would be unmarshalled as false. (I’m building to check, just to make sure.)
> 
> This was in reply to Miguel, not Tom ;-).

I know 😄

skitt: See `ServiceDiscoveryEnabled` for an existing example of this, just below :smile:

mangelajo: Oh I was so sure that without omitEmpty they become "required", I'll update the patch in the next minutes, thanks!

mangelajo: I mean, I was sure that without making it a pointer it would fail... 'XD

mangelajo: Done.

PR #1202, opened 2021-03-31
Use Submariner or ServiceDiscovery to get broker config

tpantelis:
The `Submariner` resource may not exist if only service discovery was installed so use the `ServiceDiscovery` if `Submariner` is not found.


dfarrell07 (approved):

dfarrell07:
@Jaanki Looks like the PR dependency check is failing with:

> Error: Resource not accessible by integration

mangelajo (approved):

PR #1206, opened 2021-03-31
subctl validate connections: show error message when there are no connections at all

### LINKED ISSUE #1213: subctl validate connections will show "All connections are established" for a cluster with no connections
**What happened**:
when running `subctl validate connection` on a cluster with no connections,
the output message is: `All connections are established`

The output should be: `There are no active connections`

maayanf24:
Show error message when there are no connections at all

### DIFF
--- a/pkg/subctl/cmd/validate_connections.go
+++ b/pkg/subctl/cmd/validate_connections.go
@@ -64,6 +64,13 @@ func validateConnections(cmd *cobra.Command, args []string) {
 			if gateway.Status.HAStatus != submv1.HAStatusActive {
 				continue
 			}
+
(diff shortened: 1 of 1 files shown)


tpantelis:
The PR title indicates a bug - can you explain in the summary why this is a bug? Is this linked to an issue?

skitt:
More generally, the commit title should describe what the change _does_, not where it came from; so in this case, “Show error message when there are no connections at all” would be an appropriate title, instead of being in the message body.

The message body then explains the context, links to bugs etc.

sridhargaddam (approved):

nyechiel (approved):

tpantelis (approved):

Issue #1210, opened 2021-04-01
subctl join should list master nodes when labelled as workers


sridhargaddam:
**What happened**:
In some deployments (like a master only deployment) where the master node[s] are labelled as workers, `subctl join ...` should list such nodes to the user for selection of Gateway node where there is no node with the gateway label. Currently, nodes with master label are skipped.

PR #1212, opened 2021-04-01
subctl join to list master nodes when labelled as workers

sridhargaddam:
In some deployments (like a master only deployment) where the master
node[s] are labelled as workers, subctl join should list such nodes
to the user for selection of Gateway node.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/1210

### DIFF
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -163,19 +163,27 @@ func handleNodeLabels(config *rest.Config) error {
 }
 
 func askForGatewayNode(clientset kubernetes.Interface) (struct{ Node string }, error) {
-	// List all nodes and select one
-	allNodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "!node-role.kubernetes.io/master"})
+	// List the worker nodes and select one
+	workerNodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/worker"})
 	if err != nil {
 		return struct{ Node string }{}, err
 	}
-	if len(allNodes.Items) == 0 {
-		return struct{ Node string }{}, nil
+	if len(workerNodes.Items) == 0 {
+		// In some deployments (like KIND), worker nodes are not explicitly labelled. So list non-master nodes.
+		workerNodes, err = clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "!node-role.kubernetes.io/master"})
+		if err != nil {
+			return struct{ Node string }{}, err
+		}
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/root.go
```diff
@@ -163,14 +163,22 @@ func handleNodeLabels(config *rest.Config) error {
 }
 
 func askForGatewayNode(clientset kubernetes.Interface) (struct{ Node string }, error) {
-	// List all nodes and select one
-	allNodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "!node-role.kubernetes.io/master"})
+	// List the worker nodes and select one
+	allNodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/worker"})
```
dfarrell07:
```suggestion
	workerNodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/worker"})
```

pkg/subctl/cmd/root.go
```diff
@@ -163,14 +163,22 @@ func handleNodeLabels(config *rest.Config) error {
...
 	if err != nil {
 		return struct{ Node string }{}, err
 	}
 	if len(allNodes.Items) == 0 {
-		return struct{ Node string }{}, nil
+		// In some deployments (like KIND), worker nodes are not explicitly labelled. So list non-master nodes.
+		allNodes, err = clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "!node-role.kubernetes.io/master"})
```
dfarrell07:
```suggestion
		nonMasterNodes, err = clientset.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "!node-role.kubernetes.io/master"})
```

dfarrell07 (commented):
It seems like we're using the `allNodes` var for two distinct things. Maybe name them as suggested inline?

pkg/subctl/cmd/root.go
sridhargaddam: If we use separate variable names for both, then we will have to use additional checks in Line182, Line186.

As you know, allNodes is just a local variable. Instead of having two separate variables, is it okay if I change it from `allNodes` to `workerNodes`?

dfarrell07: That makes sense @sridhargaddam, is more clear. :+1:

sridhargaddam: Ack. Thanks for reviewing @dfarrell07

tpantelis (approved):

dfarrell07 (approved):

Issue #1147, opened 2021-03-09
subctl validate: Verify if firewall configuration allows ingress UDP 4500/500 on the Gateway nodes.


sridhargaddam:
**What would you like to be added**:
As part of [Epic](https://github.com/submariner-io/submariner-operator/issues/1126) this issue is to verify the firewall configuration on all the nodes if UDP/4500 and UDP/500 is allowed or not when using kubeproxy iptables route-agent.

PR #1214, opened 2021-04-01
Verify if tunnels can be setup on Gateway node

sridhargaddam:
This PR validates if tunnels can be successfully established on the
Gateway node of local cluster. Validation is done by spawning a sniffer
pod on the local cluster Gateway node and a client pod on the non-Gateway
node of a remote cluster. The clientPod sends some UDP traffic to the
local Gateway node on the tunnelPort while tcpdump is executed in the
sniffer Pod. Finally, when both the Pods finish their execution, the
Pod output is compared with expected values and appropriate message
is displayed to the user.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/1147


pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+		return
+	}
+
+	defer sPod.DeletePod()
+	var gatewayPodIP string
+	if !localEndpoint.Spec.NATEnabled {
+		gatewayPodIP = localEndpoint.Spec.PrivateIP
+	} else {
+		gatewayPodIP = localEndpoint.Spec.PublicIP
+	}
```
mangelajo: This won't be accurate with the auto-detection mechanism. Perhaps we should try with both the Private & Public IPs and report which one worked.

mangelajo (changes requested):
Just one thing that needs handling.

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+
+	"github.com/submariner-io/submariner-operator/apis/submariner/v1alpha1"
+)
+
+const (
+	ClientSourcePort = "9898"
```
mangelajo, line 33: Ok, I think this can be problematic. If we filter on source ports to detect that a packet arrives properly to the gateway, that can fail because the SRC port under some nat conditions can be modified.

As we were talking in slack, may be we can follow the next approach:

1) Try to find packets with src "ClientSourcePort", if that fails, push a WARNING about this
2) Try to find packets with the specific content (tcpdump matching on specific packet data), if this fails ERROR.

sounds ok? :)

sridhargaddam: Yup. I'll update the PR.

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+var verboseOutput bool
+
+var validateTunnelCmd = &cobra.Command{
+	Use:   "tunnel <localkubeconfig> <remotekubeconfig>",
+	Short: "Validate if firewall allows tunnels on the Gateway node.",
```
tpantelis:
```suggestion
	Short: "Validate firewall access to Gateway node tunnels",
```

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+	Long:  "This command checks whether firewall configuration allows tunnel to be configured on the Gateway nodes.",
```
tpantelis:
```suggestion
	Long:  "This command checks whether firewall configuration allows tunnels to be configured on the Gateway nodes.",
```

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+		if len(args) != 2 {
+			return fmt.Errorf("two kubeconfigs must be specified")
+		}
+		same, err := compareFiles(args[0], args[1])
+		if err != nil {
+			return err
+		}
+		if same {
+			return fmt.Errorf("kubeconfig file <localkubeconfig> and <remotekubeconfig> need to have" +
+				" a unique content")
```
tpantelis:
```suggestion
			return fmt.Errorf("the specified kubeconfig files are the same")
```

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+	switch submariner.Spec.CableDriver {
+	case "libreswan", "wireguard":
+		tunnelPort, err = endpoint.Spec.GetBackendPort(subv1.UDPPortConfig, int32(submariner.Spec.CeIPSecNATTPort))
+		if err != nil {
+			status.QueueWarningMessage(fmt.Sprintf("Error reading tunnelPort: %v", err))
+		}
+		return tunnelPort, nil
+	default:
+		message := fmt.Sprintf("Error parsing the tunnelPort for CableDriver %q",
+			submariner.Spec.CableDriver)
```
tpantelis:
```suggestion
		message := fmt.Sprintf("Could not determine the tunnel port for cable driver %q",
			submariner.Spec.CableDriver)
```

pkg/subctl/cmd/validate_tunnel.go
```diff
@@ -0,0 +1,186 @@
...
+		status.QueueWarningMessage(message)
```
tpantelis:
```suggestion
		status.QueueFailureMessage(message)
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_tunnel.go
sridhargaddam: Done

sridhargaddam:
@mangelajo @tpantelis addressed the review comments. Can you please take a re-look. Thanks.

tpantelis (approved):

mangelajo (approved):

PR #1215, opened 2021-04-02
Use contexts for subctl verify

### LINKED ISSUE #1130: subctl verify fails if the current context doesn’t match the first kubeconfig given
**What happened**:

`subctl verify kubeconfig1 kubeconfig2` failed with

```
Could not obtain the cluster name from kube config: api.Config{Kind:"", APIVersion:"", Preferences:api.Preferences{Colors:false, Extensions:map[string]runtime.Object{}}, Clusters:map[string]*api.Cluster{"cluster1":(*api.Cluster)(0xc0003f8c00)}, AuthInfos:map[string]*api.AuthInfo{"cluster1":(*api.AuthInfo)(0xc00037c0f0)}, Contexts:map[string]*api.Context{"cluster1":(*api.Context)(0xc0003cb2c0)}, CurrentContext:"cluster2", Extensions:map[string]runtime.Object{}}subctl version: v0.9.0-m1
```

I expected the setup including both clusters to be verified.

Deploy two clusters, set the current context to the second, and run `subctl verify` with the first cluster’s kubeconfig as first argument.

skitt:
Allow the user to specify contexts as arguments to subctl verify,
instead of kubeconfigs; this allows contexts to be loaded from the
KUBECONFIG environment variable, and with the help of the loading fix
in Shipyard, avoids errors when the default context is set in the
kubeconfigs.

To ease the transition, kubeconfigs are still supported.

Depends on: https://github.com/submariner-io/shipyard/pull/511
Fixes: #1130

### DIFF
--- a/pkg/subctl/cmd/benchmark.go
+++ b/pkg/subctl/cmd/benchmark.go
@@ -78,7 +78,11 @@ func checkBenchmarkArguments(args []string, intraCluster bool) error {
 }
 
 func testThroughput(cmd *cobra.Command, args []string) {
-	configureTestingFramework(args)
+	err := configureTestingFramework(args)
+	if err != nil {
+		fmt.Println(err.Error())
+		return
+	}
 
 	if benchmark.Verbose {
 		fmt.Printf("Performing throughput tests\n")
@@ -87,7 +91,11 @@ func testThroughput(cmd *cobra.Command, args []string) {
 }
 
 func testLatency(cmd *cobra.Command, args []string) {
-	configureTestingFramework(args)
+	err := configureTestingFramework(args)
+	if err != nil {
+		fmt.Println(err.Error())
+		return
+	}
 
 	if benchmark.Verbose {
 		fmt.Printf("Performing latency tests\n")
--- a/pkg/subctl/cmd/verify.go
+++ b/pkg/subctl/cmd/verify.go
@@ -53,6 +53,7 @@ var (
 )
 
 func init() {
+	addKubecontextsFlag(verifyCmd)
 	verifyCmd.Flags().StringVar(&verifyOnly, "only", strings.Join(getAllVerifyKeys(), ","), "comma separated verifications to be performed")
 	verifyCmd.Flags().BoolVar(&disruptiveTests, "disruptive-tests", false, "enable disruptive verifications like gateway-failover")
 	verifyCmd.Flags().BoolVar(&disruptiveTests, "enable-disruptive", false, "enable disruptive verifications like gateway-failover")
@@ -75,7 +76,7 @@ func addVerifyFlags(cmd *cobra.Command) {
 }
 
 var verifyCmd = &cobra.Command{
-	Use:   "verify <kubeConfig1> <kubeConfig2>",
+	Use:   "verify --kubecontexts <kubeContext1>,<kubeContext2>",
 	Short: "Run verifications between two clusters",
 	Long: `This command performs various tests to verify that a Submariner deployment between two clusters
 is functioning properly. The verifications performed are controlled by the --only and --enable-disruptive
@@ -96,7 +97,14 @@ The following verifications are deemed disruptive:
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		testType := ""
-		configureTestingFramework(args)
+		if len(args) > 0 {
+			fmt.Println("subctl verify with kubeconfig arguments is deprecated, please use --kubecontexts instead")
+		}
+		err := configureTestingFramework(args)
+		if err != nil {
+			fmt.Println(err.Error())
+			return
+		}
 
 		disruptive := extractDisruptiveVerifications(verifyOnly)
 		if !disruptiveTests && len(disruptive) > 0 {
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -45,3 +45,5 @@ replace (
 	k8s.io/cloud-provider => k8s.io/cloud-provider v0.17.0
 	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.5.14
 )
+
+replace github.com/submariner-io/shipyard => github.com/skitt/shipyard v0.0.0-20210401160514-2d121be1c8a3
```
skitt: This will be removed before merging, once https://github.com/submariner-io/shipyard/pull/511 is merged.

tpantelis (approved):

skitt:
@mangelajo regarding your edit changing “Depends on” to “Depends-on”, note that the configured keyword is the former: https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/dependent-issues.yml#L38

mangelajo (approved):

pkg/subctl/cmd/verify.go
```diff
@@ -76,7 +74,7 @@ func addVerifyFlags(cmd *cobra.Command) {
 }
 
 var verifyCmd = &cobra.Command{
-	Use:   "verify <kubeConfig1> <kubeConfig2>",
+	Use:   "verify <kubeContext1> <kubeContext1>",
```
mkolesnik: There is already `kubecontext` flag, why not be consistent with it and use it, and in the command add a `kubecontext2` flag (or whatever name makes sense) and make sure they're properly set?

For backwards compatibility you can still accept the kubeconfig args, but those will be optional and only checked if the flags werent specified.

mkolesnik (changes requested):

pkg/subctl/cmd/verify.go
mangelajo: Stephen is not going to be around before we do the release.

Let's get this fixed for now. We can handle the nit later.

mangelajo: It's ok because the code also supports kubeconfig files for backwards compatibility,.

tpantelis (approved):

pkg/subctl/cmd/verify.go
mkolesnik: > Stephen is not going to be around before we do the release.
> 
> Let's get this fixed for now. We can handle the nit later.

It is not a nit but rather a UX issue.

As a user I can run `subctl verify --kubecontext cluster3 cluster1 cluster2` which from a "command line flags" perspective is valid, but is rather confusing.

Instead I propose we use `subctl verify --kubecontext cluster3 --othercontext cluster2` (or some such name) to allow a consistent experience.

> It's ok because the code also supports kubeconfig files for backwards compatibility,.

Yes we can still support that regardless, I'm not talking about that.

mkolesnik (changes requested):

pkg/subctl/cmd/verify.go
skitt: I’ve changed the PR to use `--kubecontexts` as added by @vthapar elsewhere.

tpantelis (approved):

pkg/subctl/cmd/verify.go
```diff
@@ -179,19 +195,14 @@ func clusterNameFromConfig(kubeConfigPath, kubeContext string) string {
 }
 
 func checkValidateArguments(args []string) error {
-	if len(args) != 2 {
-		return fmt.Errorf("two kubeconfigs must be specified")
+	if len(args) != 2 && len(kubeContexts) != 2 {
```
mkolesnik, line 210: Hmm for clarity perhaps the `kubeContexts` flag should have it's own error, and the old style should have a warning that it's deprecated?

mkolesnik (changes requested):

pkg/subctl/cmd/verify.go
skitt: I’ve left this test as-is (it’s testing negatives), but there is now a deprecation warning and passing two invalid files produces an error.

skitt: To be clear, I don’t think it’s worth spending much time on `args`-related subtleties, I want to get rid of them entirely in the next release.

skitt:
This breaks `benchmark`.

tpantelis:
I assume this should no longer be a draft?

skitt:
> I assume this should no longer be a draft?

It should, because it’s broken (see the comment just above).

dfarrell07 (approved):

pkg/subctl/cmd/verify.go
```diff
@@ -148,23 +156,43 @@ func isNonInteractive(err error) bool {
...
+			framework.TestContext.KubeConfigs = args
+
+			// Read the cluster names from the given kubeconfigs
+			for _, config := range args {
+				framework.TestContext.ClusterIDs = append(framework.TestContext.ClusterIDs, clusterNameFromConfig(config, ""))
+			}
+		} else {
+			// Something happened (possibly IsNotExist, but we don’t care about specifics)
+			return fmt.Errorf("the provided arguments (%v) aren't accessible files", args)
+		}
```
mkolesnik:
```suggestion
		// Something happened (possibly IsNotExist, but we don’t care about specifics)
		if err1 != nil || err2 != nil {
			return fmt.Errorf("the provided arguments (%v) aren't accessible files", args)
		}

		framework.TestContext.KubeConfig = ""
		framework.TestContext.KubeConfigs = args

		// Read the cluster names from the given kubeconfigs
		for _, config := range args {
			framework.TestContext.ClusterIDs = append(framework.TestContext.ClusterIDs, clusterNameFromConfig(config, ""))
		}
```

pkg/subctl/cmd/benchmark.go
```diff
@@ -78,7 +78,11 @@ func checkBenchmarkArguments(args []string, intraCluster bool) error {
 }
```
mkolesnik, line 78: These should probably also be handled, but in a different PR

pkg/subctl/cmd/verify.go
mkolesnik: I was more concerned with letting users know that args are deprecated :)

pkg/subctl/cmd/verify.go
```diff
@@ -179,19 +207,14 @@ func clusterNameFromConfig(kubeConfigPath, kubeContext string) string {
...
 func checkValidateArguments(args []string) error {
-	if len(args) != 2 {
-		return fmt.Errorf("two kubeconfigs must be specified")
+	if len(args) != 2 && len(kubeContexts) != 2 {
+		return fmt.Errorf("two kubecontexts must be specified")
 	}
-	if strings.Compare(args[0], args[1]) == 0 {
-		return fmt.Errorf("kubeconfig file <kubeConfig1> and <kubeConfig2> cannot be the same file")
-	}
-	same, err := compareFiles(args[0], args[1])
```
mkolesnik, line 188: This removal will cause a regression, was this intentionally removed (instead of adjusted to only check if files were sent as args)?

mkolesnik (changes requested):
Thanks @skitt , some comments.

Also we need follow up fixes in other PRs to `validate tunnel` and `benchmark` which are using same code bits.

pkg/subctl/cmd/verify.go
skitt: It was intentionally adjusted to only check if files were provided as arguments...

pkg/subctl/cmd/benchmark.go
skitt: Yes, that’s the plan ;-).

pkg/subctl/cmd/verify.go
skitt: I’ve restored the comparison.

mkolesnik (approved):

PR #1216, opened 2021-04-02
Use build tools infra for kustomize and controller-gen

skitt:
Instead of manually downloading kustomize and controller-gen every
time the Dapper image needs to be rebuilt, declare them as build tools
dependencies, so that they are vendored in. Build steps which need the
tools then ensure they are built.

This results in faster builds when the images need to be rebuilt
locally (many of the vendored dependencies are shared with the
operator and would be downloaded anyway).


SteveMattar (approved):
Thanks @skitt I guess we don't need `pkg/controller-gen` anymore

dfarrell07 (approved):

mangelajo (approved):
on demand install sounds good :)

Makefile
```diff
@@ -125,15 +125,20 @@ pkg/subctl/operator/common/embeddedyamls/yamls.go: pkg/subctl/operator/common/em
 	go generate pkg/subctl/operator/common/embeddedyamls/generate.go
 
 # Operator CRDs
-deploy/crds/submariner.io_servicediscoveries.yaml: ./apis/submariner/v1alpha1/servicediscovery_types.go vendor/modules.txt
-	controller-gen $(CRD_OPTIONS) paths="./..." output:crd:artifacts:config=deploy/crds
+CONTROLLER_GEN := $(CURDIR)/bin/controller-gen
+$(CONTROLLER_GEN): vendor/modules.txt
+	mkdir -p bin
```
mkolesnik:
```suggestion
	mkdir -p $$(dirname $@)
```

I guess something like this should be more robust?

Makefile
```diff
@@ -147,42 +152,45 @@ generate-clientset: vendor/modules.txt
...
 # test if VERSION matches the semantic versioning rule
 is-semantic-version:
 	[[ $(VERSION) =~ $(PATTERN) ]] || \
 	(printf '\nerror: VERSION does not match the format required by operator-sdk.\n\n' && exit 1)
 
 # Generate kustomization.yaml for bundle
-kustomization: is-semantic-version manifests
+KUSTOMIZE := $(CURDIR)/bin/kustomize
+$(KUSTOMIZE): vendor/modules.txt
+	mkdir -p bin
```
mkolesnik:
```suggestion
	mkdir -p $$(dirname $@)
```

Same here?

Makefile
```diff
@@ -147,42 +152,45 @@ generate-clientset: vendor/modules.txt
...
+	go build -o $@ sigs.k8s.io/kustomize/kustomize/v3
+
+kustomization: $(KUSTOMIZE) is-semantic-version manifests
 	operator-sdk generate kustomize manifests -q && \
-	cd config/manager && kustomize edit set image controller=$(IMG) && cd ../../ && \
+	(cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)) && \
```
mkolesnik, line 175: Nice :)

mkolesnik (changes requested):
Thanks for looking into this!

A couple of comments for your consideration :)

Makefile
skitt: Good idea, I’ll raise you a `mkdir -p $(@D)` ;-).

mkolesnik: You're a true `make` :mage:

mkolesnik (approved):

dfarrell07:
> curl: (18) transfer closed with 69405816 bytes remaining to read
The command '/bin/sh -c curl -Lo /usr/bin/operator-sdk "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu" &&     chmod a+x /usr/bin/operator-sdk' returned a non-zero code: 18
time="2021-04-09T15:11:26Z" level=fatal msg="exit status 18"

PR #1217, opened 2021-04-04
0.9.0-m1 release of subctl deploys devel images

### LINKED ISSUE #1129: 0.9.0-m1 release of subctl deploys devel images
**What happened**:

`subctl` 0.9.0-m1 deploys devel images

0.8.1-style behavior, where `subctl` deploys the default version out-of-the-box (without needing to use the `--version` flag)

tpantelis:
The `if` condition in the Makefile that calculates `versions.DefaultSubmarinerOperatorVersion` when building `subctl` needs to be made a var (surrounded by `$()`, ie `$(eq(...,devel))`) so it is properly expanded and evaluated so it doesn't always yield "devel".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ bin/subctl-%: generate-embeddedyamls $(shell find pkg/subctl/ -name "*.go") vend
 	export GOARCH GOOS; \
 	$(SCRIPTS_DIR)/compile.sh \
 		--ldflags "-X github.com/submariner-io/submariner-operator/pkg/version.Version=$(CALCULATED_VERSION) \
-			   -X=github.com/submariner-io/submariner-operator/pkg/versions.DefaultSubmarinerOperatorVersion=$(if eq($(patsubst subctl-devel-%,devel,$(CALCULATED_VERSION)),devel),devel,$(CALCULATED_VERSION))" \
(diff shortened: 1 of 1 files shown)


tpantelis:
Looking at [this e2e job](https://github.com/submariner-io/submariner-operator/pull/1217/checks?check_run_id=2264371835), `github.com/submariner-io/submariner-operator/pkg/version.Version=` which means `CALCULATED_VERSION` evaluated to "". As a result of this change,`github.com/submariner-io/submariner-operator/pkg/versions.DefaultSubmarinerOperatorVersion`  also evaluated to "". Prior to this change, `github.com/submariner-io/submariner-operator/pkg/versions.DefaultSubmarinerOperatorVersion` always evaluated to `devel` (see [this job](https://github.com/submariner-io/submariner-operator/pull/1216/checks?check_run_id=2254856878) from another PR). So it appears fixing the expression will inadvertently break `devel` builds. It seems the assumption is that `CALCULATED_VERSION`  will be set to `subctl-devel-<something>` such that the `if` part will evaluate to `devel`.  @skitt any idea?

skitt:
I don’t think `$(eq)` can be used in this fashion; see https://github.com/submariner-io/submariner-operator/pull/1222 for what I think is the correct fix (and yes, my original test was rubbish, thanks for spotting that!).

tpantelis:
> I don’t think `$(eq)` can be used in this fashion; see #1222 for what I think is the correct fix (and yes, my original test was rubbish, thanks for spotting that!).

It does work - I tested locally.

tpantelis:
Superseded by https://github.com/submariner-io/submariner-operator/pull/1227

PR #1220, opened 2021-04-06
Pull tags on submariner-operator release

mangelajo:


mkolesnik (changes requested):
This is not needed, as the checkout action already fetches the tags

https://github.com/submariner-io/submariner-operator/runs/2271314453#step:2:522

mangelajo:
oh, we are already fetching "all", not depth 1...

PR #1223, opened 2021-04-06
subctl gather: Add support for multiple clusters

### LINKED ISSUE #1218: subctl gather: Add option to gather data from multiple clusters
Add option to `subctl gather` command to collect data from multiple clusters.

* Kubeconfig can be provided with `--kubeconfig` flag or `KUBECONFIG` env variable
* User can select clusters by specifying list of contexts from which to gather. These contexts must be present in the kubeconfig specified.
* If no context is specified, default will be to gather from all clusters present in the kubeconfig.

**Why is this needed**:
Sub-task of Epic #1125

vthapar:
* Modify `kubecontext` param to be a comma separated list
* If none specified, get all contexts present in config
* Use single context for commands that operate on single cluter

Fixes #1218

### DIFF
--- a/pkg/subctl/cmd/resource.go
+++ b/pkg/subctl/cmd/resource.go
@@ -32,7 +32,7 @@ import (
 	subClientsetv1 "github.com/submariner-io/submariner/pkg/client/clientset/versioned"
 )
 
-func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, error) {
+func getMultipleRestConfigs(kubeConfigPath string, kubeContexts []string) ([]restConfig, error) {
 	var restConfigs []restConfig
 
 	rules := clientcmd.NewDefaultClientConfigLoadingRules()
@@ -41,8 +41,8 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
 	rules.ExplicitPath = kubeConfigPath
 
 	contexts := []string{}
-	if kubeContext != "" {
-		contexts = append(contexts, kubeContext)
+	if len(kubeContexts) > 0 {
+		contexts = append(contexts, kubeContexts...)
 	} else {
 		kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(rules, overrides)
 		rawConfig, err := kubeConfig.RawConfig()
--- a/pkg/subctl/cmd/root.go
+++ b/pkg/subctl/cmd/root.go
@@ -43,9 +43,10 @@ import (
 )
 
 var (
-	kubeConfig  string
-	kubeContext string
-	rootCmd     = &cobra.Command{
+	kubeConfig   string
+	kubeContext  string
+	kubeContexts []string
+	rootCmd      = &cobra.Command{
 		Use:   "subctl",
 		Short: "An installer for Submariner",
 	}
@@ -60,7 +61,15 @@ func init() {
 
 func addKubeconfigFlag(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&kubeConfig, "kubeconfig", "", "absolute path(s) to the kubeconfig file(s)")
-	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "kubeconfig context to use")
+	cmd.PersistentFlags().StringSliceVar(&kubeContexts, "kubecontext", nil, "kubeconfig context to use")
+	if len(kubeContexts) > 0 {
+		kubeContext = kubeContexts[0]
+	}
+}
+
+func addKubecontextsFlag(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringSliceVar(&kubeContexts, "kubecontexts", nil,
+		"comma separated list of kubeconfig contexts to use. If none specified, all contexts referenced by kubeconfig are used")
 }
 
 const (
--- a/scripts/kind-e2e/lib_subctl_gather_test.sh
+++ b/scripts/kind-e2e/lib_subctl_gather_test.sh
@@ -11,7 +11,17 @@ function test_subctl_gather() {
   rm -rf $out_dir
   mkdir $out_dir
 
-  ${DAPPER_SOURCE}/bin/subctl gather --kubeconfig ${KUBECONFIGS_DIR}/kind-config-${cluster} --dir $out_dir
+  ${DAPPER_SOURCE}/bin/subctl gather --dir $out_dir
+
+  for cluster in "${clusters[@]}"; do
+      with_context "${cluster}" validate_gathered_files
+  done
+
+  # Broker
+  with_context "$broker" validate_broker_resources
+}
+
+function validate_gathered_files () {
 
   # connectivity
   validate_resource_files $subm_ns 'endpoints.submariner.io' 'Endpoint'
@@ -45,9 +55,6 @@ function test_subctl_gather() {
 
   validate_pod_log_files $subm_ns '-l component=submariner-lighthouse'
   validate_pod_log_files kube-system '-l k8s-app=kube-dns'
-
-  # Broker
-  with_context "$broker" validate_broker_resources
 }
 
 function validate_pod_log_files() {
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -23,11 +23,10 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	corev1 "k8s.io/api/core/v1"
-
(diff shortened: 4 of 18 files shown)


pkg/subctl/cmd/resource.go
```diff
@@ -42,7 +44,8 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
 
 	contexts := []string{}
 	if kubeContext != "" {
-		contexts = append(contexts, kubeContext)
+		kubeContextList := strings.Split(kubeContext, ",")
```
skitt: I would rather see a `StringSliceVar`, instead of handling this manually.

skitt (changes requested):
Would it be clearer to have a new flag, with a plural name? `kubecontexts`...

dfarrell07:
Maybe the commit message should be something like "Support gathering from multiple clusters"? Otherwise it seems like you're implementing Submariner in this PR. ;)

vthapar:
> Would it be clearer to have a new flag, with a plural name? `kubecontexts`...

I considered that, but adding two flags for context seemed superfluous and changing existing flag could break existing scripts. Other option could be to add new flag and deprecate old one but still support it for one release.

vthapar:
> Maybe the commit message should be something like "Support gathering from multiple clusters"? Otherwise it seems like you're implementing Submariner in this PR. ;)

Good call.

scripts/kind-e2e/lib_subctl_gather_test.sh
```diff
@@ -11,7 +11,7 @@ function test_subctl_gather() {
   rm -rf $out_dir
   mkdir $out_dir
 
-  ${DAPPER_SOURCE}/bin/subctl gather --kubeconfig ${KUBECONFIGS_DIR}/kind-config-${cluster} --dir $out_dir
+  ${DAPPER_SOURCE}/bin/subctl gather --dir $out_dir
```
tpantelis, line 14: Since we're now gathering from all clusters, we should verify each one. The validate functions below reference `${cluster}` which was being set by `with_context`. I suggest we put L16-47 below in a `validate_gathered_files` function and invoke via:

```
for cluster in "${clusters[@]}"; do
    with_context "${cluster}" validate_gathered_files
done
```

We could use `run_all_clusters` but that will do in parallel which will make the output harder to follow.

pkg/subctl/cmd/root.go
```diff
@@ -60,7 +60,7 @@ func init() {
 
 func addKubeconfigFlag(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&kubeConfig, "kubeconfig", "", "absolute path(s) to the kubeconfig file(s)")
-	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "kubeconfig context to use")
+	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "comma separated list of kubeconfig contexts to use")
```
tpantelis: Can we explain the default behavior? I assume "If not specified, all contexts referenced by kubeconfig are used"

tpantelis (changes requested):

vthapar:
> Would it be clearer to have a new flag, with a plural name? `kubecontexts`...

Added `kubec

pkg/subctl/cmd/root.go
```diff
@@ -60,7 +61,9 @@ func init() {
 
 func addKubeconfigFlag(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&kubeConfig, "kubeconfig", "", "absolute path(s) to the kubeconfig file(s)")
-	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "kubeconfig context to use")
+	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "DEPRECATED: use kubecontexts")
```
tpantelis:
```suggestion
	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "kubeconfig context to use")
	exitOnError("deprecation error", cmd.Flags().MarkDeprecated("kubecontext", "please use --kubecontexts instead"))
```

pkg/subctl/cmd/root.go
```diff
@@ -125,7 +128,10 @@ func getClientConfig(kubeConfigPath, kubeContext string) clientcmd.ClientConfig
 
 	rules.DefaultClientConfig = &clientcmd.DefaultClientConfig
 	overrides := &clientcmd.ConfigOverrides{ClusterDefaults: clientcmd.ClusterDefaults}
-	if kubeContext != "" {
+	if kubeContexts != nil {
```
tpantelis:
```suggestion
	if len(kubeContexts) > 0 {
```

tpantelis (changes requested):
@vthapar Were you going to change `getRestConfig` et al to not pass `kubeConfig` et al and just use the globals with this PR?

vthapar:
> @vthapar Were you going to change `getRestConfig` et al to not pass `kubeConfig` et al and just use the globals with this PR?

No, that would be different PR as that would require changes in every subctl subcommand that is not within scope of this. As part of this I am just going to use globals for my use.

pkg/subctl/cmd/root.go
vthapar: We can't do this right now coz it will break lots of existing scripts. `kubecontext` should be allowed for now but let the user know that it has been deprecated and should use the new `kubecontexts`

skitt:
> > Would it be clearer to have a new flag, with a plural name? `kubecontexts`...
> 
> I considered that, but adding two flags for context seemed superfluous and changing existing flag could break existing scripts. Other option could be to add new flag and deprecate old one but still support it for one release.

OK, let’s leave it as is for now, especially since the old form doesn’t change — `--kubecontext cluster1` will still work, we’re just adding support for `--kubecontext cluster1,cluster2`.

pkg/subctl/cmd/root.go
```diff
@@ -43,9 +43,10 @@ import (
 )
 
 var (
-	kubeConfig  string
-	kubeContext string
-	rootCmd     = &cobra.Command{
+	kubeConfig   string
+	kubeContext  string
+	kubeContexts []string
```
skitt, line 48: I think it would be simpler to share a single slice for both of these (changing the type of `--kubecontext`, but it won’t change anything for users).

skitt (changes requested):
OK, I see you added `--kubecontexts` which is good, but I have a suggestion there ;-).

pkg/subctl/cmd/root.go
vthapar: > I think it would be simpler to share a single slice for both of these (changing the type of `--kubecontext`, but it won’t change anything for users).

Changing `kubecontext` type would require change in every single caller. That is why I wanted to know why we even need to pass it as an arg when it is already a global var in `cmd`

skitt:
Vishal and I discussed this separately, here’s a summary:

* this PR will add `--kubecontexts` with documented support for multiple clusters, and no external change to `--kubecontext` (deprecation will be discussed later);
* to reduce the amount of change required in the implementations, both options will be `StringSliceVar`s using `kubeContexts`, and `kubeContext` will be initialised to `kubeContexts[0]` (so existing calls using `kubeContext` don’t have to be changed).

skitt:
The reason `getRESTConfig()` & co. take their input as parameters instead of relying on the global variables is that I prefer “real” functions with no dependencies on the context (otherwise they’re methods using a hidden “class” containing all the globals).

Regarding `--kubecontext`, my fear when considering replacing it wholesale with `--kubecontexts` is that the use of a plural would end up confusing users: it creates the expectation that every command supports multiple contexts. That would be a bigger change. _E.g._ if we document `subctl join --kubecontexts cluster1`, users could expect `subctl join --kubecontexts cluster1,cluster2` to work too, joining both clusters. We may want that ultimately but it’s outside this PR’s scope.

pkg/subctl/cmd/resource.go
```diff
@@ -41,7 +41,9 @@ func getMultipleRestConfigs(kubeConfigPath, kubeContext string) ([]restConfig, e
 	rules.ExplicitPath = kubeConfigPath
 
 	contexts := []string{}
-	if kubeContext != "" {
+	if kubeContexts != nil {
```
tpantelis:
```suggestion
	if len(kubeContexts) > 0 {
```

tpantelis (changes requested):

tpantelis (approved):

mangelajo (approved):

skitt (approved):

PR #1228, opened 2021-04-06
Mark globalnet as deprecated flag in deploy-broker command

maayanf24:
1. Since the flag `globalnet` is part of flag `components`, it should be deprecated
2. Remove defaults from `components` flag help string since it will be shown anyway

### DIFF
--- a/pkg/subctl/cmd/deploybroker.go
+++ b/pkg/subctl/cmd/deploybroker.go
@@ -52,6 +52,9 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
 		"enable support for Overlapping CIDRs in connecting clusters (default disabled)")
+
+	_ = deployBroker.PersistentFlags().MarkDeprecated("globalnet", "please use --components instead")
+
 	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "169.254.0.0/16",
 		"GlobalCIDR supernet range for allocating GlobalCIDRs to each cluster")
 	deployBroker.PersistentFlags().UintVar(&defaultGlobalnetClusterSize, "globalnet-cluster-size", 8192,
@@ -69,7 +72,7 @@ func init() {
 		"list of domains to use for multicluster service discovery")
 
 	deployBroker.PersistentFlags().StringSliceVar(&componentArr, "components", defaultComponents,
-		fmt.Sprintf("The components to be installed - any of %s. The default is all components",
+		fmt.Sprintf("The components to be installed - any of %s",
 			strings.Join(validComponents, ",")))
 
 	deployBroker.PersistentFlags().StringVar(&repository, "repository", "", "image repository")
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

pkg/subctl/cmd/deploybroker.go
```diff
@@ -52,6 +52,9 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
 		"enable support for Overlapping CIDRs in connecting clusters (default disabled)")
+
+	_ = deployBroker.PersistentFlags().MarkDeprecated("globalnet", "please use --components instead")
```
sridhargaddam: Shall we also give the proper usage?
```suggestion
	_ = deployBroker.PersistentFlags().MarkDeprecated("globalnet", "please use --components globalnet")
```

dfarrell07 (approved):

mangelajo (changes requested):
Agreed with @sridhargaddam comment

pkg/subctl/cmd/deploybroker.go
mangelajo: But, will that work? don't you also need to add all the other components too? (is --components additive?)

so probably it should be --components service-discovery,globalnet,operator ? ?? 

@skitt what would it be exactly ^?

mangelajo: I don't know I start to feel that we should keep this flag, so the --components thing is internal design, and --globalnet is part of the UI to make it easier?

sridhargaddam (approved):

pkg/subctl/cmd/deploybroker.go
skitt: I need to check but I think there are some edge-cases that don’t work too well with `--components globalnet` v. `--globalnet`. When I started working on full componentisation, it quickly became apparent too that Globalnet isn’t a component in the same way as connectivity and service discovery; it’s rather a setting which affects both... I’m not sure what the best way would be to represent it in the flags.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -52,6 +52,9 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
 		"enable support for Overlapping CIDRs in connecting clusters (default disabled)")
+
+	_ = deployBroker.PersistentFlags().MarkDeprecated("globalnet", "please use --components globalnet")
```
mangelajo, line 56: This will disable service-discovery & connectivity.

I believe we should not deprecate this flag until we have a good path forward.

mangelajo (changes requested):

pkg/subctl/cmd/deploybroker.go
tpantelis: Whether or not we remove the `globalnet` flag, just installing `globalnet` is invalid so perhaps we should remove it as a component. Or disallow `--components globalnet`.

manosnoam:
If changing `--globalnet` flag, it probably should be aligned with similar naming conventions where we call `--globalnet`.
For example, in the e2e framework:

https://github.com/submariner-io/shipyard/blob/85133a4b5261f3f59bf57c759e4c6bb39c328626/test/e2e/framework/test_context.go#L70

tpantelis:
Superseded by https://github.com/submariner-io/submariner-operator/pull/1269

PR #1233, opened 2021-04-08
Pull in optional 0.9 API fixes

skitt:
A couple of non-optional entries were added to the Submariner CRD,
which breaks upgrades. This updates to a fixed version of Submariner.

### DIFF
--- a/config/crd/bases/submariner.io_submariners.yaml
+++ b/config/crd/bases/submariner.io_submariners.yaml
@@ -373,8 +373,6 @@ spec:
                         - endpoint
                         - status
                         - statusMessage
-                        - usingIP
-                        - usingNAT
                         type: object
                       type: array
                     haStatus:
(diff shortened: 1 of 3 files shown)


mangelajo (approved):

skitt:
*THIS PR CAN’T BE MERGED AS-IS*. Its purpose is to verify https://github.com/submariner-io/submariner/pull/1246; this PR will need to be updated once that PR is merged (to use the merged hash).

mangelajo:
> _THIS PR CAN’T BE MERGED AS-IS_. Its purpose is to verify [submariner-io/submariner#1246](https://github.com/submariner-io/submariner/pull/1246); this PR will need to be updated once that PR is merged (to use the merged hash).

We have a chicken egg here ;) we may need to merge this one, then the other. Otherwise all E2E on the other is red.

skitt:
OK let’s merge this, if it passes, and I’ll submit a follow-up PR to adjust the hash.

mangelajo:
Thanks, as we were speaking on the slack, probably a good solution to this in the future would be re-compiling the operator in the submariner tree when there are changes to the submariner api.

tpantelis (approved):

PR #1242, opened 2021-04-09
Fix show connection IP output

mangelajo:
It would print out the PrivateIP regardless of what it's really used.
This also prints information on NAT traversal being detected or not.

The resulting output looks like:

```
GATEWAY                         CLUSTER                 REMOTE IP       NAT  CABLE DRIVER        SUBNETS                                 STATUS          
cluster2-worker                 cluster2                172.18.0.11     no   libreswan           100.2.0.0/16, 10.2.0.0/16               connecting      
cluster1-worker                 cluster1                172.18.0.6      no   libreswan           100.1.0.0/16, 10.1.0.0/16               connecting      

```

### DIFF
--- a/pkg/subctl/cmd/show_connections.go
+++ b/pkg/subctl/cmd/show_connections.go
@@ -29,6 +29,7 @@ type connectionStatus struct {
 	gateway     string
 	cluster     string
 	remoteIP    string
+	usingNAT    string
 	cableDriver string
 	subnets     string
 	status      submv1.ConnectionStatus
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

apis/submariner/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -22,7 +22,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
+	v1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
```
mangelajo: Oh this one sneaked in...

skitt (approved):

pkg/subctl/cmd/show_connections.go
```diff
@@ -58,10 +59,12 @@ func getConnectionsStatus(submariner *v1alpha1.Submariner) []connectionStatus {
 		for _, connection := range gateway.Connections {
 			subnets := strings.Join(connection.Endpoint.Subnets, ", ")
 
+			ip, nat := remoteIPAndNATForConnection(&connection)
```
skitt: The linter is complaining about this because taking the address of a range iterator is not good. To fix this, add a loop index and use `&gateway.Connections[i]` (and use `gateway.Connections[i]` throughout instead of `connection`).

mangelajo: Oh I thought I force pushed to fix it but I must have been too sleepy 

thanks for the explanation @skitt

maayanf24 (approved):

pkg/subctl/cmd/show_connections.go
mangelajo: I just passed the connection, without the pointer, so there's no need for the index, linter is passing now :)

PR #1245, opened 2021-04-12
Update the PR E2E strategy following the submariner repo

mangelajo:
We decided to use max version by default, and min version of K8s (1.17)
as an additional test to verify we didn't break the oldest one.

This saves CI jobs and avoids excessive queueing making it easier to
merge other patches.


SteveMattar (approved):

mkolesnik (approved):
LGTM

PR #1247, opened 2021-04-12
Generate tarball of gathered data

### LINKED ISSUE #1246: Generate tarball with all the gathered data
This is a subtask of EPIC  https://github.com/submariner-io/submariner-operator/issues/1125

Jaanki:
Create a tarball of the directory which containing
files with the data gathered using 'subctl gather'

### DIFF
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -17,8 +17,11 @@ limitations under the License.
 package cmd
 
 import (
+	"archive/tar"
 	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -139,6 +142,9 @@ func gatherDataByCluster(restConfig restConfig) {
 			}
 		}
 	}
+	info.Status.Start(fmt.Sprintf("Creating a tarball of gathered data in directory %s", directory))
+	createTar(directory, info.Status)
+	info.Status.End(info.Status.ResultFromMessages())
 }
 
 func gatherConnectivity(dataType string, info gather.Info) bool {
@@ -265,4 +271,54 @@ func getAllModuleKeys() []string {
 		keys = append(keys, k)
 	}
 	return keys
+}
+
+func createTar(sourcedir string, Status *cli.Status)  {
+	dir, err := os.Open(sourcedir)
+	if err != nil {
+		Status.QueueFailureMessage(fmt.Sprintf("error opening source directory %s", directory))
+	}
+	defer dir.Close()
+
+	files, err := dir.Readdir(0) // grab the files list
+	if err != nil {
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error grabbing files from directory %s", dir))
+		}
+	}
+
+	tarfile, err := os.Create(sourcedir + ".tar") // Destination tarfile name
+	if err != nil {
+		Status.QueueFailureMessage(fmt.Sprintf("error creating destination tarfile %s", tarfile))
+	}
+	defer tarfile.Close()
+
+	tarfileWriter := tar.NewWriter(tarfile)
+	defer tarfileWriter.Close()
+
+	for _, fileInfo := range files {
+		// see https://www.socketloop.com/tutorials/go-file-path-independent-of-operating-system
+		file, err := os.Open(dir.Name() + string(filepath.Separator) + fileInfo.Name())
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error opening file %s", file))
+		}
+		defer file.Close()
+
+		// prepare the tar header
+		header := &tar.Header{
+			Name: file.Name(),
+			Size: fileInfo.Size(),
+			Mode: int64(fileInfo.Mode())}
+
+		err = tarfileWriter.WriteHeader(header)
+		if err != nil {
+			Status.QueueFailureMessage("error writing tar header")
+		}
+
+		_, err = io.Copy(tarfileWriter, file)
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error copying file %s to tar", file))
+		}
+	}
+	Status.QueueSuccessMessage(fmt.Sprintf("Successfully created tarball %s", sourcedir + ".tar"))
 }


pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
 	}
 	return keys
 }
+
+func createTar(sourcedir string, Status *cli.Status)  {
+	dir, err := os.Open(sourcedir)
+	if err != nil {
+		Status.QueueFailureMessage(fmt.Sprintf("error opening source directory %s", directory))
```
tpantelis, line 279:
```suggestion
		Status.QueueFailureMessage(fmt.Sprintf("Error opening source directory %q: %v", directory, err))
		return
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
 	}
 	return keys
 }
+
+func createTar(sourcedir string, Status *cli.Status)  {
```
tpantelis, line 276:
```suggestion
func createTar(sourcedir string, cliStatus *cli.Status)  {
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
...
+	files, err := dir.Readdir(0) // grab the files list
+	if err != nil {
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error grabbing files from directory %s", dir))
+		}
+	}
+
+	tarfile, err := os.Create(sourcedir + ".tar") // Destination tarfile name
+	if err != nil {
+		Status.QueueFailureMessage(fmt.Sprintf("error creating destination tarfile %s", tarfile))
```
tpantelis, line 292:
```suggestion
		Status.QueueFailureMessage(fmt.Sprintf("Error creating destination tar file %q: %v", tarfile, err))
		return
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
...
+		if err != nil {
+			Status.QueueFailureMessage("error writing tar header")
+		}
+
+		_, err = io.Copy(tarfileWriter, file)
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error copying file %s to tar", file))
+		}
+	}
+	Status.QueueSuccessMessage(fmt.Sprintf("Successfully created tarball %s", sourcedir + ".tar"))
```
tpantelis, line 323:
```suggestion

	Status.QueueSuccessMessage(fmt.Sprintf("Successfully created tarball %q", tarFile.Name()))
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
...
+		Status.QueueFailureMessage(fmt.Sprintf("error opening source directory %s", directory))
+	}
+	defer dir.Close()
+
+	files, err := dir.Readdir(0) // grab the files list
+	if err != nil {
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error grabbing files from directory %s", dir))
+		}
+	}
```
tpantelis, line 288:
```suggestion
	if err != nil {
	    Status.QueueFailureMessage(fmt.Sprintf("Error reading files from directory %q: %v", dir, err))
	    return
	}
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
...
+	defer tarfile.Close()
+
+	tarfileWriter := tar.NewWriter(tarfile)
+	defer tarfileWriter.Close()
+
+	for _, fileInfo := range files {
+		// see https://www.socketloop.com/tutorials/go-file-path-independent-of-operating-system
+		file, err := os.Open(dir.Name() + string(filepath.Separator) + fileInfo.Name())
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error opening file %s", file))
```
tpantelis, line 303:
```suggestion
			Status.QueueFailureMessage(fmt.Sprintf("Error opening file %q: %v", file, err))
			return
```

pkg/subctl/cmd/gather.go
```diff
@@ -266,3 +272,53 @@ func getAllModuleKeys() []string {
...
+
+		err = tarfileWriter.WriteHeader(header)
+		if err != nil {
+			Status.QueueFailureMessage("error writing tar header")
+		}
+
+		_, err = io.Copy(tarfileWriter, file)
+		if err != nil {
+			Status.QueueFailureMessage(fmt.Sprintf("error copying file %s to tar", file))
+		}
```
tpantelis, line 321: It would be best to close each file after it's written to avoid keeping each file open at once. You can still use `defer` by putting this code in an anonymous function:

```suggestion
          err := func() error {
		defer file.Close()

		// prepare the tar header
		header := &tar.Header{
			Name: file.Name(),
			Size: fileInfo.Size(),
			Mode: int64(fileInfo.Mode())}

		err = tarfileWriter.WriteHeader(header)
		if err != nil {
		        return fmt.Sprintf("Error writing tar header: %v", err)
		}

		_, err = io.Copy(tarfileWriter, file)
		if err != nil {
		        return fmt.Sprintf("Error copying file %q to tar: %v", file.Name(), err)
		}
		
		return nil
	  }()
	  
	  if err != nil {
	      Status.QueueFailureMessage(err.Error())
	      return
	  }	
```

tpantelis (changes requested):

mangelajo (changes requested):
I think that creating a tar file manually doesn't add a lot of value. 

Can we call the command line tar cfz my-tar-file.gz submariner-directory-xxxx/  ?  instead

In that way it will also be gzip compressed. Gzip & Tar are everywhere today, and worst case, we leave a directory that the user will need to zip (if used under windows somehow..)

tpantelis:
> I think that creating a tar file manually doesn't add a lot of value.
> 
> Can we call the command line tar cfz my-tar-file.gz submariner-directory-xxxx/ ? instead
> 
> In that way it will also be gzip compressed. Gzip & Tar are everywhere today, and worst case, we leave a directory that the user will need to zip (if used under windows somehow..)

Good point - not much value in re-inventing the wheel. We could exec **tar** ourselves but it's easy enough for users to tar/zip themselves - perhaps we shouldn't do anything. In past products I've worked on we  just generated the files.

mangelajo:
> > I think that creating a tar file manually doesn't add a lot of value.
> > Can we call the command line tar cfz my-tar-file.gz submariner-directory-xxxx/ ? instead
> > In that way it will also be gzip compressed. Gzip & Tar are everywhere today, and worst case, we leave a directory that the user will need to zip (if used under windows somehow..)
> 
> Good point - not much value in re-inventing the wheel. We could exec **tar** ourselves but it's easy enough for users to tar/zip themselves - perhaps we shouldn't do anything. In past products I've worked on we just generated the files.

The good of providing a standard tar.gz / or tar.xz is that it's also easier to auto-process on our side, if we wanted/needed to. Also we make sure that customers will upload something compressed in all cases.

tar.xz is probably the most reduced in size for logs, but, the tar.gz combination is probably available everywhere (tar + gzip)

mangelajo:
In last meeting we decided to drop this PR & issue, as for openshift must-gather we simply request the operator to pack the files, we don't need to handle this.

https://docs.openshift.com/container-platform/4.5/support/gathering-cluster-data.html

PR #1248, opened 2021-04-13
Add globalnet+all-K8s-versions to flake finder

dfarrell07:
There seems to be some flaky failures with Globalnet and K8s 1.20.

There may also be other flakiness related to K8s versions, as we've just
recently started testing >1.17.


tpantelis (approved):

.github/workflows/flake_finder.yml
```diff
@@ -20,8 +20,14 @@ jobs:
         include:
           # Recentness of K8s versions are limited by kindest/node image releases
           - k8s_version: 1.18.15
+          - k8s_version: 1.18.15
```
mangelajo, line 23: may be it's a little bit "animal" but what do you think of adding the whole matrix of k8s versions in L18?

In such way we would also test lighthouse which could potentially have issues with different versions of the K8s API and their behaviors

Anyway, just up for discussion, in it's current form is the most urgent thing that we need to test.

mangelajo (approved):

.github/workflows/flake_finder.yml
dfarrell07: I think this makes sense, but maybe in https://github.com/submariner-io/lighthouse/pull/503?

PR #1251, opened 2021-04-15
Add staticcheck initialisms

Jaanki:
Add below initialisations in addition to the defaults

CNI
IPAM
IPSec
NAT
NATT

The defaults come from https://staticcheck.io/docs/#configuration

Signed-off-by: Janki Chhatbar jchhatba@redhat.com


mangelajo (dismissed):
You need to rebase this one on top of master

staticcheck.conf
```diff
@@ -0,0 +1,15 @@
...
+	"HTML", "HTTP", "HTTPS",
+	"ID", "IP", "IPAM",
+	"JSON",
+	"QPS",
+	"RAM", "RPC",
+	"SLA", "SMTP", "SQL", "SSH",
+	"TCP", "TLS", "TTL",
+	"UDP", "UI", "UID", "URI", "URL", "UTF8", "UUID",
+	"VM",
+	"XML", "XMPP", "XSRF", "XSS"]
```
mangelajo: See the comments on the submariner tree, I would add those at least too NAT / IP / IPSec

skitt:
```suggestion
initialisms = ["inherit", "CNI", "IPAM"]
```

skitt (changes requested):

skitt (approved):

nyechiel (approved):

.golangci.yml
```diff
@@ -66,6 +66,10 @@ issues:
     - linters:
         - stylecheck
       text: "ST1003: struct field BrokerK8sApiServer"
+    - path: apis/submariner/v1alpha1/submariner_types.go
+      text: "ST1003: struct field NatEnabled should be NATEnabled"
+    - path: pkg/subctl/cmd/join.go
+      text: "ST1003: var disableNat should be disableNAT"
```
tpantelis: Why are these 2 excluded?

tpantelis (changes requested):

.golangci.yml
Jaanki: `NATEnabled` is a Submariner spec variable. `disableNat` is a flag to `subctl join` and is deprecated so it would anyways be removed with https://github.com/submariner-io/submariner-operator/issues/1328

tpantelis: right but the variables can be changed in code to comply rather than excluded:

NatEnabled  -> NATEnabled
disableNat -> disableNAT

Even though #1328 will remove it at some point, this exclusion may be missed so either just rename it now or exclude in code via a _// nolint_ comment (or whatever the syntax is). I think the former is easiest.

.golangci.yml
```diff
@@ -66,6 +66,8 @@ issues:
     - linters:
         - stylecheck
       text: "ST1003: struct field BrokerK8sApiServer"
+    - path: .
+      text: "ST1003: struct field NatEnabled should be NATEnabled"
```
tpantelis:
```suggestion
      text: "ST1003: struct field NatEnabled"
```

tpantelis (approved):

sridhargaddam (approved):

dfarrell07 (approved):

PR #1252, opened 2021-04-15
Add deployment details to Submariner CR

### LINKED ISSUE #1250: Detect and store deployment information
Add fields to Submariner CR to store information about Kubernetes type [vanilla k8s, Openshift, EKS etc.] and underlying cloud provider [AWS, Azure etc.]. Operator should be able to detect and store this information at the time of deployment.

**Why is this needed**:
Though submariner works for all types of k8s deployments and cloud providers, there are few minor differences between them. This can be easily detected at the time of deployment, as we already detect for Openshift. We should detect and store this information to avoid needing to detect everytime we need to use this information. A good example is `subctl gather`.

vthapar:
Add information about deployment to Submariner and other relevant CRs.

Fixes #1250


apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -94,6 +95,15 @@ type DaemonSetStatus struct {
 	MismatchedContainerImages bool                     `json:"mismatchedContainerImages"`
 }
 
+type DeploymentInfo struct {
+	// +kubebuilder:validation:Enum=k8s;ocp;eks;aks;gke
```
mangelajo: Can we do it without the validations?

I'd leave it free form. There are more types, like microk8s, k3s, etc... and more that we don't know. If we apply too much validation then we will experience issues during upgrades I believe.

I belive it should be ok, because it's not an user-input, but instead a report that we make. We can add  constants in this file so we make sure that we will use those constants when filling up such fields

mangelajo (changes requested):

apis/submariner/v1alpha1/submariner_types.go
vthapar: Agree. I prefer it free form too but realized no one will have an opinion if I don't add validations.

mangelajo: Let's convert them to constants that we can use when reporting, and we can add more later :)

tpantelis: What about k3s?

mangelajo: Yeah, exactly, that's why I was proposing to avoid validation, as it's not user-input but status that we report.

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -94,6 +95,15 @@ type DaemonSetStatus struct {
...
+	KubernetesType        string `json:"kubernetesType,omitempty"`
+	KubernetesTypeVersion string `json:"kubernetesTypeVersion,omitempty"`
+	KubernetesVersion     string `json:"kubernetesVersion,omitempty"`
+	// +kubebuilder:validation:Enum=kind;aws;azure;gcp;openstack
```
mangelajo: same for this one, there are others too (alibaba...)

tpantelis (approved):

sridhargaddam (approved):

config/crd/bases/submariner.io_servicediscoveries.yaml
```diff
@@ -60,6 +60,17 @@ spec:
                 x-kubernetes-list-type: set
               debug:
                 type: boolean
+              deploymentInfo:
```
mkolesnik: Shouldn't this all be added under the `status` field?

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -102,7 +110,20 @@ type HealthCheckSpec struct {
...
+	KubernetesTypeOCP      = "ocp"
+	KubernetesTypeEKS      = "eks"
+	KubernetesTypeAKS      = "aks"
+	KubernetesTypeGKE      = "gke"
+	CloudProviderKind      = "kind"
+	CloudProviderAWS       = "aws"
+	CloudProviderGCP       = "gcp"
+	CloudProviderAzure     = "azure"
+	CloudProviderOpenstack = "openstack"
+)
```
mkolesnik:
```suggestion
type KubernetesType string
type CloudProvider string

const (
	DefaultColorCode                     = "blue"
	K8s                   KubernetesType = "k8s"
	OCP                                  = "ocp"
	EKS                                  = "eks"
	AKS                                  = "aks"
	GKE                                  = "gke"
	DefaultKubernetesType                = K8s
	Kind                  CloudProvider  = "kind"
	AWS                                  = "aws"
	GCP                                  = "gcp"
	Azure                                = "azure"
	Openstack                            = "openstack"
)
```

Perhaps something like this?

mkolesnik (changes requested):

config/crd/bases/submariner.io_servicediscoveries.yaml
vthapar: Only reason we put it in status in `Submariner` is that Operator can't udpate a Spec that it listens on. `Submariner` is created outside of operator [helm, user, subctl] so operator can only update Status fields. In case of ServiceDiscovery, it is created by Operator itself. Status in ServiceDiscovery would ideally host information generated by ServiceDiscovery controller.

apis/submariner/v1alpha1/submariner_types.go
vthapar: I considered this, but thought using more descriptive constant name will make it easy to understand what they're used for.

mangelajo: We also have "Kind" kubernetes type that we may want to add.

And... for cloud provider ir could also have vSphere, or even unknown?

apis/submariner/v1alpha1/servicediscovery_types.go
```diff
@@ -43,6 +43,7 @@ type ServiceDiscoverySpec struct {
 	// +listType=set
 	CustomDomains  []string          `json:"customDomains,omitempty"`
 	ImageOverrides map[string]string `json:"imageOverrides,omitempty"`
+	DeploymentInfo DeploymentInfo    `json:"deploymentInfo,omitempty"`
```
mangelajo: This should be in the Status, and not the Spec, right?

I see there is not status on service discovery, we either can create it or, not add it.

mangelajo (changes requested):

apis/submariner/v1alpha1/submariner_types.go
mkolesnik: Can you put it in it's own package? Then it's as explicit as you can get.

> We also have "Kind" kubernetes type that we may want to add.

Is it though? I think K8s "type" and Kind "cloud" are enough to understand its on KIND..

> 
> And... for cloud provider ir could also have vSphere, or even unknown?

config/crd/bases/submariner.io_servicediscoveries.yaml
mkolesnik: Seems inconsistent to me.

Also this is quite "status" information and not "desired state" so putting it in spec is just conceptually wrong.

apis/submariner/v1alpha1/servicediscovery_types.go
mkolesnik: IIUC it exists - https://github.com/submariner-io/submariner-operator/blob/8388f77a46d7b1d45fe2d7fc204149e85053b827/apis/submariner/v1alpha1/servicediscovery_types.go#L58

mkolesnik (changes requested):

tpantelis:
bump

tpantelis:
bump

mkolesnik (approved):

skitt (approved):

apis/submariner/v1alpha1/submariner_types.go
```diff
@@ -96,6 +97,13 @@ type DaemonSetStatus struct {
 	MismatchedContainerImages bool                     `json:"mismatchedContainerImages"`
 }
 
+type DeploymentInfo struct {
+	KubernetesType        KubernetesType `json:"kubernetesType,omitempty"`
+	KubernetesTypeVersion string         `json:"kubernetesTypeVersion,omitempty"`
+	KubernetesVersion     string         `json:"kubernetesVersion,omitempty"`
```
mangelajo, line 109: What's the difference between those two?

mangelajo: As explained in slack it's just for example the OCP version vs K8s version,although those should go hand in hand, and easy to co-relate, it could be useful to report them here in some cases.

mangelajo (approved):
Approved, but I think we could get rid of KubernetesTypeVersion for now (simplicity) and re-add it if necessary any time.

PR #1253, opened 2021-04-15
subctl gather: Capture CNI data

### LINKED ISSUE #1157: Capture CNI/driver data for troubleshooting
Collect CNI and cable driver specific information like IPTables, OVN flows etc. for troubleshooting datapath

**Why is this needed**:
Sub task of Epic #1125

vthapar:
When gathering `logs` for `connectivity`, capture the following

All CNI:
* `ip route show`          
* `ip rule list` 
* `ip rule show table 150`

IPTables CNI only:
* `iptables -L -n -v --line-numbers`
* `iptables -L -n -v --line-numbers -t nat`

Libreswan Cable Driver:
* `ip xfrm policy`
* `ip xfrm state`
* `ipsec status`
* `ipsec --trafficstatus`

Fixes #1157 

### DIFF
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -23,10 +23,14 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	submarinerOp "github.com/submariner-io/submariner-operator/apis/submariner/v1alpha1"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd/gather"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/components"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 var (
@@ -125,6 +129,24 @@ func gatherDataByCluster(restConfig restConfig) {
 	info.DynClient, info.ClientSet, err = getClients(restConfig.config)
 	exitOnError("Error getting client %s", err)
 
+	resourceSubmariners := schema.GroupVersionResource{
+		Group:    submarinerOp.SchemeGroupVersion.Group,
+		Version:  submarinerOp.SchemeGroupVersion.Version,
+		Resource: "submariners",
+	}
+
+	var submariner submarinerOp.Submariner
+
+	resourceSubm, err := info.DynClient.Resource(resourceSubmariners).Namespace(submarinerNamespace).Get("submariner", metav1.GetOptions{})
+	if err != nil {
+		exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
+	}
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(resourceSubm.UnstructuredContent(), &submariner)
+	if err != nil {
+		exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
+	}
+	info.Submariner = submariner
+
 	for module, ok := range gatherModuleFlags {
 		if ok {
 			for dataType, ok := range gatherTypeFlags {
@@ -148,6 +170,8 @@ func gatherConnectivity(dataType string, info gather.Info) bool {
 		gather.RouteAgentPodLogs(info)
 		gather.GlobalnetPodLogs(info)
 		gather.NetworkPluginSyncerPodLogs(info)
+		gather.CNIResources(info)
+		gather.CableDriverResources(info)
 	case Resources:
 		gather.Endpoints(info, SubmarinerNamespace)
 		gather.Clusters(info, SubmarinerNamespace)
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -63,6 +63,8 @@ var gatherFuncs = map[string]func(string, gather.Info) bool{
 	components.Operator:         gatherOperator,
 }
 
+var submarinerCR *submarinerOp.Submariner
+
 func init() {
 	addKubeconfigFlag(gatherCmd)
 	addKubecontextsFlag(gatherCmd)
@@ -135,17 +137,15 @@ func gatherDataByCluster(restConfig restConfig) {
 		Resource: "submariners",
 	}
 
-	var submariner submarinerOp.Submariner
-
 	resourceSubm, err := info.DynClient.Resource(resourceSubmariners).Namespace(submarinerNamespace).Get("submariner", metav1.GetOptions{})
-	if err != nil {
-		exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
+	if err == nil {
+		var submariner submarinerOp.Submariner
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(resourceSubm.UnstructuredContent(), &submariner)
+		if err != nil {
+			exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
+		}
+		submarinerCR = &submariner
 	}
-	err = runtime.DefaultUnstructuredConverter.FromUnstructured(resourceSubm.UnstructuredContent(), &submariner)
-	if err != nil {
-		exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
-	}
-	info.Submariner = submariner
 
 	for module, ok := range gatherModuleFlags {
 		if ok {
@@ -170,9 +170,11 @@ func gatherConnectivity(dataType string, info gather.Info) bool {
 		gather.RouteAgentPodLogs(info)
 		gather.GlobalnetPodLogs(info)
 		gather.NetworkPluginSyncerPodLogs(info)
-		gather.CNIResources(info)
-		gather.CableDriverResources(info)
 	case Resources:
+		if submarinerCR != nil {
+			gather.CNIResources(info, submarinerCR.Status.NetworkPlugin)
+			gather.CableDriverResources(info, submarinerCR.Spec.CableDriver)
+		}
 		gather.Endpoints(info, SubmarinerNamespace)
 		gather.Clusters(info, SubmarinerNamespace)
 		gather.Gateways(info, SubmarinerNamespace)
--- a/pkg/subctl/cmd/gather/cni.go
+++ b/pkg/subctl/cmd/gather/cni.go
@@ -19,7 +19,6 @@ package gather
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/resource"
 	v1 "k8s.io/api/core/v1"
 )
@@ -35,6 +34,10 @@ var ipCmds = map[string]string{
 	"ip-routes":         "ip route show",
 	"ip-rules":          "ip rule list",
 	"ip-rules-table150": "ip rule show table 150",
+}
+
+var ipGatewayCmds = map[string]string{
+	"ip-routes-table150": "ip route show table 150",
 }
 
 var ipTablesCmds = map[string]string{
@@ -58,8 +61,8 @@ var networkPluginCNIType = map[string]string{
 	"unknown":       typeUnknown,
 }
 
-func CNIResources(info Info) {
-	podLabelSelector := RouteagentPodLabel
+func CNIResources(info Info, networkPlugin string) {
+	podLabelSelector := routeagentPodLabel
 	err := func() error {
 		pods, err := findPods(info.ClientSet, podLabelSelector)
 
@@ -67,12 +70,13 @@ func CNIResources(info Info) {
 			return err
 		}
 
-		info.Status.QueueSuccessMessage(fmt.Sprintf("Found %d pods matching label selector %q", len(pods.Items), podLabelSelector))
+		info.Status.QueueSuccessMessage(fmt.Sprintf("athering CNI data from %d pods matching label selector %q",
+			len(pods.Items), podLabelSelector))
 
 		for i := range pods.Items {
 			pod := &pods.Items[i]
 			logIPCmds(info, pod)
-			switch networkPluginCNIType[info.Submariner.Status.NetworkPlugin] {
+			switch networkPluginCNIType[networkPlugin] {
 			case typeIPTables:
 				logIPTablesCmds(info, pod)
 			case typeOvn:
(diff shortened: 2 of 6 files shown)


pkg/subctl/cmd/gather.go
```diff
@@ -148,6 +170,8 @@ func gatherConnectivity(dataType string, info gather.Info) bool {
 		gather.RouteAgentPodLogs(info)
 		gather.GlobalnetPodLogs(info)
 		gather.NetworkPluginSyncerPodLogs(info)
+		gather.CNILogs(info)
```
mangelajo: I would consider all this resources instead, it's technically not logs.

Can we call them resources instead? :)

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,159 @@
...
+var networkPluginCNIType = map[string]string{
+	"generic":       typeIPTables,
+	"canal-flannel": typeIPTables,
+	"weave-net":     typeIPTables,
+	"OpenShiftSDN":  typeIPTables,
+	"OVNKubernetes": typeOvn,
+	"unknown":       typeUnknown,
+}
+
+func CNILogs(info Info) {
```
mangelajo:
```suggestion
func CNIResources(info Info) {
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,159 @@
...
+func logIPTablesCmds(info Info, pod *v1.Pod) {
+	for name, cmd := range ipTablesCmds {
+		err := logCmdOutput(info, pod, cmd, name)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("%q", err))
+		}
+	}
+}
+
+func CableDriverLogs(info Info) {
```
mangelajo:
```suggestion
func CableDriverResources(info Info) {
```

mangelajo (changes requested):
Awesome, just some nits.

pkg/subctl/cmd/gather.go
vthapar: I used logs because `resources` is a very k8s term that refers to `api-resources` or `CustomResource`

pkg/subctl/cmd/gather/cni.go
mangelajo: I probably should do the same for ovn OVNResources...

mangelajo:
Anyway ignore my nits we are busy busy %)

mangelajo (approved):

pkg/subctl/cmd/gather.go
```diff
@@ -125,6 +129,24 @@ func gatherDataByCluster(restConfig restConfig) {
...
+		Group:    submarinerOp.SchemeGroupVersion.Group,
+		Version:  submarinerOp.SchemeGroupVersion.Version,
+		Resource: "submariners",
+	}
+
+	var submariner submarinerOp.Submariner
+
+	resourceSubm, err := info.DynClient.Resource(resourceSubmariners).Namespace(submarinerNamespace).Get("submariner", metav1.GetOptions{})
+	if err != nil {
+		exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
```
Jaanki: This should not `exitOnError` but print warning message that subm resource is not found. Otherwise, it skips collecting other resources from cluster where subm is not installed eg broker

```
$ ./../submariner-operator/bin/subctl gather module=connectivity
**Gathering information from cluster "cluster2"**
 ✓ Gathering connectivity logs 
....
 ✓ Found 1 deployments by label selector "app=submariner-lighthouse-coredns" in namespace "submariner-operator"
**Gathering information from cluster "cluster3"**
 ✓ Gathering operator logs 
...
**Gathering information from cluster "cluster1"**
panic: runtime error: invalid memory address or nil pointer dereference
[signal] SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14b553a]
```

For me, cluster1 is broker cluster and it should still collect information like `Endpoints` and  `Clusters`.

I understand that broker resources are collected while iterating over dataclusters but that is something that developer knows. User gets confused with the `exitOnError` response.

Jaanki (dismissed):

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+	typeIPTables = "iptables"
+	typeOvn      = "ovn"
+	typeUnknown  = "unknown"
+	libreswan    = "libreswan"
+)
+
+var ipCmds = map[string]string{
+	"ip-routes":         "ip route show",
+	"ip-rules":          "ip rule list",
+	"ip-rules-table150": "ip rule show table 150",
```
sridhargaddam, line 36:
```suggestion
	"ip-rules-table150": "ip route show table 150",
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	typeIPTables = "iptables"
+	typeOvn      = "ovn"
+	typeUnknown  = "unknown"
+	libreswan    = "libreswan"
+)
+
+var ipCmds = map[string]string{
```
sridhargaddam, line 33: We would need ip-routes from all the route-agent pods, but the other two commands (ip-rules and ip-rules-table150) are generally only present on active Gateway node. Anyways, there is no harm in collecting them - might also help us to know if there are any cleanup issues during gateway migration.

sridhargaddam (dismissed):
Nice work @vthapar. I tried the PR and its working great!!

pkg/subctl/cmd/gather.go
tpantelis: There is a `getSubmarinerResource` function that does this. However we shouldn't get it here and error if not found. In addition to what @Jaanki mentioned, In a service-discovery-only install there is no `Submariner` resource. So we should obtain it where needed and warn/skip appropriately if not found.

pkg/subctl/cmd/gather.go
```diff
@@ -148,6 +170,8 @@ func gatherConnectivity(dataType string, info gather.Info) bool {
 		gather.RouteAgentPodLogs(info)
 		gather.GlobalnetPodLogs(info)
 		gather.NetworkPluginSyncerPodLogs(info)
+		gather.CNIResources(info)
+		gather.CableDriverResources(info)
```
tpantelis: Shouldn't these be under the `Resources` type?

pkg/subctl/cmd/gather/connectivity.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	gatewayPodLabel             = "app=submariner-gateway"
-	routeagentPodLabel          = "app=submariner-routeagent"
+	RouteagentPodLabel          = "app=submariner-routeagent"
```
tpantelis: Why was this exported? I don't see it used outside the package.

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+func CNIResources(info Info) {
+	podLabelSelector := RouteagentPodLabel
+	err := func() error {
+		pods, err := findPods(info.ClientSet, podLabelSelector)
+
+		if err != nil {
+			return err
+		}
+
+		info.Status.QueueSuccessMessage(fmt.Sprintf("Found %d pods matching label selector %q", len(pods.Items), podLabelSelector))
```
tpantelis: I think we can be more specific here to indicate it's gathering CNI data.

```suggestion
		info.Status.QueueSuccessMessage(fmt.Sprintf("Gathering CNI data from %d pods matching label selector %q", len(pods.Items), podLabelSelector))
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+func logCmdOutput(info Info, pod *v1.Pod, cmd, cmdName string) error {
+	execOptions := resource.ExecOptionsFromPod(pod)
+	execConfig := resource.ExecConfig{
+		RestConfig: info.RestConfig,
+		ClientSet:  info.ClientSet,
+	}
+	execOptions.Command = []string{"/bin/bash", "-c", cmd}
+	stdOut, _, err := resource.ExecWithOptions(execConfig, execOptions)
+	if err != nil {
+		return errors.WithMessagef(err, "error running %q on pod %s : %q", cmd, pod.Name, err)
```
tpantelis, line 153: I think we should just log the failure message here as we have all the context and not return error.

```suggestion
               info.Status.QueueFailureMessage(fmt.Sprintf("Error running %q on pod %q: %v", cmd, pod.Name, err))
               return
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if stdOut != "" {
+		// the first line contains the executed command
+		stdOut = cmd + "\n" + stdOut
+		return writeLogToFile(stdOut, pod.Spec.NodeName+"_"+cmdName, info)
```
tpantelis, line 158:
```suggestion
                 err := writeLogToFile(stdOut, pod.Spec.NodeName+"_"+cmdName, info)
                 if err != nil {
                      info.Status.QueueFailureMessage(fmt.Sprintf("Error writing output from command %q on pod %q: %v", cmd, pod.Name, err))
                  }
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+func CableDriverResources(info Info) {
+	podLabelSelector := gatewayPodLabel
+	err := func() error {
+		pods, err := findPods(info.ClientSet, podLabelSelector)
+
+		if err != nil {
+			return err
+		}
+
+		info.Status.QueueSuccessMessage(fmt.Sprintf("Found %d pods matching label selector %q", len(pods.Items), podLabelSelector))
```
tpantelis:
```suggestion
		info.Status.QueueSuccessMessage(fmt.Sprintf("Gathering cable driver data from %d pods matching label selector %q", len(pods.Items), podLabelSelector))
```

pkg/subctl/cmd/gather/types.go
```diff
@@ -29,4 +30,5 @@ type Info struct {
 	DynClient   dynamic.Interface
 	ClusterName string
 	DirName     string
+	Submariner  submariner.Submariner
```
tpantelis: Since this isn't use by all components, I think it should just be stored and passed locally.

tpantelis (changes requested):

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+			info.Status.QueueFailureMessage(fmt.Sprintf("%q", err))
+		}
+	}
+}
+
+func logIPTablesCmds(info Info, pod *v1.Pod) {
+	for name, cmd := range ipTablesCmds {
+		err := logCmdOutput(info, pod, cmd, name)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("%q", err))
```
tpantelis:
```suggestion
			info.Status.QueueFailureMessage(fmt.Sprintf("Error running IP tables command: %v", err))
```

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,161 @@
...
+		info.Status.QueueFailureMessage(fmt.Sprintf("Failed to gather CNI data from pods matching label selector %q: %s",
+			podLabelSelector, err))
+	}
+}
+
+func logIPCmds(info Info, pod *v1.Pod) {
+	for name, cmd := range ipCmds {
+		err := logCmdOutput(info, pod, cmd, name)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("%q", err))
```
tpantelis:
```suggestion
			info.Status.QueueFailureMessage(fmt.Sprintf("Error running IP command: %v", err))
```

tpantelis (dismissed):

pkg/subctl/cmd/gather.go
mangelajo: Oh good point

pkg/subctl/cmd/gather/cni.go
mangelajo: Oh sorry, I asked him to change that, we need both.  OVN uses rules in table 150.

mangelajo: @vthapar can we add both?

pkg/subctl/cmd/gather/cni.go
mangelajo: yeah, let's keep them just in case :)

pkg/subctl/cmd/gather.go
mangelajo: Yeah, I agree, better to make them resource types.

pkg/subctl/cmd/gather.go
vthapar: I'll remove this error. This code will be reworked once https://github.com/submariner-io/submariner-operator/pull/1252 and subsequent patches are merged. For some of LH gathering, we need serviceDiscovery CR.

EDIT: Also, we need to revisit and optimize gathering broker resources. It should be possible to mark out broker cluster and not pull this information for every cluster.

pkg/subctl/cmd/gather/connectivity.go
vthapar: Missed this cleanup after some refactoring.

pkg/subctl/cmd/gather/types.go
vthapar: Ack. Was in two minds regarding this. Will make it local for now and add if later more components need this.

pkg/subctl/cmd/gather/cni.go
vthapar: Yeah, on rest it will capture no output and we will not log it. It is easier and faster to just try grab on all then figure out which is which.

pkg/subctl/cmd/gather/cni.go
```diff
@@ -0,0 +1,191 @@
...
+func logCNIGatewayNodeResources(info Info) {
+	podLabelSelector := gatewayPodLabel
+	err := func() error {
+		pods, err := findPods(info.ClientSet, podLabelSelector)
+
+		if err != nil {
+			return err
+		}
+
+		info.Status.QueueSuccessMessage(fmt.Sprintf("athering CNI data from %d pods matching label selector %q",
```
sridhargaddam:
```suggestion
		info.Status.QueueSuccessMessage(fmt.Sprintf("Gathering CNI data from %d pods matching label selector %q",
```

sridhargaddam (commented):
Looks good @vthapar. One small nit.

pkg/subctl/cmd/gather/cni.go
mangelajo: Please let's keep `ip rule show table 150`   it will be useful for OVN on gateways.

mangelajo (approved):

Jaanki (approved):

PR #1254, opened 2021-04-15
Gather CNI OVN data

mangelajo:
Gather OVN data as part of the CNI resources

Depends on #1253

### DIFF
--- a/pkg/subctl/cmd/gather/cni.go
+++ b/pkg/subctl/cmd/gather/cni.go
@@ -52,6 +52,19 @@ var libreswanCmds = map[string]string{
 	"ipsec-trafficstatus": "ipsec --trafficstatus",
 }
 
+const ovnShowCmd = "ovn-nbctl show"
+
+var ovnCmds = map[string]string{
+	"ovn_show":             ovnShowCmd,
+	"ovn_logical_routers":  "ovn-nbctl list Logical_Router",
+	"ovn_lrps":             "ovn-nbctl list Logical_Router_Port",
+	"ovn_logical_switches": "ovn-nbctl list Logical_Switch",
+	"ovn_lsps":             "ovn-nbctl list Logical_Switch_Port",
+	"ovn_routes":           "ovn-nbctl list Logical_Router_Static_Route",
+	"ovn_policies":         "ovn-nbctl list Logical_Router_Policy",
+	"ovn_acls":             "ovn-nbctl list ACL",
+}
+
 var networkPluginCNIType = map[string]string{
 	"generic":       typeIPTables,
 	"canal-flannel": typeIPTables,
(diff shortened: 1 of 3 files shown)


pkg/subctl/cmd/gather/cni.go
```diff
@@ -107,6 +121,43 @@ func logIPTablesCmds(info Info, pod *v1.Pod) {
 	}
 }
 
+func OVNData(info Info) {
+	if networkPluginCNIType[info.Submariner.Status.NetworkPlugin] != typeOvn {
+		return
+	}
+
+	pods, err := findPods(info.ClientSet, ovnMasterPodLabelOCP)
```
sridhargaddam, line 159: nit: Can you please add some notes about why we try with two labels? Might help us to know when to remove the additional check in future.

pkg/subctl/cmd/gather/cni.go
```diff
@@ -107,6 +121,43 @@ func logIPTablesCmds(info Info, pod *v1.Pod) {
...
+			info.Status.QueueFailureMessage("Failed to gather any OVN master pods: " + err.Error())
+		}
+	}
+
+	var masterPod *v1.Pod
+	// ovn-nbctl commands only work on one of the masters, figure out which one
+	for i := range pods.Items {
+		err = logCmdOutput(info, &pods.Items[i], ovnShowCmd, ovnShowName)
+		if err == nil {
+			masterPod = &pods.Items[i]
```
sridhargaddam, line 174: break

pkg/subctl/cmd/gather/cni.go
```diff
@@ -107,6 +121,43 @@ func logIPTablesCmds(info Info, pod *v1.Pod) {
...
+	}
+
+	for name, command := range ovnCmds {
+		err = logCmdOutput(info, masterPod, command, name)
+		if err != nil {
+			info.Status.QueueFailureMessage(fmt.Sprintf("Failed to exec OVN command: %s", err))
+		}
+	}
+
+	info.Status.QueueSuccessMessage("Successfully gathered OVN details")
```
sridhargaddam: Probably we can do something like 

```
	if status.HasFailureMessages() {
		status.End(cli.Failure)
		return
	}
        info.Status.QueueSuccessMessage("Successfully gathered OVN details")
        status.End(cli.Success)
```

sridhargaddam (changes requested):
Thanks for the PR @mangelajo. Some small nits.

pkg/subctl/cmd/gather/cni.go
```diff
@@ -107,6 +121,43 @@ func logIPTablesCmds(info Info, pod *v1.Pod) {
 	}
 }
 
+func OVNData(info Info) {
+	if networkPluginCNIType[info.Submariner.Status.NetworkPlugin] != typeOvn {
+		return
+	}
+
+	pods, err := findPods(info.ClientSet, ovnMasterPodLabelOCP)
+	if err != nil && apiErrors.IsNotFound(err) {
```
tpantelis: `findPods` won't return `IsNotFound`. Perhaps you want to check `len(pods) > 0`.

pkg/subctl/cmd/gather/cni.go
tpantelis: I don't think we want to end the status since it's shared. Also there could be failure messages from prior components. It might be better for each component to create its own status but that would require a refactor. I suggest following what we've done for other components, ie log what it will be doing but no specific success message at the end. Any subsequent failure messages indicate it failed.

So above L151:

`info.Status.QueueSuccessMessage(fmt.Sprintf("Gathering OVN data from master pod %", masterPod.Name))`

tpantelis (changes requested):

pkg/subctl/cmd/gather/cni.go
Jaanki: len(pods) is checked in `findponds`. If len = 0 , it returns an error which we are already checking.

tpantelis: No it doesn't

```
func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.PodList, error) {
        pods, err := clientSet.CoreV1().Pods("").List(metav1.ListOptions{LabelSelector: byLabelSelector})

        if err != nil {
                return nil, errors.WithMessage(err, "error listing pods")
        }

        return pods, nil
}

```

Jaanki: ohh..that check was removed by a subsequent commit. So yes, the check is needed

Jaanki: https://github.com/submariner-io/submariner-operator/commit/a38930deb6c34ae556954ae9183fff3d58a7bcbd

pkg/subctl/cmd/gather/cni.go
mangelajo: Ah, I'll comment, It's because OCP uses different labels compared to ovn-kubernetes upstream.

pkg/subctl/cmd/gather/cni.go
mangelajo: right

pkg/subctl/cmd/gather/cni.go
mangelajo: List doesn't return IsNotFound? :? Ok  I wanted to make sure we didn't exit because of an error, let me check.

tpantelis: Well it can return IsNotFound if the resource kind itself doesn't exist. But we're looking up pods so that won't happen. In this case you want to know if any pods instances were found.

mangelajo:
Updated, thanks for all the comments & reviews.

sridhargaddam (approved):
Thanks @mangelajo

tpantelis (approved):

mangelajo:
rebasing this on latest work by vthapar.

Jaanki (approved):

mangelajo:
rebase done

PR #1261, opened 2021-04-19
Refactor retrieval/storage of Submariner resource for gather

tpantelis:


pkg/subctl/cmd/gather.go
```diff
@@ -129,22 +124,15 @@ func gatherDataByCluster(restConfig restConfig) {
...
-	resourceSubm, err := info.DynClient.Resource(resourceSubmariners).Namespace(submarinerNamespace).Get("submariner", metav1.GetOptions{})
-	if err == nil {
-		var submariner submarinerOp.Submariner
-		err = runtime.DefaultUnstructuredConverter.FromUnstructured(resourceSubm.UnstructuredContent(), &submariner)
-		if err != nil {
-			exitOnError(fmt.Sprintf("Failed to get submariner resource %q", err), err)
-		}
-		submarinerCR = &submariner
+	info.Submariner, err = getSubmarinerResourceWithError(restConfig.config)
+	if err != nil && !apierrors.IsNotFound(err) {
```
vthapar, line 133: Submariner CRD is not installed on broker, iirc. If CRD isn't even present, we will not get IsNotFound but a different error. In case of ServiceDiscovery only deployments, it might be present. Either ways, `IsNotFound()` by itself is not reliable enough.

EDIT: Nvm, tested and looks like we do get IsNotFound error

vthapar (changes requested):

vthapar (approved):

pkg/subctl/cmd/gather.go
mangelajo: Yes, to install the operator we install both CRDs, then you create the CRs or not..

pkg/subctl/cmd/gather.go
```diff
@@ -171,10 +159,10 @@ func gatherConnectivity(dataType string, info gather.Info) bool {
...
 		gather.NetworkPluginSyncerPodLogs(info)
 	case Resources:
-		if submarinerCR != nil {
-			gather.CNIResources(info, submarinerCR.Status.NetworkPlugin)
-			gather.CableDriverResources(info, submarinerCR.Spec.CableDriver)
-			gather.OVNResources(info, submarinerCR.Status.NetworkPlugin)
+		if info.Submariner != nil {
+			gather.CNIResources(info, info.Submariner.Status.NetworkPlugin)
+			gather.CableDriverResources(info, info.Submariner.Spec.CableDriver)
+			gather.OVNResources(info, info.Submariner.Status.NetworkPlugin)
```
mangelajo, line 165:
```suggestion
			gather.CNIResources(info)
			gather.CableDriverResources(info)
			gather.OVNResources(info)
```

Then inside we can access info.Submariner......

mangelajo (approved):
I'll send a follow up on my comments ;)

PR #1272, opened 2021-04-21
Capture sysctl -a as part of the route agent logs in gather

mangelajo:
This is a critical part of the dataplane configuration in Linux
that we need to check for various details.

Depends on submariner-io/submariner#1288

### DIFF
--- a/pkg/subctl/cmd/gather/cni.go
+++ b/pkg/subctl/cmd/gather/cni.go
@@ -36,6 +36,7 @@ var systemCmds = map[string]string{
 	"ip-routes":         "ip route show",
 	"ip-rules":          "ip rule list",
 	"ip-rules-table150": "ip rule show table 150",
+	"sysctl-a":          "sysctl -a",
 }
 
 var ipGatewayCmds = map[string]string{


mangelajo:
Wait, the images are missing the sysctl tool:

We need to install procps-ng on the routeagent images.

mkolesnik:
> Wait, the images are missing the sysctl tool:
> 
> We need to install procps-ng on the routeagent images.

Won't it make more sense to use a sidecar container for this purpose?

aswinsuryan:
> > Wait, the images are missing the sysctl tool:
> > We need to install procps-ng on the routeagent images.
> 
> Won't it make more sense to use a sidecar container for this purpose?

@mkolesnik You mean the entire gather commands should be implemented using a sidecar container? If so I think it would be a bigger change for 0.9.0

mangelajo:
Yes, we need procps-ng installed on the route-agent images.

I agree, we could do this as a sidecar, we have other tools that we can use for diagnosing too (may be with the net-test image).

It would be a bigger change in gather  (compared to this one-line + pkg install)

For 0.9 I think it's ok to go with procps-ng in route agent, and then we can optimize for 0.10
```
Total download size: 338 k
Installed size: 958 k
Downloading Packages:
procps-ng-3.3.16-2.fc33.x86_64.rpm                                                                                                                                                                            884 kB/s | 338 kB     00:00    
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                                                         427 kB/s | 338 kB     00:00   
```

I prefer easier support to customers vs saving 338KB of compressed image (1MB of uncompressed...).
@mkolesnik @nyechiel  @skitt @aswinsuryan @vthapar @dfarrell07  @sridhargaddam  @maayanf24 @Jaanki what do you think?

mangelajo:
> > Wait, the images are missing the sysctl tool:
> > We need to install procps-ng on the routeagent images.
> 
> Won't it make more sense to use a sidecar container for this purpose?

So technically I agree, we could deploy a daemonset (since we need it on all nodes) with netutils, and then use sysctl from there, or any other tool that we need for gathering additional stuff. Just seems risky for a last minute addition.

aswinsuryan:
> I prefer easier support to customers vs saving 338KB of compressed image (1MB of uncompressed...).

Ya I agree,  quick fix would be a better idea for 0.9.0 and use a sidecar later.

Jaanki:
> I prefer easier support to customers vs saving 338KB of compressed image (1MB of uncompressed...).

I agree. And this is just for route agent pods so installing the pkg to routeagent image seems a better approach at the moment. We can then revisit this in 0.10.0 if needed.

aswinsuryan:
Raised this one just  in case if we are going ahead with the current approach 

https://github.com/submariner-io/submariner/pull/1288

Jaanki:
Depends on https://github.com/submariner-io/submariner/pull/1288

In case needed

mangelajo:
@aswinsuryan @Jaanki @tpantelis @skitt this one is ready now.

nyechiel (approved):

tpantelis (approved):

Jaanki (approved):

mangelajo:
+1 on my side to override the merge, as the failing job is:
```

[submariner-operator]$ [cluster2] find /go/src/github.com/submariner-io/submariner-operator/output/kubeconfigs -type f -printf %p:
[submariner-operator]$ [cluster2] bash -c curl -Ls https://get.submariner.io | VERSION=devel DESTDIR=/go/bin bash
Installing subctl version devel
  OS detected:           linux
  Architecture detected: amd64
  Download URL:          https://github.com/submariner-io/submariner-operator/releases/download/subctl-devel/subctl-devel-linux-amd64.tar.xz

Downloading...
curl: (22) The requested URL returned error: 404 
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now

```

And this change does not introduce any functional change to normal submariner, only to gather.

mangelajo:
Thanks for the rebase @tpantelis

PR #1287, opened 2021-04-22
subctl gather: Print output directory

### LINKED ISSUE #1280: subctl gather should print the directory in which logs are stored.
**What happened**:
Currently when `subctl gather` is executed, it does not display the folder name in which the logs are stored. It would be very helpful (particularly while running the gather command multiple times) to include a message about the folder name.

tpantelis:
Fixes #1280

### DIFF
--- a/pkg/subctl/cmd/gather.go
+++ b/pkg/subctl/cmd/gather.go
@@ -112,6 +112,8 @@ func gatherData() {
 	for _, config := range configs {
 		gatherDataByCluster(config, directory)
 	}
+
+	fmt.Printf("Files are stored under directory %q\n", directory)
 }
 
 func gatherDataByCluster(restConfig restConfig, directory string) {


nyechiel (approved):

Jaanki (commented):
Or should be print this at the end of execution with the message "SUccessfully gathered info at <directory>"

pkg/subctl/cmd/gather.go
```diff
@@ -102,6 +102,8 @@ func gatherData() {
 		directory = "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
 	}
 
+	fmt.Printf("Files will be stored under directory %q\n", directory)
```
vthapar: Would be better to have this at the end with a msg like `Files stored in directory xyz`. Gather output can get very long with multiple clusters and it is always best to have user relevant output at the end so user doesn't need to scroll back to look it up.

vthapar (changes requested):

pkg/subctl/cmd/gather.go
mangelajo: I vote for the end.

As long as our error control is tight and we won't stop mid-flight.

vthapar (approved):

sridhargaddam (approved):

mangelajo (approved):

PR #1290, opened 2021-04-22
Capture logs of crashed pod

### LINKED ISSUE #1289: subctl gather will not capture logs for crashed pods
capture logs for crashed pods which can be captured with the -p flag when using kubectl logs

Jaanki:
Capture logs of crashed pod

by using Previous (-p) option of PodLogOptions

The logic implemented is

```
Capture the logs with -p
if previous_pod_found
    write_logs_to_file.log.prev
else
   ignore

Capture the logs without -p on the same pod
write_logs_to_file.log
```

Closes: #1289

### DIFF
--- a/pkg/subctl/cmd/gather/logs.go
+++ b/pkg/subctl/cmd/gather/logs.go
@@ -26,7 +26,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 func gatherPodLogs(podLabelSelector string, info Info) {
@@ -43,14 +42,10 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
 
 		info.Status.QueueSuccessMessage(fmt.Sprintf("Found %d pods matching label selector %q", len(pods.Items), podLabelSelector))
 
-		podLogOptions := &corev1.PodLogOptions{}
+		podLogOptions := corev1.PodLogOptions{}
 		podLogOptions.Container = container
 		for i := range pods.Items {
-			pod := &pods.Items[i]
-			err := podLogsToFile(pod, podLogOptions, info)
-			if err != nil {
-				return errors.WithMessagef(err, "error getting logs for pod %q", pod.Name)
-			}
+			outputPodLogs(&pods.Items[i], podLogOptions, info)
 		}
 
 		return nil
@@ -62,16 +57,26 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
 	}
 }
 
-func podLogsToFile(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
-	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+func outputPodLogs(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) {
+	err := outputPreviousPodLog(pod, podLogOptions, info)
+	if err != nil {
+		info.Status.QueueFailureMessage(fmt.Sprintf("Error outputting previous log for pod %q: %v", pod.Name, err))
+	}
 
-	logs, err := processLogStream(logRequest)
+	err = outputCurrentPodLog(pod, podLogOptions, info)
+	if err != nil {
+		info.Status.QueueFailureMessage(fmt.Sprintf("Error outputting current log for pod %q: %v", pod.Name, err))
+	}
+}
+
+func writePodLogToFile(logStream io.ReadCloser, info Info, podName, fileExtension string) error {
+	logs, err := getLogFromStream(logStream)
 	if err != nil {
 		return err
 	}
 
 	logs = scrubSensitiveData(info, logs)
-	err = writeLogToFile(logs, pod.Name, info)
+	err = writeLogToFile(logs, podName, info, fileExtension)
 	if err != nil {
 		return err
 	}
@@ -79,15 +84,9 @@ func podLogsToFile(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info In
 	return nil
 }
 
-func processLogStream(logrequest *rest.Request) (string, error) {
-	logStream, err := logrequest.Stream()
-	if err != nil {
-		return "", errors.WithMessage(err, "error opening log stream")
-	}
-	defer logStream.Close()
-
+func getLogFromStream(logStream io.ReadCloser) (string, error) {
 	logs := new(bytes.Buffer)
-	_, err = io.Copy(logs, logStream)
+	_, err := io.Copy(logs, logStream)
 	if err != nil {
 		return "", errors.WithMessage(err, "error copying the log stream")
 	}
@@ -95,8 +94,8 @@ func processLogStream(logrequest *rest.Request) (string, error) {
 	return logs.String(), nil
 }
 
-func writeLogToFile(data, podName string, info Info) error {
-	fileName := filepath.Join(info.DirName, info.ClusterName+"_"+podName+".log")
+func writeLogToFile(data, podName string, info Info, fileExtension string) error {
+	fileName := filepath.Join(info.DirName, info.ClusterName+"_"+podName+fileExtension)
 	f, err := os.Create(fileName)
 	if err != nil {
 		return errors.WithMessagef(err, "error opening file %s", fileName)
@@ -119,4 +118,40 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
 	}
 
 	return pods, nil
+}
+
+func outputPreviousPodLog(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) error {
+	podLogOptions.Previous = true
+	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &podLogOptions)
+	logStream, _ := logRequest.Stream()
+
+	// TODO: Check for error other than "no previous pods found"
+
+	// if no previous pods found, logstream == nil, ignore it
+	if logStream != nil {
+		info.Status.QueueWarningMessage(fmt.Sprintf("Found logs for previous instances of pod %s", pod.Name))
+		err := writePodLogToFile(logStream, info, pod.Name, ".log.prev")
+		if err != nil {
+			return err
+		}
+		defer logStream.Close()
+	}
+	return nil
+}
+
+func outputCurrentPodLog(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) error {
+	// Running with Previous = false on the same pod
+	podLogOptions.Previous = false
+	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &podLogOptions)
+	logStream, err := logRequest.Stream()
+	if err != nil {
+		return errors.WithMessage(err, "error opening log stream")
+	}
+	defer logStream.Close()
+
+	err = writePodLogToFile(logStream, info, pod.Name, ".log")
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/pkg/subctl/cmd/gather/cni.go
+++ b/pkg/subctl/cmd/gather/cni.go
@@ -245,7 +245,7 @@ func logCmdOutput(info Info, pod *v1.Pod, cmd, cmdName string, ignoreError bool)
 	if stdOut != "" {
 		// the first line contains the executed command
 		stdOut = cmd + "\n" + stdOut
-		err := writeLogToFile(stdOut, pod.Spec.NodeName+"_"+cmdName, info)
+		err := writeLogToFile(stdOut, pod.Spec.NodeName+"_"+cmdName, info, ".log")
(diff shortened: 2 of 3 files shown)


pkg/subctl/cmd/gather/logs.go
```diff
@@ -57,7 +57,7 @@ func gatherPodLogs(podLabelSelector string, info Info) {
 }
 
 func podLogsToFile(pod *corev1.Pod, info Info) error {
-	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{})
+	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{Previous:true})
```
mangelajo: We should only use Previous:true if the request with previous:false fails.

mangelajo: I guess that we need to try L62 with previous=false, ... if that fails, then we try again with Previous:true

mangelajo (changes requested):

mangelajo:
@Jaanki thanks for looking at this so quickly!

pkg/subctl/cmd/gather/logs.go
vthapar: Why not get both? We could have scenarios where a pod crashed, caused some failures but then recovered. In such cases we will want old and current logs both.

Jaanki:
GW pod in my setup is in `CrashLoopBackOff`. This patch generates a `.log.old` file with logs from previous pod and `.log` file with logs from current pod.

Attached are the logs files generated on running `subctl gather --module=connectivity --type=logs` 

[submariner-20210423170957.zip](https://github.com/submariner-io/submariner-operator/files/6367207/submariner-20210423170957.zip)

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,62 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
-func podLogsToFile(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
+func getlogs(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
+	podLogOptions.Previous = true
 	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+	logStream, _ := logRequest.Stream()  // TODO: Check for error other than "no previous pods found"
 
-	logs, err := processLogStream(logRequest)
+	// if no previous pods found, logstream == nil, ignore it
+	if logStream != nil {
+		info.Status.QueueWarningMessage(fmt.Sprintf("found previous instances of pod %s", pod.Name))
```
Jaanki: This warning message can be omitted

mangelajo: It's ok to keep I think.

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,62 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
+		err := writeLogs(logStream, info, pod.Name, ".log.old")
+		if err != nil {
+			return err
+		}
+		defer logStream.Close()
+	}
```
mangelajo: let's make this part a function getPreviousLogs ...

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,62 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
+	podLogOptions.Previous = false
+	logRequest = info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+	logStream, err = logRequest.Stream()
+	if err != nil {
+		return errors.WithMessage(err, "error opening log stream")
+	}
+	defer logStream.Close()
+	err = writeLogs(logStream, info, pod.Name, ".log")
+	if err != nil {}
+	return nil
```
mangelajo: And this part another function:  "getExistingLogs()"

In such way we divide the function in smaller more meaningful chunks

mangelajo (changes requested):
Just a nit, looks good

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,46 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
 	}
 }
 
-func podLogsToFile(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
-	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+func getlogs(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
```
tpantelis: This should be camel-case but why was this renamed? `get` implies it retrieves and returns something which this function doesn't. This function outputs the logs so the name should reflect that. We could rename to `outputPodLogs`.

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,46 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
+	logs = scrubSensitiveData(info, logs)
+	err = writeLogToFile(logs, podName, info, fileExtension)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getLogsFromStream(logStream io.ReadCloser) (string, error) {
```
tpantelis:
```suggestion
func getLogFromStream(logStream io.ReadCloser) (string, error) {
```

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,46 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
 	if err != nil {
 		return err
 	}
-
 	return nil
 }
 
-func processLogStream(logrequest *rest.Request) (string, error) {
-	logStream, err := logrequest.Stream()
+func writeLogs(logStream io.ReadCloser, info Info, podName, fileExtension string) error {
```
tpantelis:
```suggestion
func writePodLogToFile(logStream io.ReadCloser, info Info, podName, fileExtension string) error {
```

pkg/subctl/cmd/gather/logs.go
```diff
@@ -120,3 +124,39 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
 
 	return pods, nil
 }
+
+func getPreviousLogs(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
```
tpantelis: Similar to above, this doesn't retrieve/return something. Also this mutates the `podLogOptions` param so pass by value.

```suggestion
func outputPreviousPodLog(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) error {
```

pkg/subctl/cmd/gather/logs.go
```diff
@@ -120,3 +124,39 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
...
+		err := writeLogs(logStream, info, pod.Name, ".log.old")
+		if err != nil {
+			return err
+		}
+		defer logStream.Close()
+	}
+	return nil
+}
+
+func getExistingLogs(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
```
tpantelis:
```suggestion
func outputCurrentPodLog(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) error {
```

pkg/subctl/cmd/gather/logs.go
```diff
@@ -120,3 +124,39 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
...
+	podLogOptions.Previous = true
+	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+	logStream, _ := logRequest.Stream()
+
+	// TODO: Check for error other than "no previous pods found"
+
+	// if no previous pods found, logstream == nil, ignore it
+	if logStream != nil {
+		info.Status.QueueWarningMessage(fmt.Sprintf("found previous instances of pod %s", pod.Name))
+		err := writeLogs(logStream, info, pod.Name, ".log.old")
```
tpantelis: "_.old_" to me kind of implies a copy of another file.  I suggest "_.log.prev_".

pkg/subctl/cmd/gather/logs.go
```diff
@@ -120,3 +124,39 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
 
 	return pods, nil
 }
+
+func getPreviousLogs(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
+	podLogOptions.Previous = true
+	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+	logStream, _ := logRequest.Stream()
+
+	// TODO: Check for error other than "no previous pods found"
```
tpantelis, line 128: yeah we shouldn't ignore the error.

tpantelis (changes requested):

pkg/subctl/cmd/gather/logs.go
```diff
@@ -120,3 +124,39 @@ func findPods(clientSet kubernetes.Interface, byLabelSelector string) (*corev1.P
...
+
+	// if no previous pods found, logstream == nil, ignore it
+	if logStream != nil {
+		info.Status.QueueWarningMessage(fmt.Sprintf("found previous instances of pod %s", pod.Name))
```
tpantelis:
```suggestion
		info.Status.QueueWarningMessage(fmt.Sprintf("Found logs for previous instances of pod %s", pod.Name))
```

pkg/subctl/cmd/gather/logs.go
Jaanki: This patch does that.

I have a operator pod that restarted once.
`
submariner-operator       submariner-operator-7979fd59bb-psg98             1/1     Running     1          35h`

And this is reflected in the output
```
⚠ Gathering operator logs 
 ✓ Found 1 pods matching label selector "name=submariner-operator"
 ⚠ Found logs for previous instances of pod submariner-operator-7979fd59bb-psg98
```
and created
cluster2_submariner-operator-7979fd59bb-psg98.log
cluster2_submariner-operator-7979fd59bb-psg98.log.prev

pkg/subctl/cmd/gather/logs.go
tpantelis: What about the TODO here? It's not a big deal - I guess we can deal with it later.

tpantelis (approved):

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,45 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
 	}
 }
 
-func podLogsToFile(pod *corev1.Pod, podLogOptions *corev1.PodLogOptions, info Info) error {
-	logRequest := info.ClientSet.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, podLogOptions)
+func outputPodLogs(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) error {
+	err := outputPreviousPodLog(pod, podLogOptions, info)
+	if err != nil {
+		return err
```
mangelajo: The problem here is that if this one fails the next one won’t even be attempted. I would save the error and return an aggregated error at the end of the function so we don’t skip the other log dumper.

mangelajo (changes requested):

pkg/subctl/cmd/gather/logs.go
Jaanki: yes. Good catch @mangelajo

mangelajo (approved):

pkg/subctl/cmd/gather/logs.go
```diff
@@ -62,41 +61,55 @@ func gatherPodLogsByContainer(podLabelSelector, container string, info Info) {
...
+	if outputCurrentPodLogErr != nil {
+		outputPodLogsErrors = append(outputPodLogsErrors, fmt.Sprintf("error %s getting logs from current pod %s",
+			outputCurrentPodLogErr.Error(), pod.Name))
+	}
+
+	var outputPodLogsError string
+	for _, err := range outputPodLogsErrors {
+		outputPodLogsError += err + "\n" // Prints each error in a new line
+	}
+	return errors.New(outputPodLogsError)
```
tpantelis: Rather than concatenating errors, I suggest to simplify it and just output the errors directly:

```
func outputPodLogs(pod *corev1.Pod, podLogOptions corev1.PodLogOptions, info Info) {
    err := outputPreviousPodLog(pod, podLogOptions, info)
    if err != nil {
        info.Status.QueueFailureMessage(fmt.Sprintf("Error outputting previous log for pod %q: %v", pod.Name, err))		
    }

    err := outputCurrentPodLog(pod, podLogOptions, info)
    if err != nil {
        info.Status.QueueFailureMessage(fmt.Sprintf("Error outputting current log for pod %q: %v", pod.Name, err))
    }
}
```

tpantelis (changes requested):

tpantelis (approved):

PR #1291, opened 2021-04-22
Remove misleading OVN CNI not supported msg

### LINKED ISSUE #1288: subctl gather always complains with  ⚠ OVN CNI not supported yet, although it's supported
**What happened**:
subctl gather always complains with  ⚠ OVN CNI not supported yet, although it's supported

subctl gather to successfully collect OVN data

Run `subctl gather --module=connectivity` on an environment with CNI=OVN

**Environment**:
- Submariner version (use `subctl version`): 0.9-rc0

vthapar:
OVN CNI is now handled explicitly in its own function instead of generic
`CNIResources`. Remove OVN from CNI type check in `CNIResources`

Fixes #1288

### DIFF
--- a/pkg/subctl/cmd/gather/cni.go
+++ b/pkg/subctl/cmd/gather/cni.go
@@ -98,7 +98,7 @@ func CNIResources(info Info, networkPlugin string) {
 			case typeIPTables:
 				logIPTablesCmds(info, pod)
 			case typeOvn:
-				info.Status.QueueWarningMessage("OVN CNI not supported yet")
+				// no-op. Handled in OVNResources()
 			case typeUnknown:
 				info.Status.QueueFailureMessage("Unsupported CNI Type")
 			}


pkg/subctl/cmd/gather/cni.go
```diff
@@ -97,8 +97,6 @@ func CNIResources(info Info, networkPlugin string) {
 			switch networkPluginCNIType[networkPlugin] {
 			case typeIPTables:
 				logIPTablesCmds(info, pod)
-			case typeOvn:
```
mangelajo, line 100: let's keep the "typeOvn: " with no action, otherwise we will get the typeUnknown "failure".

mangelajo (changes requested):
let's keep the "typeOvn: " with no action, otherwise we will get the typeUnknown "failure".

tpantelis (approved):

mangelajo (approved):

PR #1302, opened 2021-04-26
Beautify subctl benchmark latency output

### LINKED ISSUE #1273: benchmark latency output should be more readable
**What happened**:
subctl benchmark latency output/kubeconfigs/kind-config-cluster3 --intra-cluster
MIGRATED TCP REQUEST/RESPONSE TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 10.3.96.4 (10.3.96) port 0 AF_INET : first burst 0
Minimum Latency Microseconds,Mean Latency Microseconds,Maximum Latency Microseconds,Stddev Latency Microseconds,Transaction Rate Tran/s
37,54.66,6769,58.29,18222.924

Each value should be under its item

Run `subctl benchmark latency output/kubeconfigs/kind-config-cluster3 --intra-cluster`

and same within 2 clusters as well

**Environment**:
- Submariner version (use `subctl version`): 0.9.0

Jaanki:
by writing it to a table

Closes: #1273

### DIFF
--- a/pkg/subctl/benchmark/latency.go
+++ b/pkg/subctl/benchmark/latency.go
@@ -17,6 +17,7 @@ package benchmark
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/onsi/gomega"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
@@ -118,4 +119,12 @@ func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
 
 	framework.By(fmt.Sprintf("Waiting for the client pod %q to exit, returning what client sent", nettestClientPod.Pod.Name))
 	nettestClientPod.AwaitFinishVerbose(Verbose)
+	nettestClientPod.CheckSuccessfulFinish()
+	latencyHeaders := strings.Split(nettestClientPod.TerminationMessage, "\n")[1]
+	headers := strings.Split(latencyHeaders, ",")
+	latencyValues := strings.Split(nettestClientPod.TerminationMessage, "\n")[2]
+	values := strings.Split(latencyValues, ",")
+	for i, v := range headers {
+		fmt.Printf("%s:\t%s\n", v, values[i])
+	}
 }
--- a/pkg/subctl/benchmark/throughput.go
+++ b/pkg/subctl/benchmark/throughput.go
@@ -145,4 +145,6 @@ func runThroughputTest(f *framework.Framework, testParams benchmarkTestParams) {
 
 	framework.By(fmt.Sprintf("Waiting for the client pod %q to exit, returning what client sent", nettestClientPod.Pod.Name))
 	nettestClientPod.AwaitFinishVerbose(Verbose)
+	nettestClientPod.CheckSuccessfulFinish()
+	fmt.Println(nettestClientPod.TerminationMessage)
 }


Jaanki:
Output looks like this

`$ ./../submariner-operator/bin/subctl benchmark latency output/kubeconfigs/kind-config-cluster2 output/kubeconfigs/kind-config-cluster3
MIGRATED TCP REQUEST/RESPONSE TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 10.3.0.7 (10.3.0.) port 0 AF_INET : first burst 0
Minimum Latency Microseconds,Mean Latency Microseconds,Maximum Latency Microseconds,Stddev Latency Microseconds,Transaction Rate Tran/s
42,59.26,10964,57.96,16815.804
Minimum Latency (μs)  Mean Latency (μs)   Maximum Latency (μs)  Stddev Latency (μs)   Transaction Rate (Tran/s)  
42                    59.26               10964                 57.96                 16815.804                  
MIGRATED TCP REQUEST/RESPONSE TEST from 0.0.0.0 (0.0.0.0) port 0 AF_INET to 10.3.32.3 (10.3.32) port 0 AF_INET : first burst 0
Minimum Latency Microseconds,Mean Latency Microseconds,Maximum Latency Microseconds,Stddev Latency Microseconds,Transaction Rate Tran/s
72,104.37,7143,63.73,9557.990
Minimum Latency (μs)  Mean Latency (μs)   Maximum Latency (μs)  Stddev Latency (μs)   Transaction Rate (Tran/s)  
72                    104.37              7143                  63.73                 9557.990                   
`

Jaanki:
Depends on https://github.com/submariner-io/shipyard/pull/529

pkg/subctl/benchmark/latency.go
```diff
@@ -117,5 +128,31 @@ func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
 		nettestClientPod.Pod.Name, clusterAName, nettestClientPod.Pod.Spec.NodeName, remoteIP))
 
 	framework.By(fmt.Sprintf("Waiting for the client pod %q to exit, returning what client sent", nettestClientPod.Pod.Name))
-	nettestClientPod.AwaitFinishVerbose(Verbose)
+	latency := nettestClientPod.AwaitFinishVerbose(Verbose)
```
tpantelis: The data you want can be accessed via `nettestClientPod.TerminationMessage` w/o explicitly returning it.

Jaanki: Didn't know that. Made necessary changes.

pkg/subctl/benchmark/latency.go
```diff
@@ -118,4 +129,30 @@ func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
 
 	framework.By(fmt.Sprintf("Waiting for the client pod %q to exit, returning what client sent", nettestClientPod.Pod.Name))
 	nettestClientPod.AwaitFinishVerbose(Verbose)
+	latencyValues := strings.Split(nettestClientPod.TerminationMessage, "\n")[2]
+	values := strings.Split(latencyValues, ",")
```
tpantelis, line 126: The table output looks good. A minor nit could be that, since it always only has one row, a vertical table would be more suitable.

We are making assumptions about the format and content of the output which is fine although it could break if the tool was changed. We can minimize that by also parsing the comma-separated metric names from the second row (we can replace _Microseconds_ with _μs_ for terseness). We can do some minimal validation by verifying `len(parsedNames) == len(parsedvalues)` and let it panic if there aren't 3 rows.  Also I think we should run `subctl benchmark latency ${KUBECONFIGS_DIR}/kind-config-cluster1 --intra-cluster --verbose` in E2E (after `test_subctl_gather` in _e2e.sh_) as a sanity check and to catch any validation/panics. 

It still dumps the raw pod output which we're replacing and don't need now. I think we should change `AwaitFinishVerbose` in shipyard to only print the output if `verbose` is set, ie

```
    if verbose {
        Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
    }
```

tpantelis (changes requested):

pkg/subctl/benchmark/latency.go
Jaanki: > A minor nit could be that, since it always only has one row, a vertical table would be more suitable.
> We are making assumptions about the format and content of the output which is fine although it could break if the tool was changed. 

Yes. and its simpler to implement and won't need any furhter validations for when the tool changes.. But outputs for all other commands are in this format so implemented this way.

> I think we should change AwaitFinishVerbose in shipyard to only print the output if verbose is set

That would not print any output for `benchmark throughput` if verbose is not passed .

Jaanki: @tpantelis How about something like this https://github.com/Jaanki/shipyard/commit/08ce98a919a95e05fef919126fd509b1552f4418 which empties `np.TerminationMessage` if `trim = true` and call `AwaitFinishVerbose` with `trim = true` in benchmark latency. This will just output the table and not raw output

tpantelis: The point of no verbosity is to eliminate or minimize output.  I'm not clear on what you mean by "empties np.TerminationMessage" and the purpose of your proposed `trim` flag - it amounts to just printing a blank line if `verbose == false` && `trim == true` .  So it's really not printing anything if `verbose == false` which we can simply accomplish with what I proposed above.

tpantelis: Also what about parsing the comma-separated metric names and executing `subctl benchmark` in E2E?

Jaanki: `AwaitFinishVerbose` is called by `benchmark throughput` and `benchmark latency`. Removing the else section in 
```
    if verbose {
        Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
    } else {
        fmt.Printf("%s", removeDupDataplaneLines(np.TerminationMessage))
   }
```
 will not print output of `benchmark throughput` (which is in proper format) when `Verbose=false`.

The raw output in `benchmark latency` is coming from
```
else {
        fmt.Printf("%s", removeDupDataplaneLines(np.TerminationMessage))
   }
```
What I am doing with `trim` is skpping this `else` and manually formatting the output by `latencyPrinter.Print(value)` to elimanate raw output and print just the formatted output.

Jaanki: > The table output looks good. A minor nit could be that, since it always only has one row, a vertical table would be more suitable.

Like this
```

Minimum Latency Microseconds:	50
Mean Latency Microseconds:	79.09
Maximum Latency Microseconds:	10893
Stddev Latency Microseconds:	88.03
Transaction Rate Tran/s:	12599.844

Minimum Latency Microseconds:	91
Mean Latency Microseconds:	142.28
Maximum Latency Microseconds:	9801
Stddev Latency Microseconds:	124.74
Transaction Rate Tran/s:	7006.188

```

tpantelis: Perhaps `benchmark throughput` should pretty print too but that would be another PR.  So we have 3 callers of `AwaitFinishVerbose` and each wants something different:

1. print full verbose output
2. print partial output
3. print no output

Adding yet another param to effect case 3 seems like a hack. We could change `verbose` to a tri-state var, eq `Normal`, `Verbose`, `None` but I think that's overkill.  To me, case 3 should really be `verbose == false` and we should eliminate case 2. So both `latency` and `benchmark` would pass false and `benchmark` would do `fmt.Printf("%s", np.TerminationMessage)` (unless we pretty-print that as well).

Jaanki: `benchmark throughput` is already pretty printed.

```
    if verbose {
        Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
    } else {
        fmt.Printf("%s", removeDupDataplaneLines(np.TerminationMessage))
   }
```

This prints the output from the pod. For others case, it was fine. But for `benchmark latency`, it is not pretty printed. What I am doing is NOT printing the output from the pod directly but formatting it. If `trim=true` and `verbose=false`, do not print the output directly but format it. 
```

This does not change anything for `benchmark throughput` because `trim=false`. But for `benchmark latency`, `trim=true` and hence when `verbose=false`, print just the table and when `verbose=true`, print the raw output + the table

Comment (Jaanki) [reply pkg/subctl/benchmark/latency.go:126]: The scenario is quite simple:
For `benchmark throughtput`

1. when `verbose=true`, print raw output from
```
if verbose {
        Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
```
2. when verbose=false, print

```
else {
        fmt.Printf("%s", removeDupDataplaneLines(np.TerminationMessage))
   }
```
because the output is already nice and hence `trim=false`

For `benchmark latency`,
1. when `verbose=true`, print raw output from
```
if verbose {
        Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
```
and the table

2. when verbose=false, DO NOT print
```
else {
        fmt.Printf("%s", removeDupDataplaneLines(np.TerminationMessage))
   }
```
because the output is not pretty but instead print the table created in this patch and hence `trim=true`

So `trim` flag basically decided whether to print raw output or formatted output.

File: pkg/subctl/benchmark/latency.go [hunk 3]
```diff
@@ -118,4 +129,30 @@ func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
...
+
+	var value []interface{}
+	value = append(value, latencyTypes{
+		minLatency:      values[0],
+		meanLatency:     values[1],
+		maxLatency:      values[2],
+		stddevLatency:   values[3],
+		transactionRate: values[4],
+	})
+	latencyPrinter.Print(value)
```
Jaanki: Doing this instead
```
latencyHeaders := strings.Split(nettestClientPod.TerminationMessage, "\n")[1]
headers := strings.Split(latencyHeaders, ",")
latencyValues := strings.Split(nettestClientPod.TerminationMessage, "\n")[2]
values := strings.Split(latencyValues, ",")
for i, v := range headers {
     fmt.Println(fmt.Sprintf("%s:\t%s", v, values[i]))
}
```

prints a vertical table and does not hard-code any information and does not require adding validations in E2E tests because if the tool changes, the added headers will be automatically added by the `for` loop.

pkg/subctl/benchmark/latency.go
tpantelis: I understand what you're trying to do. But adding a flag named `trim` to mean "don't print anything if not verbose" seems a bit odd to me. I think we can simplify it and accomplish what we want w/o muddying `AwaitFinishVerbose` by adding another param. As I proposed earlier, we modify `AwaitFinishVerbose` to:

```
if verbose {
    Logf("Pod %q output:\n%s", np.Pod.Name, removeDupDataplaneLines(np.TerminationMessage))
}
```

The  original `AwaitFinish` function passes in true for `verbose` so no change in behavior there.

For `benchmark latency`, we pass the `verbose` flag through as we do now. If `verbose == true` we get both the raw and pretty-printed output which I think is fine. If `verbose == false` we get just the pretty-printed output as desired. 

For `benchmark throughtput`, we pass the `verbose` flag through as we do now. If `verbose == true` we get the raw output as before from `AwaitFinishVerbose` but, since the raw output is already pretty-printed, nothing further is needed. 

For `verbose == false`, we can simply print the desired output after `AwaitFinishVerbose`, ie:

```
nettestClientPod.AwaitFinishVerbose(Verbose)
if !Verbose {
    fmt.Print(nettestClientPod.TerminationMessage)
}
```

So this satisfies the 3 use cases cleanly w/o changing the `AwaitFinishVerbose` API.

Jaanki: > For verbose == false, we can simply print the desired output after AwaitFinishVerbose, ie:
> 
 ```
nettestClientPod.AwaitFinishVerbose(Verbose)
 if !Verbose {
     fmt.Print(nettestClientPod.TerminationMessage)
}
```

Shouldn't we check for the pod status before printing?
```
finished := np.Pod.Status.Phase == v1.PodSucceeded || np.Pod.Status.Phase == v1.PodFailed
	if finished {
		np.TerminationCode = np.Pod.Status.ContainerStatuses[0].State.Terminated.ExitCode
                np.TerminationMessage = np.Pod.Status.ContainerStatuses[0].State.Terminated.Message
```

Comment (tpantelis) [reply pkg/subctl/benchmark/latency.go:126]: yeah `CheckSuccessfulFinish`. We weren't ding that before either.

**Comment by Jaanki** [comment]
Depends on https://github.com/submariner-io/shipyard/pull/535

File: pkg/subctl/benchmark/latency.go [hunk 4]
```diff
@@ -118,4 +119,14 @@ func runLatencyTest(f *framework.Framework, testParams benchmarkTestParams) {
 
 	framework.By(fmt.Sprintf("Waiting for the client pod %q to exit, returning what client sent", nettestClientPod.Pod.Name))
 	nettestClientPod.AwaitFinishVerbose(Verbose)
+	nettestClientPod.CheckSuccessfulFinish()
+	latencyHeaders := strings.Split(nettestClientPod.TerminationMessage, "\n")[1]
+	headers := strings.Split(latencyHeaders, ",")
+	latencyValues := strings.Split(nettestClientPod.TerminationMessage, "\n")[2]
+	values := strings.Split(latencyValues, ",")
+	for i, v := range headers {
```
Jaanki, line 127: Prints vertical table

```
Minimum Latency Microseconds:	50
Mean Latency Microseconds:	79.09
Maximum Latency Microseconds:	10893
Stddev Latency Microseconds:	88.03
Transaction Rate Tran/s:	12599.844

Minimum Latency Microseconds:	91
Mean Latency Microseconds:	142.28
Maximum Latency Microseconds:	9801
Stddev Latency Microseconds:	124.74
Transaction Rate Tran/s:	7006.188
```

tpantelis: That must have been `intra-cluster == false`  where it runs 2 latency tests. So that's all it prints sans verbose? I think we need some context prior to each run by always printing

`framework.By("Performing latency tests from Gateway pod to Gateway pod")`

and

`framework.By("Performing latency tests from Non-Gateway pod to Non-Gateway pod")`

tpantelis (dismissed):

pkg/subctl/benchmark/latency.go
Jaanki: These lines are printed only when verbose=true as per current implementation
```
framework.By("Performing latency tests from Gateway pod to Gateway pod")
		runLatencyTest(f, latencyTestParams)
```

go.mod
```diff
@@ -51,3 +51,5 @@ replace (
 	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.5
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a
 )
+
+replace github.com/submariner-io/shipyard v0.9.0-rc0 => github.com/submariner-io/shipyard v0.9.0-rc0.0.20210428164906-1a52a9816269
```
skitt: This needs to be added to the main list of dependencies (otherwise it won’t be handled during the release).

skitt (changes requested):

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/submariner-io/admiral v0.9.0-rc0
 	github.com/submariner-io/cloud-prepare v0.0.0-20210422124637-01644d90a37b
 	github.com/submariner-io/lighthouse v0.9.0-rc0
-	github.com/submariner-io/shipyard v0.9.0-rc0
+	github.com/submariner-io/shipyard v0.9.0-rc0.0.20210428164906-1a52a9816269
```
mangelajo: This is still unmerged, let's reset go.mod & go.sum  :) we can grab those changes once merged.

mangelajo (changes requested):
just the go.mod/sum

mangelajo (approved):

skitt (approved):

PR #1307, opened 2021-04-28
refactor: separate between the component and the image names

SteveMattar:
The component name is not always equivalent to the image name as we experience
in the downstream environment.
Make this separation and allow overriding the image name during the
project's build.
The imagePrefix and imagePostfix are deprecated now and will be removed in
the next release.


mangelajo (approved):
Thank you @SteveMattar it looks very good.

mangelajo:
@SteveMattar I wonder, if we are not going to use the image prefix & suffix anymore, we could as well drop them, now that I see the code again, it doesn't look like a risk.

SteveMattar:
> @SteveMattar I wonder, if we are not going to use the image prefix & suffix anymore, we could as well drop them, now that I see the code again, it doesn't look like a risk.

@mangelajo yes we can, its not risky at all.
Should I do that?

tpantelis (approved):

PR #1313, opened 2021-04-29
Update subctl diagnose to return exit code on failures

### LINKED ISSUE #1310: subctl diagnose to return relevant exit code on Submariner failures
`subctl diagnose` prints "Status.Error/Warnings" and exit normally, even if there's a connection failure.
This is a different response than already implemented by `subctl show` - that returns bad exit code on connection failures.

Add relevant exit code for subctl diagnose.

**Why is this needed**:
common exit codes for subctl commands

maayanf24:
Fixes #1310

### DIFF
--- a/pkg/subctl/cmd/validate_all.go
+++ b/pkg/subctl/cmd/validate_all.go
@@ -69,8 +69,8 @@ func validateAll(cmd *cobra.Command, args []string) {
 		fmt.Println()
 		validationStatus = validationStatus && validateVxLANConfigWithinCluster(item.config, item.clusterName, submariner)
 		fmt.Println()
-		fmt.Printf("Skipping tunnel firewall validation as it requires two kubeconfigs." +
-			" Please run \"subctl validate firewall tunnel\" command manually.\n")
+		fmt.Printf("Skipping tunnel firewall check as it requires two kubeconfigs." +
+			" Please run \"subctl diagnose firewall tunnel\" command manually.\n")
 		fmt.Println()
 	}
 
--- a/pkg/subctl/cmd/validate_all.go
+++ b/pkg/subctl/cmd/validate_all.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
@@ -37,8 +38,10 @@ func validateAll(cmd *cobra.Command, args []string) {
 	configs, err := getMultipleRestConfigs(kubeConfig, kubeContexts)
 	exitOnError("Error getting REST config for cluster", err)
 
+	validationStatus := true
+
 	for _, item := range configs {
-		validateK8sVersionInCluster(item.config, item.clusterName)
+		validationStatus = validationStatus && validateK8sVersionInCluster(item.config, item.clusterName)
 		fmt.Println()
 
 		status.Start(fmt.Sprintf("Retrieving Submariner resource from %q", item.clusterName))
@@ -52,22 +55,26 @@ func validateAll(cmd *cobra.Command, args []string) {
 		status.End(cli.Success)
 		fmt.Println()
 
-		validateCNIInCluster(item.config, item.clusterName, submariner)
+		validationStatus = validationStatus && validateCNIInCluster(item.config, item.clusterName, submariner)
 		fmt.Println()
-		validateConnectionsInCluster(item.config, item.clusterName)
+		validationStatus = validationStatus && validateConnectionsInCluster(item.config, item.clusterName)
 		fmt.Println()
-		checkPods(item, submariner, OperatorNamespace)
+		validationStatus = validationStatus && checkPods(item, submariner, OperatorNamespace)
 		fmt.Println()
-		checkOverlappingCIDRs(item, submariner)
+		validationStatus = validationStatus && checkOverlappingCIDRs(item, submariner)
 		fmt.Println()
-		validateKubeProxyModeInCluster(item.config, item.clusterName)
+		validationStatus = validationStatus && validateKubeProxyModeInCluster(item.config, item.clusterName)
 		fmt.Println()
-		validateFirewallMetricsConfigWithinCluster(item.config, item.clusterName)
+		validationStatus = validationStatus && validateFirewallMetricsConfigWithinCluster(item.config, item.clusterName)
 		fmt.Println()
-		validateVxLANConfigWithinCluster(item.config, item.clusterName, submariner)
+		validationStatus = validationStatus && validateVxLANConfigWithinCluster(item.config, item.clusterName, submariner)
 		fmt.Println()
 		fmt.Printf("Skipping tunnel firewall validation as it requires two kubeconfigs." +
 			" Please run \"subctl validate firewall tunnel\" command manually.\n")
 		fmt.Println()
+	}
+
+	if !validationStatus {
+		os.Exit(1)
 	}
 }
--- a/pkg/subctl/cmd/validate_cni.go
+++ b/pkg/subctl/cmd/validate_cni.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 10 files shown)


pkg/subctl/cmd/validate_cni.go
```diff
@@ -87,7 +87,7 @@ func validateCNIInCluster(config *rest.Config, clusterName string, submariner *v
 			" Supported network plugins: %v\n", submariner.Status.NetworkPlugin, supportedNetworkPlugins)
 		status.QueueFailureMessage(message)
 		status.End(cli.Failure)
-		return
+		exitWithErrorMsg(message)
```
mangelajo: This will exit when running validate all, right?

Shouldn't we do this in L70. ?

And only after we have checked all the clusters, not only the first one.

mangelajo (changes requested):
I'm not going exhaustively over all cases, but:

this changes the behavior, now as soon as there is any failure (even for diagnose all) it will stop and exit, we should save the errors and propagate upwards, while still continuing to run, and then, just before normal exit we can exit with error if any error happened otherwise, we cannot get a full diagnostic as soon as one single error appears.

pkg/subctl/cmd/validate_cni.go
sridhargaddam: I agree with @mangelajo

tpantelis: +1.  We can have each validate function return a success/fail bool that is checked at the top-level.

nyechiel:
(removing from the 0.9 board as this is already being tracked via the linked Issue)

nyechiel (approved):

pkg/subctl/cmd/validate_all.go
```diff
@@ -37,8 +37,12 @@ func validateAll(cmd *cobra.Command, args []string) {
 	configs, err := getMultipleRestConfigs(kubeConfig, kubeContexts)
 	exitOnError("Error getting REST config for cluster", err)
 
+	validationStatus := true
+
 	for _, item := range configs {
-		validateK8sVersionInCluster(item.config, item.clusterName)
+		if !validateK8sVersionInCluster(item.config, item.clusterName) {
+			validationStatus = false
+		}
```
mangelajo:
```suggestion
		validationStatus = validationStatus && validateK8sVersionInCluster(item.config, item.clusterName)
```

pkg/subctl/cmd/validate_all.go
```diff
@@ -52,22 +56,40 @@ func validateAll(cmd *cobra.Command, args []string) {
 		status.End(cli.Success)
 		fmt.Println()
 
-		validateCNIInCluster(item.config, item.clusterName, submariner)
+		if !validateCNIInCluster(item.config, item.clusterName, submariner) {
+			validationStatus = false
+		}
```
mangelajo: validationStatus = validationStatus && ...

pkg/subctl/cmd/validate_connections.go
```diff
@@ -48,11 +50,16 @@ func validateConnections(cmd *cobra.Command, args []string) {
 			continue
 		}
 		status.End(cli.Success)
-		validateConnectionsInCluster(item.config, item.clusterName)
+		if !validateConnectionsInCluster(item.config, item.clusterName) {
+			validationStatus = false
```
mangelajo: validationStatus = validationStatus &&

mangelajo (changes requested):

pkg/subctl/cmd/validate_all.go
```diff
@@ -52,22 +56,40 @@ func validateAll(cmd *cobra.Command, args []string) {
...
 		fmt.Println()
 		fmt.Printf("Skipping tunnel firewall validation as it requires two kubeconfigs." +
 			" Please run \"subctl validate firewall tunnel\" command manually.\n")
 		fmt.Println()
 	}
+
+	if !validationStatus {
+		exitWithErrorMsg("Diagnose all return with an error")
```
tpantelis: Any errors that occurred earlier were already outputted so I don't think we need an extra message here - we can just exit. Same elsewhere.

```suggestion
		os.Exit(1)
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_cni.go
```diff
@@ -207,7 +218,7 @@ func validateCalicoIPPoolsIfCalicoCNI(config *rest.Config) {
 		}
 	}
 
-	status.End(status.ResultFromMessages())
```
mangelajo, line 210: Should we really remove this status.End? or keep it (in addition to the return true?)

tpantelis: We need to keep this and also return false if there were any errors. 

```
result := status.ResultFromMessages()
status.End(result)

return result != cli.Failure
```

pkg/subctl/cmd/validate_deployment.go
```diff
@@ -224,7 +242,7 @@ func checkPodsStatus(k8sClient kubernetes.Interface, operatorNamespace string) b
 			message := fmt.Sprintf("Pod %q is not running. (current state is %v)", pod.Name, pod.Status.Phase)
 			status.QueueFailureMessage(message)
 			status.End(cli.Failure)
-			return false
+			exitWithErrorMsg(message)
```
tpantelis: Why was this changed to exit here?

pkg/subctl/cmd/validate_fw_vxlan.go
```diff
@@ -59,47 +61,56 @@ func validateFirewallVxLANConfig(cmd *cobra.Command, args []string) {
...
-		return
+		return false
 	}
 
 	gateways := getGatewaysResource(config)
 	if gateways == nil {
-		status.QueueWarningMessage("There are no gateways detected on the cluster.")
-		return
+		message := "There are no gateways detected on the cluster."
+		status.QueueWarningMessage(message)
```
tpantelis: nit but why was this changed to store message in a var?

```suggestion
		status.QueueWarningMessage("There are no gateways detected on the cluster.")
```

pkg/subctl/cmd/validate_fw_vxlan.go
```diff
@@ -59,47 +61,56 @@ func validateFirewallVxLANConfig(cmd *cobra.Command, args []string) {
...
+		return false
 	}
 
-	if len(gateways.Items[0].Status.Connections) == 0 {
-		status.QueueWarningMessage("There are no active connections to remote clusters.")
-		return
+	if len(gateways.Items) == 0 || len(gateways.Items[0].Status.Connections) == 0 {
+		message := "There are no active connections to remote clusters."
+		status.QueueWarningMessage(message)
```
tpantelis: same here

tpantelis (changes requested):
I think we should also run `subctl diagnose`  as part of the E2E test as a sanity check.

pkg/subctl/cmd/validate_fw_vxlan.go
```diff
@@ -59,47 +62,52 @@ func validateFirewallVxLANConfig(cmd *cobra.Command, args []string) {
...
 	clientSet, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		message := fmt.Sprintf("Error creating API server client: %s", err)
 		status.QueueFailureMessage(message)
-		return
+		return false
 	}
 
 	gateways := getGatewaysResource(config)
 	if gateways == nil {
```
tpantelis: `if gateways == nil || len(gateways.Items) == 0 { `

pkg/subctl/cmd/validate_fw_vxlan.go
```diff
@@ -59,47 +62,52 @@ func validateFirewallVxLANConfig(cmd *cobra.Command, args []string) {
...
 		status.QueueWarningMessage("There are no gateways detected on the cluster.")
-		return
+		return false
 	}
 
-	if len(gateways.Items[0].Status.Connections) == 0 {
+	if len(gateways.Items) == 0 || len(gateways.Items[0].Status.Connections) == 0 {
```
tpantelis:
```suggestion
	if len(gateways.Items[0].Status.Connections) == 0 {
```

tpantelis (changes requested):

pkg/subctl/cmd/validate_fw_vxlan.go
dfarrell07: Was adding this new `len(gateways.Items) == 0` check intentional for this change? If so, it might be good to document it in the commit message.

dfarrell07: Ah, ditto to the review I was just working on.

maayanf24: it will prevent a panic when there is a gateway object with no items

mangelajo (approved):

pkg/subctl/cmd/validate_fw_vxlan.go
mangelajo: let's keep it simple and add this now.

sridhargaddam (approved):
thanks @maayanf24

tpantelis (approved):

PR #1314, opened 2021-04-29
Run benchmark commands in e2e

tpantelis:
... for sanity checks to catch a panic or abnormal exit. Also allows one to see the output in a CI job w/o having to run it locally.


dfarrell07 (approved):
I wonder if we should eventually extract these subctl benchmark/gather/similar to Shipyard, maybe with make targets, maybe integrate them into E2E/post-mortem there.

mangelajo:
> I wonder if we should eventually extract these subctl benchmark/gather/similar to Shipyard, maybe with make targets, maybe integrate them into E2E/post-mortem there.

May be not for all (benchmark) but for gather & validate, I think we may add them to post mortem at least.

nyechiel:
Note that this PR also includes this fix: https://github.com/submariner-io/submariner-operator/pull/1316

mangelajo (approved):

Issue #1315, opened 2021-04-29
Inter-cluster benchmark throughput fails with globalnet


nyechiel:
**What happened**:

https://github.com/submariner-io/submariner-operator/pull/1314/checks?check_run_id=2467139879

It succeeds for Gateway to Gateway but fails for Non-Gateway to Non-Gateway - the client pod doesn’t complete.

sridhargaddam:
Fix is included as part of following PR - https://github.com/submariner-io/submariner-operator/pull/1314

PR #1316, opened 2021-04-29
Fix inter-cluster benchmark throughput failure on non-gateway nodes

sridhargaddam:
As part of benchmark throughput tests with Globalnet, we first spawn
a nettest server pod (that runs iperf3 server) on Gateway nodes, create
a service in front of it and try to connect to the globalIP of the service
from a clientPod. After the gateway-to-gateway benchmark is completed, when
the test-case tries to benchmark non-gateway to non-gateway benchmark tests,
it was not terminating the previous nettest server pod. Because of this,
we will have two backend nettest server pods, one on the Gateway node
(which belongs to previous test) and the other on the non-Gateway node,
and traffic was getting load-balanced (which is not what we want).

This PR fixes this issue by ensuring that we delete the nettest server pod
before starting non-gateway to non-gateway benchmark test.

Fixes issue: https://github.com/submariner-io/submariner-operator/issues/1315

### DIFF
--- a/pkg/subctl/benchmark/throughput.go
+++ b/pkg/subctl/benchmark/throughput.go
@@ -159,6 +159,7 @@ func runThroughputTest(f *framework.Framework, testParams benchmarkTestParams) {
 	// to remove this dependency with iptables-chain, lets delete the service after the nettest server Pod is terminated.
 	// [*] https://github.com/submariner-io/submariner/issues/1166
 	if framework.TestContext.GlobalnetEnabled && testParams.ClientCluster != testParams.ServerCluster {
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

nyechiel (approved):
Thanks for looking at this so quickly, @sridhargaddam!

nyechiel:
(I am fine with including this in 0.9.0 if the release process hasn't started yet, but I will leave it up to @mangelajo to decide, hence not merging for now)

maayanf24 (approved):

skitt (approved):

sridhargaddam:
The fix is included as part of the following PR - https://github.com/submariner-io/submariner-operator/pull/1314

PR #1324, opened 2021-05-03
fix(subctl): add missing anonymous permissions

SteveMattar:
When running subctl show command using anonymous user we run into
Unauthorized error. Getting some of the resources is forbidden.
Grant the missing permissions.

Closes Bugzilla issue:  https://bugzilla.redhat.com/show_bug.cgi?id=1950960

### DIFF
--- a/config/rbac/anonymous/role.yaml
+++ b/config/rbac/anonymous/role.yaml
@@ -1,0 +1,24 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: submariner-anonymous
+rules:
+  - apiGroups:
+      - submariner.io
+    resources:
+      - submariners
+      - servicediscoveries
+    verbs:
+      - get
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
+    verbs:
+      - get
+  - apiGroups:
+      - networks.config.openshift.io
+    resources:
+      - clusters
+    verbs:
+      - get
--- a/config/rbac/anonymous/role_binding.yaml
+++ b/config/rbac/anonymous/role_binding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: submariner-anonymous
+subjects:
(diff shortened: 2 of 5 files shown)


mangelajo:
I have multiple questions @SteveMattar 

1) what's system:anonymous ? 
2) how to reproduce it
3) And should really system:anonymous have permission to those objects? read/write, only read?
4) Why installation is not an issue (we create some of those objects, like the submariner object).

mangelajo (changes requested):
It looks more like something wrong with the tokens of the reporter than a bug.
According to kubernetes docs: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests

```
Anonymous requests 
When enabled, requests that are not rejected by other configured authentication methods are treated as anonymous requests, and given a username of system:anonymous and a group of system:unauthenticated.
For example, on a server with token authentication configured, and anonymous access enabled, a request providing an invalid bearer token would receive a 401 Unauthorized error. A request providing no bearer token would be treated as an anonymous request.
```

We don't want to provide access to un-authenticated users. We must dig deeper on the reproduction of the issue by @manosnoam I haven't experienced this on my environment with OCP.

SteveMattar:
> I have multiple questions @SteveMattar 
> 
> 1) what's system:anonymous ? 
> 2) how to reproduce it
> 3) And should really system:anonymous have permission to those objects? read/write, only read?
> 4) Why installation is not an issue (we create some of those objects, like the submariner object).

@mangelajo 
1) When you run `oc logout` the user becomes `system:anonymous`
2) Noam was trying to run subctl show commands and he faced the Unauthorized error message when it was trying to access some resources
3) I'm not sure if this user need to have any permission at all but if yes it should be definitely readonly
4) The installation is fine. If I'm logged in, `oc login`, there is no issue.

nyechiel:
@manosnoam - this is not a bug, but an expected behavior of the clients. 

When you run `oc logout` the user becomes system:anonymous, and by design does not have permissions to view/modify resources. Note that this is not specific to `subctl`, but also for `oc` or `kubectl`. You should either use kubeadmin or log in with a proper test user.

PR #1357, opened 2021-05-12
Don't run certain GHA jobs on forks

mkolesnik:
Don't run jobs that aren't triggered on pull request on forks.
Forks aren't likely to need these jobs, and theyre more likely to fail
there.

### DIFF
--- a/.github/workflows/flake_finder.yml
+++ b/.github/workflows/flake_finder.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     name: E2E
+    if: github.repository_owner == 'submariner-io'
     timeout-minutes: 30
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/flake_finder.yml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     name: E2E
+    if: github.repository_owner == 'submariner-io'
```
skitt, line 11: Shouldn’t this also be added to the `upgrade-e2e` job? Or is there a way to disable a workflow globally?

skitt (changes requested):

.github/workflows/flake_finder.yml
mkolesnik: Ah missed it on this file, thanks.

Unfortunately I didn't find find a way to disable on the workflow level, but if you manage to find something it would be nice :)

skitt (approved):

tpantelis (approved):

PR #1361, opened 2021-05-13
Run full E2E matrix only when PR approved

### LINKED ISSUE #1305: Run full E2E matrix only when "ready-for-testing"
E2E test jobs take a long time to complete while the CI has a limited capacity to run concurrent jobs (currently 20 jobs across the entire *@submariner-io* organization.

As such, it would be better to run a default `make e2e` smoke test on each PR sync, and the full test matrix that we have today only when a "ready-for-testing" label is added to the PR.

In order to prevent accidental merging of unlabelled PRs, we can have the PR labelled automatically once it has approval(s).
One possible way to do it is: https://github.com/marketplace/actions/label-approved-pull-requests

**Why is this needed**:
To have faster feedback from small jobs, while still maintaining coverage of E2E.

mkolesnik:
There's no need to run the full E2E test matrix before a PR is approved
for testing, so adding a default test for most of the time and changing
the matrix job to run only when the `ready-to-test` label is present.

The label will be applied once 2 approvals are given on the PR, but can
also be added (or removed) manually as needed.

Resolves: #1305 


.github/workflows/label-ready-approved.yml
```diff
@@ -0,0 +1,13 @@
...
+name: Label Ready to Test
+jobs:
+  label:
+    name: When Approved
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label when approved
+        uses: abinoda/label-when-approved-action@v1.0.7
+        env:
+          APPROVALS: "2"
```
dfarrell07: Looking at the docs, I can't tell if these are CODEOWNER approvals or any general approvals (I suspect second, we'll see after merge). Either way, probably still right to set it at 2.

dfarrell07 (approved):
This is awesome @mkolesnik, thank you! Looking forward to adding more jobs to the Full matrix. :smiling_imp:

.github/workflows/label-ready-approved.yml
mkolesnik: From reading the code [1] it's any approvals, but I think it's fine for this context anyhow since it will still save running all the tests when the PR is fresh and still under review.

[1] https://github.com/abinoda/label-when-approved-action/blob/master/entrypoint.sh#L49

mangelajo (approved):

.github/workflows/label-ready-approved.yml
```diff
@@ -0,0 +1,13 @@
+on: pull_request_review
```
mangelajo, line 1:
```suggestion
on: pull_request_review_target
```

mangelajo: Apparently pull_request_review won't access the secrets, we have a proposed workaround here:

https://stackoverflow.com/questions/67247752/how-to-use-secret-in-pull-request-review-similar-to-pull-request-target

PR #1364, opened 2021-05-14
Refactor/Rename validate commands

tpantelis:
The command name was changed to **diagnose** so the files, variable and function names should reflect that as well. The files were moved under a new `diagnose` package which required exporting some vars/functions from the `cmd` package. A new `ExecuteMultiCluster` function was added to `cmd` to simplify handling multiple kube contexts.


mkolesnik (commented):
Let's make sure to merge PRs requiring backports before this one

tpantelis:
****

> Let's make sure to merge PRs requiring backports before this one

👍 We can hold off on this for a bit.

tpantelis:
bump

mangelajo:
> > Let's make sure to merge PRs requiring backports before this one
> 
>  We can hold off on this for a bit.

Seems like a good time to merge this. Thanks @tpantelis !

mangelajo (approved):

pkg/internal/cli/status.go
```diff
@@ -134,6 +134,21 @@ func (s *Status) End(output Result) {
 	s.warningQueue = []string{}
 }
 
+func (s *Status) EndWithFailure(message string) {
+	s.QueueFailureMessage(message)
```
mkolesnik:
```suggestion
func (s *Status) EndWithFailure(message string, a ...interface{}) {
	s.QueueFailureMessage(fmt.Sprintf(message, a...))
```

I see most calls to this function use `fmt.Sprintf` so it makes it more tidy to just use it internally

pkg/internal/cli/status.go
```diff
@@ -134,6 +134,21 @@ func (s *Status) End(output Result) {
...
+	s.End(Failure)
+}
+
+func (s *Status) EndWithSuccess(message string) {
+	s.QueueSuccessMessage(message)
```
mkolesnik:
```suggestion
func (s *Status) EndWithSuccess(message string, a ...interface{}) {
	s.QueueSuccessMessage(fmt.Sprintf(message, a...))
```

pkg/internal/cli/status.go
```diff
@@ -134,6 +134,21 @@ func (s *Status) End(output Result) {
...
+	s.End(Success)
+}
+
+func (s *Status) EndWithWarning(message string) {
+	s.QueueWarningMessage(message)
```
mkolesnik:
```suggestion
func (s *Status) EndWithWarning(message string, a ...interface{}) {
	s.QueueWarningMessage(fmt.Sprintf(message, a...))
```

mkolesnik (changes requested):
Some initial suggestions to consider

pkg/subctl/cmd/root.go
```diff
@@ -64,23 +64,27 @@ func Execute() error {
 func init() {
 	rootCmd.AddCommand(cmdversion.Cmd)
 	cloudCmd := cloud.NewCommand(&kubeConfig, &kubeContext)
-	addKubeContextFlag(cloudCmd)
+	AddKubeContextFlag(cloudCmd)
 	rootCmd.AddCommand(cloudCmd)
 }
 
-func addKubeConfigFlag(cmd *cobra.Command) {
+func AddToRootCommand(cmd *cobra.Command) {
```
mkolesnik, line 71: I'd argue that it's better if the "higher level" command add it's options and not vice versa.

It makes it easier to maintain and understand the code, and it's more in line with how you actually discover the commands (e.g. when you run `subctl` you get a list of it's sub commands).

WDYT?

pkg/subctl/cmd/diagnose/cni.go
```diff
@@ -15,36 +15,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-var supportedNetworkPlugins = []string{constants.NetworkPluginGeneric, constants.NetworkPluginCanalFlannel, constants.NetworkPluginWeaveNet,
-	constants.NetworkPluginOpenShiftSDN, constants.NetworkPluginOVNKubernetes, constants.NetworkPluginCalico}
-
-var validateCNICmd = &cobra.Command{
-	Use:   "cni",
-	Short: "Check the CNI network plugin",
-	Long:  "This command checks if the detected CNI network plugin is supported by Submariner.",
-	Run:   validateCNIConfig,
-}
+var supportedNetworkPlugins = []string{"generic", "canal-flannel", "weave-net", "OpenShiftSDN", "OVNKubernetes"}
```
mkolesnik: Any particular reason to use hard coded strings and not constants as before?

pkg/subctl/cmd/version/version.go
```diff
@@ -44,3 +54,36 @@ func subctlVersion(cmd *cobra.Command, args []string) {
 func PrintSubctlVersion(w io.Writer) {
 	fmt.Fprintf(w, "subctl version: %s\n", version.Version)
 }
+
+func CheckRequirements(config *rest.Config) (string, []string, error) {
+	failedRequirements := []string{}
```
mkolesnik, line 59: No need for this to be a list.. It's either no values or one..

pkg/subctl/cmd/execute.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return gateways, nil
+}
+
+func ExecuteMultiCluster(run func(*Cluster) bool) {
+	success := true
+
+	for _, config := range mustGetMultipleRestConfigs(kubeConfig, kubeContexts) {
+		cluster := &Cluster{
```
mkolesnik: Why initialize `cluster` here only to override it on L100?

pkg/subctl/cmd/execute.go
```diff
@@ -0,0 +1,115 @@
...
+			Config:     config.config,
+			Name:       config.clusterName,
+			Submariner: nil,
+		}
+
+		fmt.Printf("Cluster %q\n", cluster.Name)
+
+		cluster, errMsg := NewCluster(config.config, config.clusterName)
+		if cluster == nil {
```
mkolesnik, line 95:
```suggestion
		if errMsg != "" {
```

This would be more robust.

Also, why not return `error` type from `NewCluster`?

mkolesnik (changes requested):

mkolesnik:
Generally I find this PR very hard to review...

It's huge and has a lot of code change, and GitHub doesn't recognize most of the renames (not sure if you used `git mv` or it's a GH issue or what).

To compound things further, you did a lot of extra refactoring such as adding `ExecuteMultiCluster` which rewrites a lot of code in a non-trivial matter.

It would be better if you did the renames (and the necessary changes) in one commit, and extra refactoring (such as extracting `ExecuteMultiCluster`, `EndWith*` and others) in a commit on top of it. It could still be part of the same PR, but reviewing such iterative changes is usually easier than reviewing all the changes in one big flat commit.

pkg/subctl/cmd/diagnose/cni.go
tpantelis: The constants were added after I originally push this PR. When I later rebased, it did not show as a conflict.

pkg/subctl/cmd/execute.go
tpantelis: yeah - probably left over from an initial implementation - I did this several weeks ago so I don't recall - this def looks extraneous.

pkg/subctl/cmd/execute.go
tpantelis: Trying to remember but I think it's so the function can return a properly  capitalized string and it can be printed here as is.  This is an artifact of the strict linting which prevents one from returning an error with a capitalized string. So if an error is returned then it has to be uncapitalized which means we'd have to either capitalized it here or wrap it with another capitalized message that wouldn't add any value or print it uncapitalized which is awkward.

pkg/subctl/cmd/version/version.go
tpantelis: yeah the way it's currently implemented but I simply moved this as is so if we're to change the API then I think that should be done in a separate PR. I rather not make this one any larger than it already is 😄

pkg/subctl/cmd/root.go
tpantelis: Unfortunately that would create a cyclic package dependency.

Jaanki (approved):
LGTM Thanks @tpantelis

pkg/subctl/cmd/diagnose/cni.go
mangelajo: Oh I missed this rebase artifact

Jaanki (approved):

pkg/subctl/cmd/diagnose/all.go
```diff
@@ -0,0 +1,75 @@
...
+
+	status := cli.NewStatus()
+	if cluster.Submariner == nil {
+		status.Start(cmd.SubmMissingMessage)
+		status.End(cli.Warning)
+		return success
+	}
+
+	success = success && checkCNIConfig(cluster)
+	fmt.Println()
```
mkolesnik, line 51: Since they're all status prints now, do these empty lines really make sense?

tpantelis: You can look at the output from the [e2e](https://github.com/submariner-io/submariner-operator/pull/1364/checks?check_run_id=2776980288). I guess it's a matter of opinion.  I'm fine either way.

mkolesnik (approved):
What I did manage to compare looks fine, however my review was rather mediocre quality because of the size of the commit

pkg/subctl/cmd/diagnose/all.go
mkolesnik: Yes I ran it locally and it seemed more in line without the line breaks, like in `subctl join` we dont do line breaks

tpantelis: This is unchanged from before. I think it would be better to propose such a change in another PR.

PR #1367, opened 2021-05-17
Use go-semver instead of go-version

skitt:
go-semver’s license is pre-approved by the CNCF, we’d need an
exception for go-version.


dfarrell07 (approved):

Jaanki (approved):

mangelajo (approved):

mangelajo:
I see that go-version is still an indirect requirement, would that be enough?

tpantelis (approved):

PR #1371, opened 2021-05-20
Add network discovery for calico CNI

### LINKED ISSUE #1232: Add network discovery for Calico CNI
Add network discovery to find when cluster is deployed with Calico Cni

**Why is this needed**:
As part of [Epic](https://github.com/submariner-io/submariner-operator/issues/1126) there is a [validation for Calico IPPools](https://github.com/submariner-io/submariner-operator/issues/1149).
The Calico Cni need to be correctly identified in order to apply this validation only when the the cluster is using Calico Cni

Submariner: https://github.com/submariner-io/submariner/pull/1354
Operator: https://github.com/submariner-io/submariner-operator/pull/1371

maayanf24:
The service / cluster CIDR will be discover according
to the generic network discovery, only change is that
in case it's a calico network it will be shown correctly

Fixes issue #1232


pkg/discovery/network/calico_test.go
```diff
@@ -0,0 +1,66 @@
...
+})
+
+func testDiscoverCalicoWith(objects ...runtime.Object) *ClusterNetwork {
+	clientSet := newTestClient(objects...)
+	clusterNet, err := discoverCalicoNetwork(clientSet)
+	Expect(err).NotTo(HaveOccurred())
+	return clusterNet
+}
+
+var calicoCfgMap v1.ConfigMap = v1.ConfigMap{
```
tpantelis: This doesn't need to be global - I suggest to move it inside L27.

```suggestion
var calicoCfgMap v1.ConfigMap = &v1.ConfigMap{
```

pkg/discovery/network/calico_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Expect(clusterNet).NotTo(BeNil())
+			Expect(clusterNet.NetworkPlugin).To(Equal("calico"))
+			Expect(clusterNet.PodCIDRs).To(BeEmpty())
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDRFromService}))
+		})
+	})
+
+	When("There is a kube-api pod", func() {
+		It("Should return the ClusterNetwork structure with the pod CIDR and the service CIDR", func() {
+			clusterNet := testDiscoverWith(
```
tpantelis: This should call discoverCalicoNetwork directly as L30 does. I suggest to use `BeforeEach`/`JustBeforeEach`:

```
var (
    initObjs []runtime.Object
    clusterNet *ClusterNetwork
    err error
)

BeforeEach(func() {
    initObjs = nil
})

JustBeforeEach(func() {
    clientSet := newTestClient(initObjs...)
    clusterNet, err = discoverCalicoNetwork(clientSet) 
})

...
When("...", func() {
    JustBeforeEach(func() {
        initObjs =  []runtime.Object{
             calicoCfgMap,
             fakePod("kube-apiserver", ...),
             fakePod("kube-controller-manager", ...)
        }
    })

    It("...", func() {
        Expect(err).NotTo(HaveOccurred())
        ...
    })
})

```

pkg/discovery/network/calico_test.go
```diff
@@ -0,0 +1,66 @@
...
+	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("discoverCalicoNetwork", func() {
+	When("There are no generic k8s pods to look at", func() {
+		It("Should return the ClusterNetwork structure with the pod CIDR and the service CIDR", func() {
```
tpantelis:
```suggestion
	When("no kube pod information is available",  func() {
		It("should return a ClusterNetwork with only service CIDRs", func() {
```

pkg/discovery/network/calico_test.go
```diff
@@ -0,0 +1,66 @@
...
+			clusterNet := testDiscoverCalicoWith(&calicoCfgMap)
+			Expect(clusterNet).NotTo(BeNil())
+			Expect(clusterNet.NetworkPlugin).To(Equal("calico"))
+			Expect(clusterNet.PodCIDRs).To(BeEmpty())
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDRFromService}))
+		})
+	})
+
+	When("There is a kube-api pod", func() {
+		It("Should return the ClusterNetwork structure with the pod CIDR and the service CIDR", func() {
```
tpantelis:
```suggestion
	When("kube pod information is available", func() {
		It("should return a ClusterNetwork with pod and service CIDRs", func() {
```

pkg/discovery/network/calico_test.go
```diff
@@ -0,0 +1,66 @@
...
+				&calicoCfgMap,
+				fakePod("kube-apiserver", []string{"kube-apiserver", "--service-cluster-ip-range=" + testServiceCIDR}, []v1.EnvVar{}),
+				fakePod("kube-controller-manager", []string{"kube-controller-manager", "--cluster-cidr=" + testPodCIDR}, []v1.EnvVar{}),
+			)
+			Expect(clusterNet).NotTo(BeNil())
+			Expect(clusterNet.NetworkPlugin).To(Equal("calico"))
+			Expect(clusterNet.PodCIDRs).To(Equal([]string{testPodCIDR}))
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDR}))
+		})
+	})
```
tpantelis, line 97: We should also add a test where the  calico ConfigMap doesn't exist.

tpantelis (changes requested):

pkg/discovery/network/calico.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func discoverCalicoNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, error) {
+	_, err := clientSet.CoreV1().ConfigMaps("kube-system").Get(context.TODO(), "calico-config", metav1.GetOptions{})
```
sridhargaddam: We should not assume that calico-config will be in kube-system namespace. Please use the logic that is used in validate_cni.go file

https://github.com/submariner-io/submariner-operator/blob/b52cf2d150b9a167d8a413a0bb5071d8359adda7/pkg/subctl/cmd/validate_cni.go#L114

pkg/discovery/network/calico.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, errors.WithMessage(err, "error obtaining the \"calico\" ConfigMap")
+	}
+
+	clusterNetwork, err := discoverNetwork(clientSet)
+	if err != nil {
+		return nil, err
+	}
+
+	if clusterNetwork != nil {
+		clusterNetwork.NetworkPlugin = "calico"
```
sridhargaddam: The networkPlugin names should be made as Constants and the same should be referred from different repos. 
We need changes even in route-agent to use those constants otherwise it will cause regression 

https://github.com/submariner-io/submariner/blob/ca6d02e1efe507c209fd2d1eba2b301937daea94/pkg/routeagent_driver/handlers/kubeproxy/kp_iptables.go#L82

sridhargaddam (changes requested):
Thanks for the PR @maayanf24

pkg/discovery/network/calico.go
maayanf24: @sridhargaddam - see PR here:
https://github.com/submariner-io/submariner/pull/1354

nyechiel:
(removing from the Project board as this is tracked via the parent Issue)

sridhargaddam (approved):
Thanks for addressing the review comments @maayanf24

tpantelis (approved):

PR #1374, opened 2021-05-21
Fix some supply-chain vulnerabilities

skitt:
We need to ensure that the artifacts we download match what we
expect. For Go modules we can rely on go.sum; for everything else we
need to add external verification.

In the operator, we can move lichen to be tracked by go.sum (although
it should eventually be handled in Shipyard). For the operator SDK, we
can check the GPG signature in addition to keeping a local checksum.

This also moves the operator SDK handling to the Makefile, only for
those few targets which need it; in most cases it is not necessary,
which greatly reduces the Dapper image construction time.


dfarrell07 (approved):

Makefile
```diff
@@ -201,6 +208,25 @@ golangci-lint: generate-embeddedyamls
...
+#   (replacing ${fingerprint} with the full fingerprint);
+# * to update scripts/operator-sdk-signing-keyring.gpg, run
+#     gpg --no-options -q --batch --no-default-keyring \
+#     --output scripts/operator-sdk-signing-keyring.gpg \
+#     --dearmor scripts/operator-sdk-signing-key.asc
+$(OPERATOR_SDK):
+	curl -Lo $@ "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu"
+	curl -Lo $@.asc "https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk-v${OPERATOR_SDK_VERSION}-x86_64-linux-gnu.asc"
+	gpgv --keyring scripts/operator-sdk-signing-keyring.gpg $@.asc $@
+	sha256sum -c scripts/operator-sdk.sha256
```
mangelajo, line 227: Why do we need gpgv + sha256? 

I think gpgv would be enough and we don't need to maintain the hash locally.

Unless I'm missing something of course.

skitt: The two serve different purposes, and the `gpgv` part is really only an example.

The checksums allow us to verify that the binary we get matches the binary we expect: we don’t have to store the binary ourselves, we can download it as needed, but we do want to ensure that it hasn’t been tampered with.
The GPG key allows us to give some level of trust to unknown artifacts coming from the upstream repository; it’s only really useful when we bump the version. It doesn’t protect us from wilful tampering by someone with access to the corresponding private key.

dfarrell07:
Would it make sense to ask Operator-SDK to make their artifacts available in a more secure way (package manager?), so that other consumers don't need logic like this?

mkolesnik (approved):
You're a true `Makefile` wizard, any chance you can do the same for Submariner and `protobuf`? :D

skitt:
> You're a true `Makefile` wizard, any chance you can do the same for Submariner and `protobuf`? :D

Yup, I plan to do this for all the projects ;-).

dfarrell07:
This seems like a bug in our test logic?

> 2021-05-26T16:58:28.6452779Z [36m[submariner-operator]$ [cluster2] cat /tmp/subctl-gather-output/cluster2_submariner-operator-5bfdb87494-6rnt6.log.prev[0m
2021-05-26T16:58:28.6464727Z cat: /tmp/subctl-gather-output/cluster2_submariner-operator-5bfdb87494-6rnt6.log.prev: No such file or directory

I think you added this in https://github.com/submariner-io/submariner-operator/pull/1321 recently, @Jaanki?

https://github.com/submariner-io/submariner-operator/blame/devel/scripts/kind-e2e/lib_subctl_gather_test.sh#L81

I backed up the logs and am re-rerunning the failed job.

https://pastebin.com/vtbnYpQu

Jaanki:
> https://pastebin.com/vtbnYpQu

@dfarrell07 This paste is private.

It's unlisted, but yes I see it's not working when I'm not logged it. I think it's being moderated because it's too large or something.

Here's a gist:

https://gist.github.com/dfarrell07/3601e2229b98c09f648296406bafb31a

tpantelis (approved):

PR #1375, opened 2021-05-23
refactor(bundle): kustomize changes

SteveMattar:
This is relevant only for deploying submariner using the bundle.

When working with `subctl` the manifests definition and objects creation are handled automatically, but when it comes to the operator bundle we have to make sure those resources are aligned.

* Add the missing broker manifests.
* Update the CSV description and samples.

**Note**: The `submariner-k8s-broker` namespace cannot be created during the operator bundle installation.
The user has to make sure the namespace exists before installing the bundle.
Same for the  SCC on Openshift.

Backported in #1398


config/samples/submariner_v1alpha1_servicediscovery.yaml
```diff
@@ -2,15 +2,15 @@
...
-  brokerK8sApiServer: "192.168.67.110:8443"
-  brokerK8sApiServerToken: "$(SECRET_API_SERVER_TOKEN)"
-  brokerK8sRemoteNamespace: "submariner-broker"
-  brokerK8sCA: "$(SECRET_CA)"
-  repository: "quay.io/submariner"
+  brokerK8sApiServer: "$(BROKER_K8S_API_SERVER)"
+  brokerK8sApiServerToken: "$(BROKER_K8S_API_TOKEN)"
+  brokerK8sRemoteNamespace: "$(SUBMARINER_BROKER_NAMESPACE)"
+  brokerK8sCA: "$(BROKER_K8S_CA)"
+  repository: repo
```
mangelajo, line 15: Isn't it better to keep quay.io/submariner here ?

config/samples/submariner_v1alpha1_servicediscovery.yaml
```diff
@@ -2,15 +2,15 @@
...
+  namespace: "$(SUBMARINER_OPERATOR_NAMESPACE)"
+  clusterID: "cluster1"
   debug: false
   globalnetEnabled: true
-  brokerK8sApiServer: "192.168.67.110:8443"
-  brokerK8sApiServerToken: "$(SECRET_API_SERVER_TOKEN)"
-  brokerK8sRemoteNamespace: "submariner-broker"
-  brokerK8sCA: "$(SECRET_CA)"
-  repository: "quay.io/submariner"
+  brokerK8sApiServer: "$(BROKER_K8S_API_SERVER)"
```
mangelajo: Should we may be keep the original example? (as we don't accept https/http in front I think it's a little bit weird)

config/samples/submariner_v1alpha1_submariner.yaml
```diff
@@ -2,26 +2,29 @@
...
 spec:
-  namespace: "submariner-operator"
-  serviceCIDR: "192.168.66.0/24"
-  clusterCIDR: "192.168.67.0/24"
-  globalCIDR: "169.254.0.0/16"
-  clusterID: "east"
-  colorCodes: "blue"
+  serviceCIDR: ""
+  clusterCIDR: ""
+  globalCIDR: "169.254.1.0/24"
```
mangelajo: I would also keep the global CIDR empty by default. (globalnet disabled)

config/samples/submariner_v1alpha1_submariner.yaml
```diff
@@ -2,26 +2,29 @@
...
+  clusterID: "cluster1"
   debug: false
   natEnabled: true
   serviceDiscoveryEnabled: true
   broker: "k8s"
-  brokerK8sApiServer: "192.168.67.110:8443"
-  brokerK8sApiServerToken: "$(SECRET_API_SERVER_TOKEN)"
-  brokerK8sRemoteNamespace: "submariner-broker"
-  brokerK8sCA: "$(SECRET_CA)"
+  brokerK8sApiServer: "$(BROKER_K8S_API_SERVER)"
```
mangelajo: Same about the API server

config/rbac/lighthouse-agent/cluster_role_binding.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: submariner-lighthouse-agent
-    namespace: $(SA_NAMESPACE)
+    namespace: placeholder
```
mangelajo, line 9: why placeholder? what does that mean.

config/broker/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
+---
+namespace: submariner-k8s-broker
+
+resources:
+  - namespace.yaml        # TODO: cannot define a namespace object in bundle/manifests
+  - ../rbac/broker-admin/service_account.yaml
```
mangelajo: Ok, this is something that has to happen inside the broker controller:
https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/broker_controller.go#L81

So the user creates the broker namespace, a broker CR, and then the operator will create: Roles, bindings, SAs, etc...

@skitt what do you think?

May be for now it's ok to have one example/bundle. But the user would need to customize the namespace, I guess.

config/manifests/bases/submariner.clusterserviceversion.yaml
```diff
@@ -48,61 +48,67 @@ spec:
...
-    * **debug**: *'true'* to enable logging debugging information;
+    On the broker cluster:
+
+    1. Create the broker namespace `submariner-k8s-broker`.
+    2. Install the submariner operator from the catalog.
+    3. Create a `Broker` CR.
+
+    On connected clusters:
+
+    1. Create the broker namespace `submariner-k8s-broker`.
```
mangelajo: It shouldn't be necessary to create the broker namespace in connected clusters.

We need to fix our bundles to avoid that. (or keep them separate for now) (see my other comments about the submariner-operator controller).

mangelajo (commented):
See my comments, let's setup a meeting tomorrow with @skitt

config/samples/submariner_v1alpha1_servicediscovery.yaml
SteveMattar: @mangelajo all files under `/config` are templates only. The final result after generating the bundle is located in `/bundle/manifests`.

config/samples/submariner_v1alpha1_submariner.yaml
SteveMattar: Yeah, I was struggling with whether or not to keep this value as an example.

config/rbac/lighthouse-agent/cluster_role_binding.yaml
SteveMattar: again its a template

config/broker/kustomization.yaml
SteveMattar: Same for the scc definition. I will update the description again.

config/manifests/bases/submariner.clusterserviceversion.yaml
SteveMattar: @mangelajo you are right but the reason we have to do it is because the bundle includes the broker manifests as well and when the users installs the operator on any cluster OLM will automatically deploy all manifests including the broker's. There is no way to add a condition when using this approach. As you mentioned in the previous comment, if the operator handles the namespace and the rest of the broker resources then we don't need this step.

config/samples/submariner_v1alpha1_servicediscovery.yaml
SteveMattar: My first thought was let's avoid non-global values but I tend to agree.

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -94,6 +104,18 @@ spec:
     ### Prerequisites
     Submariner has a few requirements to get started, all are documented [here](https://submariner.io/getting-started/quickstart/kind/#prerequisites).
 
+    For an Openshift cluster perform the following steps to bind the submariner service accounts to the **privileged** SCC.
+
+    ```shell
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-operator
```
mangelajo:
```suggestion
```

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -94,6 +104,18 @@ spec:
...
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-gateway
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-routeagent
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-globalnet
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-networkplugin-syncer
```
mangelajo:
```suggestion
```

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -94,6 +104,18 @@ spec:
...
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-lighthouse-agent
```
mangelajo:
```suggestion
```

config/manifests/bases/submariner.clusterserviceversion.yaml
SteveMattar: The user should now manually apply the broker manifests.
The description was updated accordingly.

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -94,6 +104,15 @@ spec:
     ### Prerequisites
     Submariner has a few requirements to get started, all are documented [here](https://submariner.io/getting-started/quickstart/kind/#prerequisites).
 
+    If running on an OpenShift cluster perform the following steps to bind the submariner service accounts to the **privileged** SCC.
```
nyechiel:
```suggestion
    If running on an OpenShift cluster, perform the following steps to bind the Submariner service accounts to the **privileged** SCC.
```

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -103,61 +122,68 @@ spec:
...
-    Deploy the broker:
+    On the broker cluster:
+    1. Create the broker namespace called `submariner-k8s-broker`.
+    2. Apply the broker RBAC definitions:
 
-    `subctl deploy-broker --kubeconfig <PATH-TO-KUBECONFIG-BROKER>`
+        `kubectl apply -k https://github.com/submariner-io/submariner-operator/tree/release-0.9/config/broker`
 
-    For each cluster you want to join, issue the following command:
+    3. Install the submariner operator from the catalog.
```
nyechiel:
```suggestion
    3. Install the Submariner Operator from the catalog.
```

nyechiel: Should we provide a link?

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -103,61 +122,68 @@ spec:
...
+    4. Create a new `Broker` custom resource named **submariner-broker**.
 
-    `subctl join --kubeconfig <PATH-TO-JOINING-CLUSTER> broker-info.subm --clusterid <ID>`
+    On connected clusters:
+    1. Install the submariner operator from the catalog.
```
nyechiel:
```suggestion
    1. Install the Submariner Operator from the catalog.
```

nyechiel: Ditto

packagemanifests/0.9.0/submariner.v0.9.0.clusterserviceversion.yaml
```diff
@@ -94,6 +104,15 @@ spec:
     ### Prerequisites
     Submariner has a few requirements to get started, all are documented [here](https://submariner.io/getting-started/quickstart/kind/#prerequisites).
 
+    If running on an OpenShift cluster perform the following steps to bind the submariner service accounts to the **privileged** SCC.
```
nyechiel:
```suggestion
    If running on an OpenShift cluster, perform the following steps to bind the Submariner service accounts to the **privileged** SCC.
```

packagemanifests/0.9.0/submariner.v0.9.0.clusterserviceversion.yaml
```diff
@@ -103,61 +122,68 @@ spec:
...
-    Deploy the broker:
+    On the broker cluster:
+    1. Create the broker namespace called `submariner-k8s-broker`.
+    2. Apply the broker RBAC definitions:
 
-    `subctl deploy-broker --kubeconfig <PATH-TO-KUBECONFIG-BROKER>`
+        `kubectl apply -k https://github.com/submariner-io/submariner-operator/tree/release-0.9/config/broker`
 
-    For each cluster you want to join, issue the following command:
+    3. Install the submariner operator from the catalog.
```
nyechiel:
```suggestion
    3. Install the Submariner Operator from the catalog.
```

packagemanifests/0.9.0/submariner.v0.9.0.clusterserviceversion.yaml
```diff
@@ -103,61 +122,68 @@ spec:
...
+    4. Create a new `Broker` custom resource named **submariner-broker**.
 
-    `subctl join --kubeconfig <PATH-TO-JOINING-CLUSTER> broker-info.subm --clusterid <ID>`
+    On connected clusters:
+    1. Install the submariner operator from the catalog.
```
nyechiel:
```suggestion
    1. Install the Submariner Operator from the catalog.
```

bundle/manifests/submariner.clusterserviceversion.yaml
SteveMattar: The description appears on the catalog web page. The user is already on the right place and just needs to hit the install button.

nyechiel:
@skitt @mangelajo can you please review?

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -94,6 +103,15 @@ spec:
...
 
+    If running on an OpenShift cluster, perform the following steps to bind the Submariner service accounts to the **privileged** SCC.
+
+    ```shell
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-gateway
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-routeagent
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-globalnet
+    oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-lighthouse-coredns
+    ```
+
```
mangelajo, line 114: Can we move these under the suggested "#### Deployment with the operator (without subctl)" ?  as those wouldn't apply to subctl?

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -103,61 +121,71 @@ spec:
...
 
-    #### Deploying with subctl
-    Install the tool:
-    ```bash
-    curl -Ls https://get.submariner.io | bash
-    export PATH=$PATH:~/.local/bin
-    echo export PATH=\$PATH:~/.local/bin >> ~/.profile
-    ```
-    Deploy the broker:
+    On the broker cluster:
```
mangelajo, line 124:
```suggestion
#### Deployment with subctl
For deploying the submariner operator and submariner we recommend the use of subctl. Please have a look at our [quickstarts](https://submariner.io/getting-started/quickstart/) for more details.

#### Deployment with the operator (without subctl)
    On the broker cluster:
```

mangelajo (changes requested):
I think I would like to keep a pointer still to subctl, before anybody ventures into standalone operator management they should know that subctl is available and that we prefer that solution instead.

bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -103,61 +121,71 @@ spec:
     For complete information about subctl, please refer to [this page](https://submariner.io/operations/deployment/subctl).
     In addition to Operator and subctl, Submariner also provides [Helm Charts](https://submariner.io/operations/deployment/helm).
```
SteveMattar, line 123:
```suggestion
### Deployment
    Submariner provides an [Operator](https://github.com/submariner-io/submariner-operator) for easy API-based
    installation and management.
    A command line utility, [subctl](https://github.com/submariner-io/submariner-operator/releases), wraps the
    Operator to aid users with manual deployments and easy experimentation.
    subctl greatly simplifies the deployment of Submariner, and is therefore the recommended deployment method.
    For complete information about subctl, please refer to [this page](https://submariner.io/operations/deployment/subctl).
    In addition to Operator and subctl, Submariner also provides [Helm Charts](https://submariner.io/operations/deployment/helm).
```
@mangelajo `subctl` is mentioned here. There is a reference with a link for all sort of deployments we provide: subctl, helm,...
I think this specific description should focus on the bundle installation only and if the user wants to use a different method he should follow our documentation.
Does it make sense?

mangelajo: I want to highlight that they should use subctl as first option, unless they have a good reason to use the operator standalone. As it will save many general issues.

Do you think there is a better way to highlight it?

skitt (approved):

bundle/manifests/submariner.clusterserviceversion.yaml
SteveMattar: In our current description we recommend for subctl:

> subctl greatly simplifies the deployment of Submariner, and is therefore the recommended deployment method

We can maybe rephrase the sentence to make the message even more clear.

dfarrell07 (approved):

bundle/manifests/submariner.clusterserviceversion.yaml
SteveMattar: @mangelajo

mangelajo (approved):

PR #1383, opened 2021-05-26
Don't run E2E on push

mkolesnik:
We're already running E2E (in greater capacity) on each PR, and have
branch protections to make sure only up to date, tested, and approved
code can be merged, so having E2E run on push is just redundant (and
detrimental since it's prone to "hidden" failures which prevent latest
images from publishing).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,30 +8,8 @@ on:
       - release-*
 
 jobs:
-  e2e:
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):
Agreed, it only causes problems IMHO. :+1:

tpantelis (approved):

PR #1400, opened 2021-06-02
Use golang 1.16 embed instead of our own

skitt:
Embedded files are always relative to the package embedding them, and
can't be symlinks, so we need to move config and deploy.


pkg/yamls/yamls.go
```diff
@@ -0,0 +1,159 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package embeddedyamls
+
+import _ "embed"
+
+//go:embed deploy/crds/submariner.io_brokers.yaml
```
Jaanki: Embed all files into a file system.

Eg:

import "embed"
// content is our static web server content.
//go:embed image template html/index.html
var content embed.FS

Jaanki (changes requested):

mangelajo:
Cool

pkg/yamls/yamls.go
Jaanki: The calling functions expect a string. So embed them as a string as you are currently doing.

Jaanki (changes requested):
`submariner-operator/pkg/subctl/operator/common/embeddedyamls/generators/yamls2go.go` needs to be deleted

Makefile
```diff
@@ -129,27 +135,33 @@ bin/subctl-%: generate-embeddedyamls $(shell find pkg/subctl/ -name "*.go") vend
...
+# There's a chicken-and-egg situation here: we need the embedded resources
+# to build anything, including the builds required by the controller; so
+# we touch the generated files before calling the generator, and delete
+# empty files afterwards
+
+pkg/yamls/deploy/crds/submariner.io_servicediscoveries.yaml: $(CONTROLLER_GEN) ./apis/submariner/v1alpha1/servicediscovery_types.go vendor/modules.txt
+	for file in $(GENERATED_YAMLS); do mkdir -p $${file%/*}; done
+	touch $(GENERATED_YAMLS)
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) paths=./... output:crd:artifacts:config=pkg/yamls/deploy/crds
+	find $(GENERATED_YAMLS) -empty -delete
```
skitt: This is ugly and duplicated below, I’ll fix it.

pkg/yamls/yamls.go
skitt: Right, and this would also avoid the chicken-and-egg problem...

Jaanki:
bump

Jaanki:
reopen

PR #1416, opened 2021-06-11
Configure monitoring as a core service on OCP

### LINKED ISSUE #1299: Monitoring on OCP should be configured as a core service instead of user workload
From OCM:

> Some steps are required
>
> 1. create one ServiceMonitor resource(https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#servicemonitor) in openshift-monitoring namespace. It should includes the necessary info of your metrics endpoint so that OCP prometheus knows how to scrape metrics from it.
> 1. create corresponding role/rolebinding so that the serviceaccount of OCP prometheus pod has the permission to get the info of pod/endpoint/svc in the namespace where your metrics endpoint running
>
> Below is our existing work which expose an endpoint to OCP prometheus:
>
> https://github.com/open-cluster-management/kui-web-terminal-chart/blob/main/stable/kui-web-terminal/templates/kui-web-terminal-metrics.yaml is the yaml for ServiceMonitor
>
> https://github.com/open-cluster-management/kui-web-terminal-chart/blob/main/stable/kui-web-terminal/templates/kui-metrics-reader.yaml and https://github.com/open-cluster-management/kui-web-terminal-chart/blob/main/stable/kui-web-terminal/templates/read-kui-metrics.yaml are the yaml for role/rolebinding

skitt:
This sets up the ServiceMonitors in the OpenShift monitoring namespace
instead of the monitored services’ namespaces, which enables
monitoring as a core service without user configuration.

Metric-reading roles are set up in all cases for now.

We pull in pkg/metrics from the Operator SDK because the SDK’s version
doesn’t support setting up cross-namespace ServiceMonitors; fixing it
upstream isn’t an option because it has been dropped there in favour
of kustomize handling.

Fixes: #1299

### DIFF
--- a/config/rbac/submariner-metrics-reader/role_binding.yaml
+++ b/config/rbac/submariner-metrics-reader/role_binding.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: read-submariner-metrics
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
+roleRef:
+  kind: Role
+  name: submariner-metrics-reader
+  apiGroup: rbac.authorization.k8s.io
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Copyright Contributors to the Submariner project.
+// Copyright 2018 The Operator-SDK Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/util"
+
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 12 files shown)


dfarrell07 (approved):
We need to parallel these RBAC changes in Helm, right?

skitt:
> We need to parallel these RBAC changes in Helm, right?

That’s a good point, yes, we do!

skitt:
Here is a screenshot of metrics being gathered from the `openshift-monitoring` namespace, measuring in the `submariner-operator` namespace, automatically configured by the `subctl` built from this PR with the operator image built from this PR (pushed with the `:ocpmon` tag on Quay):

Screenshot from 2021-06-14 16-47-14

skitt:
@SteveMattar is the bundle configuration correct here?

config/rbac/submariner-metrics-reader/role_binding.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: read-submariner-metrics
+  namespace: submariner-operator
```
SteveMattar:
```suggestion
```
The namespace here will be generated by kustomize

SteveMattar (changes requested):
@skitt it looks good I added one comment.

nyechiel:
@SteveMattar can you please review again?

SteveMattar (approved):

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,181 @@
...
+			s.Spec.ClusterIP = existingService.Spec.ClusterIP
+		}
+		err = client.Update(ctx, s)
+		if err != nil {
+			return nil, err
+		}
+		log.Info("Metrics Service object updated", "Service.Name",
+			s.Name, "Service.Namespace", s.Namespace)
+		return s, nil
+	}
```
tpantelis, line 81: It seems we could use the admiral `CreateOrUpdate` API here...

tpantelis (changes requested):

pkg/metrics/metrics.go
skitt: As far as I can tell, the Admiral API doesn’t return the object as really written, does it?

tpantelis: No - we could change that or you could do a `Get` here on success.

tpantelis (approved):

PR #1422, opened 2021-06-15
Add gateway container ports to match up the LoadBalancer

mangelajo:
This seems to make the aws implementation of the load balancer
more stable, it doesn't completely fix the issue, but seems to alleviate
the number of error states.

Related-Issue: https://github.com/submariner-io/submariner/issues/1410


nyechiel (approved):

nyechiel:
@mangelajo does this change solve the issue described in https://github.com/submariner-io/submariner/issues/1410? It sounds that not completely, so can you explain what/how this improve things?

mangelajo:
> @mangelajo does this change solve the issue described in [submariner-io/submariner#1410](https://github.com/submariner-io/submariner/issues/1410)? It sounds that not completely, so can you explain what/how this improve things?

That's correct, I will explain better in the PR/Commit message. What I've seen is less error states in the LoadBalancer object when the ports are named the same. It could be accidental but, it will harm nobody to name the ports ':-)

It doesn't fix it.

I'm adding a workaround to avoid lengthy waits for the natt discovery protocol with special handling for the load balancer cases.

tpantelis (approved):

PR #1444, opened 2021-06-23
Add a cross-build check

skitt:
We regularly break Windows builds without realising, e.g. by pulling
in go-iptables in subctl. This adds a job to cross-build  subctl; it
won't detect many failures at the point of origin (in another
project), but it will flag them when the changes affect subctl.


mkolesnik (commented):
This sounds reasonable, but why not check all cross compile?

skitt:
> This sounds reasonable, but why not check all cross compile?

We could, I wanted to avoid comments about adding too much work in CI ;-).

.github/workflows/cross.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Cross-Build Checks
+
+on:
+  pull_request:
+  push:
+    tags:
+      - 'v**'
```
mkolesnik: I think I missed this one, whats the purpose of testing this on tags?

mkolesnik:
> We could, I wanted to avoid comments about adding too much work in CI ;-).

Well given that it's 1 job I think it's reasonable.

If we're worried about clogging up we can change it to have it run only when `ready-to-test` label is applied.

.github/workflows/cross.yml
skitt: Copy-pasta...

dfarrell07 (approved):

tpantelis (approved):

PR #1452, opened 2021-06-28
Change globalCIDR to 242.0.0.0/8

### LINKED ISSUE #1458: Use 242.0.0./8 for default GlobalnetCIDRRange
Replace default GlobalnetCIDRRange of `169.254.0.0/16` with `242.0.0.0/8` and add validations to disallow CIDRs as per validations for `EndpointSlices`

**Why is this needed**:
Globalnetv2+Headless use case adds Globalnet IPs to `EndpointSlices` instead of the PodIPs. EndpointSlices don't allow any link-local, loopback or multicast IPs. If GlobalCIDR configured for a cluster falls within one of these invalid ranges, EndpointSlices can't be created and Headless Services can't be supported by ServiceDiscovery.

vthapar:
Current default globalCIDR of 169.254.x.0 cannot be used for
headless services as EndpointSlices can't have IP in link-local
CIDR range. Change to 242.254.x.0

* Change default Global CIDR to `242.254.0.0/16`
* Add validations for globalCIDRs configured by user

Fixes #1458 

### DIFF
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -27,7 +27,9 @@ import (
 	"net"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	"github.com/submariner-io/submariner-operator/pkg/broker"
 	"github.com/submariner-io/submariner-operator/pkg/internal/cli"
@@ -269,7 +271,7 @@ func ValidateGlobalnetConfiguration(globalnetInfo *GlobalnetInfo, netconfig Conf
 	}
 
 	if globalnetCIDR != "" {
-		_, _, err := net.ParseCIDR(globalnetCIDR)
+		err := IsValidCIDR(globalnetCIDR)
 		if err != nil {
 			return "", fmt.Errorf("specified globalnet-cidr is invalid: %s", err)
 		}
@@ -367,4 +369,48 @@ func AssignGlobalnetIPs(globalnetInfo *GlobalnetInfo, netconfig Config) (string,
 	}
 	status.End(cli.Success)
 	return globalnetCIDR, nil
+}
+
+func IsValidCIDR(cidr string) error {
+	ip, _, err := net.ParseCIDR(cidr)
+
+	if err != nil {
+		return err
+	}
+	if ip.IsUnspecified() {
+		return fmt.Errorf("%s can't be unspecified", cidr)
+	}
+	if ip.IsLoopback() {
+		return fmt.Errorf("%s can't be in loopback range", cidr)
+	}
+	if ip.IsLinkLocalUnicast() {
+		return fmt.Errorf("%s can't be in link-local range", cidr)
+	}
+	if ip.IsLinkLocalMulticast() {
+		return fmt.Errorf("%s can't be in link-local multicast range", cidr)
+	}
+
+	return nil
+}
+
+func ValidateExistingGlobalNetworks(config *rest.Config, namespace string) error {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("error creating the kubernetes clientset: %s", err)
+	}
+
+	globalnetInfo, _, err := GetGlobalNetworks(clientset, namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("error getting existing globalnet configmap: %s", err)
+	}
+
+	if globalnetInfo != nil {
+		err = IsValidCIDR(globalnetInfo.GlobalnetCidrRange)
+		return fmt.Errorf("invalid GlobalnetCidrRange: %s", err)
+	}
+
+	return nil
 }
--- a/pkg/discovery/globalnet/globalnet_test.go
+++ b/pkg/discovery/globalnet/globalnet_test.go
@@ -215,3 +215,37 @@ var _ = Describe("AllocateGlobalCIDR: Fail", func() {
 		})
 	})
 })
+
+var _ = Describe("IsValidCidr", func() {
+	When("Unspecified CIDR", func() {
+		err := IsValidCIDR("")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+		err = IsValidCIDR("1.2")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("Loopback CIDR", func() {
+		err := IsValidCIDR("127.0.0.0/16")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("Link-Local CIDR", func() {
+		err := IsValidCIDR("169.254.0.0/16")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("Link-Local Multicast CIDR", func() {
+		err := IsValidCIDR("224.0.0.0/24")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+})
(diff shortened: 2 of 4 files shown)


tpantelis (approved):

pkg/discovery/globalnet/globalnet.go
```diff
@@ -368,3 +368,25 @@ func AssignGlobalnetIPs(globalnetInfo *GlobalnetInfo, netconfig Config) (string,
...
+		return err
+	}
+	if ip.IsUnspecified() {
+		return fmt.Errorf("%s can't be unspecified", cidr)
+	}
+	if ip.IsLoopback() {
+		return fmt.Errorf("%s can't be in loopback range", cidr)
+	}
+	if ip.IsLinkLocalUnicast() {
+		return fmt.Errorf("%s can't be in linklocal range", cidr)
```
dfarrell07:
```suggestion
		return fmt.Errorf("%s can't be in link-local range", cidr)
```

As you use below

dfarrell07:
```suggestion
		return fmt.Errorf("%s can't be in link-local range", cidr)
```

pkg/discovery/globalnet/globalnet_test.go
```diff
@@ -215,3 +215,37 @@ var _ = Describe("AllocateGlobalCIDR: Fail", func() {
...
+	})
+
+	When("Loopback CIDR", func() {
+		err := IsValidCIDR("127.0.0.0/16")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("Link Local CIDR", func() {
```
dfarrell07:
```suggestion
	When("Link-Local CIDR", func() {
```

pkg/discovery/globalnet/globalnet_test.go
```diff
@@ -215,3 +215,37 @@ var _ = Describe("AllocateGlobalCIDR: Fail", func() {
...
+		err := IsValidCIDR("169.254.0.0/16")
+		It("Should return error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("Link Local Multicast CIDR", func() {
```
dfarrell07:
```suggestion
	When("Link-Local Multicast CIDR", func() {
```

dfarrell07 (commented):
Nit, but there are three version of "link-local" (+"linklocal", "link local") here and it would be nice to be consistent.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -56,7 +56,7 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
 		"enable support for Overlapping CIDRs in connecting clusters (default disabled)")
-	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "169.254.0.0/16",
+	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "242.254.0.0/16",
```
sridhargaddam:
```suggestion
	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "242.0.0.0/8",
```

We used "/16" with LLA as it was the only range and we could not go beyond that. But while moving to 242.x.y.z, we can use the complete "/8", i.e., 242.0.0.0/8 which allows 16777214 addresses :)

pkg/subctl/cmd/deploybroker.go
```diff
@@ -56,7 +56,7 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
...
 		"GlobalCIDR supernet range for allocating GlobalCIDRs to each cluster")
 	deployBroker.PersistentFlags().UintVar(&defaultGlobalnetClusterSize, "globalnet-cluster-size", 8192,
```
sridhargaddam: With the increase in the default Globalnet CIDR, we can even increase the default globalnet-cluster-size, but with new GN2.0 design we will not be consuming too many IPs anyway, so lets leave the default to 8192 itself.

pkg/subctl/cmd/deploybroker.go
```diff
@@ -203,7 +203,9 @@ func isValidGlobalnetConfig() (bool, error) {
 	if err != nil || defaultGlobalnetClusterSize == 0 {
```
sridhargaddam, line 205: @vthapar In the API CreateGlobalnetConfigMap() of this file, we will check if there is already an existing GlobalnetConfigMap on the Broker and avoid creating it if present. Shall we check the default globalCIDR that is present in the config Map and update it to the new CIDR (242.0.0.0/8). Although we are not handling upgrades, this is an easy step and will go a long way to help our users who forget to delete the globalnetConfigMap on the Broker. What do you say?

CC: @mangelajo @vthapar @tpantelis @nyechiel

sridhargaddam (changes requested):

pkg/subctl/cmd/deploybroker.go
vthapar: @sridhargaddam Then should we also change the CIDR size available to each cluster to a bigger number? 255 may not be enough, right?

pkg/subctl/cmd/deploybroker.go
vthapar: If we don't want to support upgrades, we shouldn't support this. Detecting old CIDR and changing it to new falls under supporting upgrades. It will cause confusion if some aspects of an upgrade are handled and others not. As long as we throw error if old one is using an unsupported type, it should be fine.

pkg/subctl/cmd/deploybroker.go
vthapar: I think we will get more value from increasing no. of IPs in each cluster than no. of clusters that we can support, but agree that 8192 is good enough for now, we're even less likely to hit this limit with v2.

sridhargaddam: Right, 8192 addresses should be sufficient enough for any cluster in the new implementation.

pkg/subctl/cmd/deploybroker.go
sridhargaddam: In that case, we should return error when deploy-broker is executed and Broker already has a globalnetConfigMap with old CIDR. Can you please include the change as part of CreateGlobalnetConfigMap handling.

sridhargaddam (changes requested):

pkg/subctl/cmd/deploybroker.go
vthapar: Couldn't add it to CreateGlobalnetConfigMap() coz it is in broker and would've resulted in a cyclic dependency between globalnet and broker. Maybe move CreateGlobalnetConfigMap to globalnet instead?

sridhargaddam (approved):

pkg/subctl/cmd/deploybroker.go
```diff
@@ -56,9 +55,9 @@ var validComponents = []string{components.ServiceDiscovery, components.Connectiv
 func init() {
 	deployBroker.PersistentFlags().BoolVar(&globalnetEnable, "globalnet", false,
 		"enable support for Overlapping CIDRs in connecting clusters (default disabled)")
-	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "169.254.0.0/16",
+	deployBroker.PersistentFlags().StringVar(&globalnetCIDRRange, "globalnet-cidr-range", "242.0.0.0/8",
 		"GlobalCIDR supernet range for allocating GlobalCIDRs to each cluster")
-	deployBroker.PersistentFlags().UintVar(&defaultGlobalnetClusterSize, "globalnet-cluster-size", 8192,
+	deployBroker.PersistentFlags().UintVar(&defaultGlobalnetClusterSize, "globalnet-cluster-size", 65536,
```
mangelajo, line 60: :rocket:

mangelajo: This would support 256 clusters, while 16384 would support 1024 clusters.

Shall we may be try such middle land? %)

PR #1456, opened 2021-06-29
Some fixes for `subctl-devel` release job

mkolesnik:
* Release should be also deleted or it leaves a "draft release".
* Tag name should be `subctl-${BRANCH}` e.g. `subctl-devel`.
* Added notes to the release created in GH to match what we had.
* Seems `gh` CLI is already available, no need to install it.

### DIFF
--- a/.github/workflows/release_subctl_devel.yml
+++ b/.github/workflows/release_subctl_devel.yml
@@ -19,24 +19,19 @@ jobs:
 
       - name: Generate subctl version
         run: |
-          echo "SUBCTL_TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
-          echo "RELEASE=release-${GITHUB_REF##*/}" >> $GITHUB_ENV
+          echo "BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+          echo "RELEASE=subctl-${GITHUB_REF##*/}" >> $GITHUB_ENV
 
       - name: Generate the subctl release artifacts
-        run: make build-cross VERSION=${{ env.SUBCTL_TAG }}
+        run: make build-cross VERSION=${{ env.BRANCH }}
 
-      - name: Install GH CLI
-        run: |
-          curl -sLo /tmp/gh.deb https://github.com/cli/cli/releases/download/v1.11.0/gh_1.11.0_linux_amd64.deb
-          sudo apt install /tmp/gh.deb
-
-      - name: Delete old release
-        run: git push -d origin ${{ env.RELEASE }} || true
-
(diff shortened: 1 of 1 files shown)


skitt (approved):

.github/workflows/release_subctl_devel.yml
```diff
@@ -19,24 +19,19 @@ jobs:
...
+          echo "RELEASE=subctl-${GITHUB_REF##*/}" >> $GITHUB_ENV
 
       - name: Generate the subctl release artifacts
-        run: make build-cross VERSION=${{ env.SUBCTL_TAG }}
+        run: make build-cross VERSION=${{ env.BRANCH }}
 
-      - name: Install GH CLI
-        run: |
-          curl -sLo /tmp/gh.deb https://github.com/cli/cli/releases/download/v1.11.0/gh_1.11.0_linux_amd64.deb
-          sudo apt install /tmp/gh.deb
```
mangelajo, line 31: why removing it?, is it installed by default?

.github/workflows/release_subctl_devel.yml
```diff
@@ -19,24 +19,19 @@ jobs:
...
-      - name: Create the release with the updated artifacts
+      - name: Recreate the release with the updated artifacts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          gh release delete ${{ env.RELEASE }} -y || true
+          git push -d origin ${{ env.RELEASE }} || true
           gh release create ${{ env.RELEASE }} dist/*.tar.xz --prerelease \
-            --title "Cutting Edge: ${{ env.SUBCTL_TAG }}" \
+            --title "Cutting Edge: ${{ env.BRANCH }}" \
```
mangelajo, line 35: but binaries need to go to subctl-devel or subctl-release-0.9 otherwise we need to change the scripts. Won't this mess with the release-xxx or devel branch?

mangelajo (commented):
questions

.github/workflows/release_subctl_devel.yml
mkolesnik: as I wrote in commit message

> Seems gh CLI is already available, no need to install it.

.github/workflows/release_subctl_devel.yml
mkolesnik: this is just the name of the release, not sure what youre thinking about

mangelajo: Oh right, this is the title.. RELEASE=subctl-xxxx

mangelajo (approved):
thanks for the answers :)

PR #1457, opened 2021-06-29
Truncate token prefixes for long cluster ids

### LINKED ISSUE #1426: subctl fails with an SA error if the cluster is too long
```
$ subctl join   ./broker-info.subm  --clusterid default-api-test12-cluster-a-devcluster-openshift-com-6443-master 

* ./broker-info.subm says broker is at: https://api.test12-cluster-a.devcluster.openshift.com:6443
* There are 1 labeled nodes in the cluster:
  - ip-10-12-33-27.us-east-2.compute.internal
    Discovered network details:
        Network plugin:  OVNKubernetes
        Service CIDRs:   [172.42.0.0/16]
        Cluster CIDRs:   [10.212.0.0/14]
 ✓ Discovering network details
 ✓ Validating Globalnet configurations
 ✓ Discovering multi cluster details
 ✓ Deploying the Submariner operator 
 ✓ Created Lighthouse service accounts and roles
 ✗ Creating SA for cluster 
Error creating SA for cluster: error getting cluster sa token: ServiceAccount cluster-default-api-test12-cluster-a-devcluster-openshift-com-6443-master does not have a secret of type token

subctl version: v0.9
```

I’m guessing this is because the cluster id is too long. Regardless, we need to figure out why this is failing, and if it is caused by the length of the cluster id, enforce that earlier, in the cluster id validity check.

skitt:
When an SA is created, Token Controller creates a token for it; that
token's name is auto-generated, with a random suffix, and is at most
63 characters in length, of which 5 are random. Since we look for the
generated token by name, we need to take this into account, and
truncate the prefix using the same rules as the token generator.

Fixes: #1426

### DIFF
--- a/pkg/broker/rbac.go
+++ b/pkg/broker/rbac.go
@@ -132,6 +132,10 @@ func NewBrokerRoleBinding(serviceAccount, role string) *rbacv1.RoleBinding {
 	return binding
 }
 
+// MaxGeneratedNameLength is the maximum generated length for a token, excluding the random suffix
+// See k8s.io/apiserver/pkg/storage/names
+const MaxGeneratedNameLength = 63 - 5
+
 func GetClientTokenSecret(clientSet clientset.Interface, brokerNamespace, submarinerBrokerSA string) (*v1.Secret, error) {
 	sa, err := clientSet.CoreV1().ServiceAccounts(brokerNamespace).Get(context.TODO(), submarinerBrokerSA, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

Jaanki (approved):

skitt:
Backported to 0.9 as https://github.com/submariner-io/submariner-operator/pull/1473.

skitt:
Backported to 0.8 as #1474.

PR #1459, opened 2021-06-30
Return NotFound error as-is in GetGlobalNetworks

vthapar:
In GetGlobalNetworks we wrap error returned with a custom error.
This breaks the validation check in calling methods if they want
to ignore IsNotFound error. Fix is to return IsNotFound error as is.

### DIFF
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -293,7 +293,7 @@ func ValidateGlobalnetConfiguration(globalnetInfo *GlobalnetInfo, netconfig Conf
 func GetGlobalNetworks(k8sClientset *kubernetes.Clientset, brokerNamespace string) (*GlobalnetInfo, *v1.ConfigMap, error) {
 	configMap, err := broker.GetGlobalnetConfigMap(k8sClientset, brokerNamespace)
 	if err != nil {
-		return nil, nil, fmt.Errorf("error reading configMap: %s", err)
+		return nil, nil, err
 	}
 
 	globalnetInfo := GlobalnetInfo{}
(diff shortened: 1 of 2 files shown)


pkg/discovery/globalnet/globalnet.go
```diff
@@ -293,6 +293,9 @@ func ValidateGlobalnetConfiguration(globalnetInfo *GlobalnetInfo, netconfig Conf
 func GetGlobalNetworks(k8sClientset *kubernetes.Clientset, brokerNamespace string) (*GlobalnetInfo, *v1.ConfigMap, error) {
 	configMap, err := broker.GetGlobalnetConfigMap(k8sClientset, brokerNamespace)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil, err
+		}
 		return nil, nil, fmt.Errorf("error reading configMap: %s", err)
```
sridhargaddam:
```suggestion
		return nil, nil, err
```

I feel we can just return err and let the caller check for the type of error.

mangelajo: you can also do errors.Wrap(err, "error reading configMap") which I think will preserve all the other attributes (being able to call IsNotFound, etc).

Always better to use errors.Wrapf / errors.Wrap

sridhargaddam (approved):

pkg/discovery/globalnet/globalnet.go
skitt: Yes, always use `errors.Wrap()`. Technically, this doesn’t preserve the original error’s _attributes_ directly, but it supports `Unwrap()`, which is [the documented way of handling wrapped errors](https://pkg.go.dev/errors) and [is supported by K8s’ `.Is...` functions](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors).

skitt (approved):

tpantelis (approved):

PR #1460, opened 2021-06-30
Ensure we use our new subctl

skitt:
We need to test the new subctl during deployment; our specific PATH
setting was lost with dc201779cb47 ("Remove Dockerfile.dapper"), this
restores it in the Makefile and ensures bin/subctl is built before
deployment.


dfarrell07 (approved):

skitt:
Merged with #1459.

PR #1461, opened 2021-06-30
Validate old globalnet only if globalnet enabled

vthapar:
During an upgrade we should only check if old globalnet config
is valid if globalnet is currently enabled.


tpantelis (approved):

skitt:
The upgrade still fails:

> Error validating existing globalCIDR configmap: invalid GlobalnetCidrRange: invalid CIDR address:

Jaanki (approved):

vthapar:
> The upgrade still fails:
> 
> > Error validating existing globalCIDR configmap: invalid GlobalnetCidrRange: invalid CIDR address:

Should pass now.

PR #1468, opened 2021-07-02
Restrict the route-agent capabilities

### LINKED ISSUE #1240: Investigate privilege settings for the route agent
The route agent pods are currently configured with all capabilities, privileged execution, as root with read-write root, on the host network. Some of these settings can probably be tightened; this needs to be investigated and the results at least documented (especially if they don’t lead to configuration changes).

skitt:
This uses an init container to set up entries in /proc (which requires
a privileged container), then runs the main container in
non-privileged mode (with DAC_OVERRIDE, NET_ADMIN, and NET_RAW).

This is still a PoC: the init container needs to determine what the CNI interface is, instead of hard-coding `weave`. I’m interested in @sridhargaddam and @mangelajo’s opinions on setting the default `rp_filter` to 2 to handle `vx-submariner`...

Fixes: #1240

### DIFF
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -47,17 +47,8 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner) *appsv1.DaemonSet {
 		"app": "submariner-routeagent",
 	}
 
-	allowPrivilegeEscalation := true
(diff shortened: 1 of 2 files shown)


skitt:
This also hasn’t been tested with OVN (and won’t work currently, because it configures the wrong interface).

PR #1473, opened 2021-07-08
0.9 backport: Truncate token prefixes for long cluster ids

skitt:
When an SA is created, Token Controller creates a token for it; that
token's name is auto-generated, with a random suffix, and is at most
63 characters in length, of which 5 are random. Since we look for the
generated token by name, we need to take this into account, and
truncate the prefix using the same rules as the token generator.

Fixes: #1426
(cherry picked from commit 4ea3f75d1513757b8c6dd5e664a5331fa32a4753)

### DIFF
--- a/pkg/broker/rbac.go
+++ b/pkg/broker/rbac.go
@@ -129,6 +129,10 @@ func NewBrokerRoleBinding(serviceAccount, role string) *rbacv1.RoleBinding {
 	return binding
 }
 
+// MaxGeneratedNameLength is the maximum generated length for a token, excluding the random suffix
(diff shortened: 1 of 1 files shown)


SteveMattar (approved):

tpantelis (approved):

sridhargaddam (approved):

dfarrell07 (approved):

mangelajo (approved):

dfarrell07:
Could it be this line hanging for 15m?

```
docker inspect -f {{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}} cluster2-control-plane
```

PR #1477, opened 2021-07-12
Remove mount iptables lock file

aswinsuryan:
Since the iptables is not used for setting path MTU anymore
it is removed


mangelajo (approved):
Thanks for the cleanup!!

skitt (approved):
Does this mean we can drop `iptables` from the gateway image?

aswinsuryan:
> Does this mean we can drop `iptables` from the gateway image?

Maybe yes, @mangelajo @sridhargaddam is that correct?

skitt:
> > Does this mean we can drop `iptables` from the gateway image?
> 
> Maybe yes, @mangelajo @sridhargaddam is that correct?

Let’s find out: https://github.com/submariner-io/submariner/pull/1469

PR #1484, opened 2021-07-13
Proper checks for gateways

Jaanki:
gateways, when not configured, are not nil but an empty
array. This patch does the following
1. Replaces nil check with len check
2. Uses cluster.GetGetways in show to get getways


pkg/subctl/cmd/execute.go
```diff
@@ -86,6 +86,10 @@ func (c *Cluster) GetGateways() (*submarinerv1.GatewayList, error) {
 	return gateways, nil
 }
 
+func (c *Cluster) HasNoGateways(gateways *submarinerv1.GatewayList) bool {
+	return len(gateways.Items) == 0
+}
```
tpantelis: This really shouldn't be a method on `Cluster` since it doesn't access any of its fields. However I don't think this function really adds much value anyway - it passes in an object merely to call `len` on it - it's easy enough to call `len(gateways.Items)` directly, especially since call sites access `gateways.Items` anyway.

tpantelis (changes requested):

pkg/subctl/cmd/execute.go
Jaanki: Done

tpantelis (approved):

skitt (approved):

PR #1485, opened 2021-07-13
Run Flake Finder every 2h for 0.10-rc0 testing

dfarrell07:
This frequency should be reverted once 0.10.0 RC testing is done.

Run the Flake Finder every 2h to collect a lot of data about RC0 leading
up to Test Day on 2021-07-15.


mkolesnik (approved):
Sounds reasonable to me, we need to remember to turn this down later

sridhargaddam (approved):

PR #1489, opened 2021-07-15
Migrate to RBAC v1

skitt:
rbac.authorization.k8s.io/v1beta1 is deprecated, replaced by v1 which
is available since 1.8.

See
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122
for details.

### DIFF
--- a/config/rbac/auth_proxy/auth_proxy_client_clusterrole.yaml
+++ b/config/rbac/auth_proxy/auth_proxy_client_clusterrole.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


dfarrell07:
Does this mean we should stop supporting 1.17? https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/e2e-full.yml#L28

dfarrell07 (approved):
It still runs/passes with 1.17, so if we decide we want to update that oldest-tested-version we can do it later.

skitt:
> Does this mean we should stop supporting 1.17? https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/e2e-full.yml#L28

No, the versions we rely on here have been available since 1.8 (not 1.18).

tpantelis (approved):

PR #1495, opened 2021-07-19
Include globalnet 2.0 resources in subctl gather

tpantelis:
ClusterGlobalEgressIP
GlobalEgressIP
GlobalIngressIP

### DIFF
--- a/scripts/kind-e2e/lib_subctl_gather_test.sh
+++ b/scripts/kind-e2e/lib_subctl_gather_test.sh
@@ -29,6 +29,9 @@ function validate_gathered_files () {
   validate_resource_files $subm_ns 'endpoints.submariner.io' 'Endpoint'
   validate_resource_files $subm_ns 'clusters.submariner.io' 'Cluster'
   validate_resource_files $subm_ns 'gateways.submariner.io' 'Gateway'
+  validate_resource_files all 'clusterglobalegressips.submariner.io' 'ClusterGlobalEgressIP'
+  validate_resource_files all 'globalegressips.submariner.io' 'GlobalEgressIP'
+  validate_resource_files all 'globalingressips.submariner.io' 'GlobalIngressIP'
 
   validate_pod_log_files $subm_ns '-l app=submariner-gateway'
   validate_pod_log_files $subm_ns '-l app=submariner-routeagent'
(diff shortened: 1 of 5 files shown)


sridhargaddam (commented):
This PR looks good. @tpantelis can you also capture the output of "ipset list" from the Gateway nodes similar to how we are capturing the ipsec-status command here - https://github.com/submariner-io/submariner-operator/blob/0105e4d08ef7236073547f1383566fc76c548f6d/pkg/subctl/cmd/gather/cni.go#L53

Please note, we can capture this output (i.e., ipset list) from the Globalnet Pods only as this binary will be present only on Globalnet container.

scripts/kind-e2e/lib_subctl_gather_test.sh
```diff
@@ -89,18 +92,27 @@ function validate_resource_files() {
...
+    nsarg="-A"
+  fi
 
   if [[ $cluster_name == "" ]]; then
     cluster_name=${cluster}
   fi
 
-  names=$(kubectl get $resource --namespace=$ns $selector -o=jsonpath='{.items..metadata.name}')
+  names=$(kubectl get $resource $nsarg $selector -o=jsonpath='{.items..metadata.name}')
+  namespaces=$(kubectl get $resource $nsarg $selector -o=jsonpath='{.items..metadata.namespace}')
```
vthapar: This has potential to run into problems where number of entries don't match coz some got deleted between these two `get` commands. e.g. Gather test is run at end of e2e when namespace is deleted and objects are being deleted but not all gone yet. Not sure if we can perfectly address the issue, but would be good to echo some warning message if the no. of items in these two don't match to help troubleshoot any flaky failures.

vthapar (changes requested):

scripts/kind-e2e/lib_subctl_gather_test.sh
tpantelis: I changed it to capture the whole JSON and use `jq` to parse the fields.

tpantelis:
> This PR looks good. @tpantelis can you also capture the output of "ipset list" from the Gateway nodes similar to how we are capturing the ipsec-status command here -
> 

Done.

sridhargaddam (approved):
Thanks for addressing this use-case @tpantelis

vthapar (approved):

dfarrell07:
Do we want to go ahead and merged this so it makes it in to 0.10.0-rc1? /cc @mangelajo

tpantelis:
> Do we want to go ahead and merged this so it makes it in to 0.10.0-rc1? /cc @mangelajo

Yes. We need a committer to approve.

mangelajo (approved):

PR #1503, opened 2021-07-21
Support microshift deployments, with no dns operator

mangelajo:
MicroShift[1] uses openshift coredns images, but it doesn't
enable the dns operator for footprint reasons, this means
that we need to patch the Corefile manually like for
vanilla k8s+coredns, but in the right namespaces.

Also MicroShift doesn't use the :53 port, but :5353, listening
on two ports doesn't work ok, we detect the existing Corefile
listening port, to use the specific port already being used,
defaulting to 53.

[1] https://github.com/redhat-et/microshift

### DIFF
--- a/controllers/servicediscovery/servicediscovery_controller_test.go
+++ b/controllers/servicediscovery/servicediscovery_controller_test.go
@@ -60,6 +60,91 @@ var _ = Describe("", func() {
 
 var _ = Describe("ServiceDiscovery controller tests", func() {
 	Context("Reconciliation", testReconciliation)
+	Context("coreDNS configmap parsing", func() {
+		When("no port is found", func() {
+			It("should return the default port", func() {
+				Expect(findCoreDNSListeningPort("")).To(Equal(coreDNSDefaultPort))
+			})
+		})
+
+		When("non-standard port is found and the coreconfig was already modified", func() {
+			It("should return the port", func() {
+				coreFileContents := `
+#lighthouse-start AUTO-GENERATED SECTION. DO NOT EDIT
+clusterset.local:123 {
+    forward . 10.43.188.46
+}
+#lighthouse-end
+.:456 {
+    errors
+    health
+    kubernetes cluster.local in-addr.arpa ip6.arpa {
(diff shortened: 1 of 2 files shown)


mangelajo:
@vthapar updated the patch with the extra test cases commented over slack :)

controllers/servicediscovery/servicediscovery_controller_test.go
```diff
@@ -60,6 +60,91 @@ var _ = Describe("", func() {
...
 var _ = Describe("ServiceDiscovery controller tests", func() {
 	Context("Reconciliation", testReconciliation)
+	Context("coreDNS configmap parsing", func() {
+		When("no port is found", func() {
+			It("should return the default port", func() {
+				Expect(findCoreDNSListeningPort("")).To(Equal(coreDNSDefaultPort))
+			})
+		})
+
+		When("no standard port is found and the coreconfig was already modified", func() {
```
vthapar:
```suggestion
		When("non-standard port is found and the coreconfig was already modified", func() {
```

controllers/servicediscovery/servicediscovery_controller_test.go
```diff
@@ -60,6 +60,91 @@ var _ = Describe("", func() {
...
+        policy sequential
+    }
+    cache 30
+    reload
+}
+`
+				Expect(findCoreDNSListeningPort(coreFileContents)).To(Equal("456"))
+			})
+
+			When("no standard port is found and Coreconfig was not modified (OCP coreconfig)", func() {
```
vthapar:
```suggestion
			When("non-standard port is found and Coreconfig was not modified (OCP coreconfig)", func() {
```

vthapar (dismissed):
minor nit, rest looks good

controllers/servicediscovery/servicediscovery_controller_test.go
mangelajo: Done :)

vthapar (approved):

tpantelis (approved):

PR #1500, opened 2021-07-21
Update cloud prepare commands to use new deployer

mkolesnik:
As an OCP GatewayDeployer has been added, the commands are updated to use it.

Depends on https://github.com/submariner-io/cloud-prepare/pull/72

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>

PR #1505, opened 2021-07-21
Better handling of gateways in `cloud preapre`

### LINKED ISSUE #1501: subctl cloud prepare must also open the ESP/AH protocols for all nodes (gateways and workers)
* Submariner, when using IPSEC, and Private to Private IP connectivity with no-nat is detedted
will attempt to use ESP/AH protocols for communication (more performant).
If those protocols aren't open for gateways, connecitivty will not work.

* When not using dedicated gateways (load-balancer mode), all the worker nodes need to have NATT/NAT-discovery/ESP/AH open for the same reasons.

mkolesnik:
As reported on #1501, for LB gateways we want to open the ipsec & nat discovery ports. For dedicated gateways, we also want ESP&AH protocols accepted on the gateway SG to support private-to-private tunnelling.

Depends on #1500 
Resolves: #1501

### DIFF
--- a/pkg/subctl/cmd/cloud/prepare/aws.go
+++ b/pkg/subctl/cmd/cloud/prepare/aws.go
@@ -26,9 +26,8 @@ import (
 )
 
 var (
-	gwInstanceType           string
-	gateways                 int
-	disableDedicatedGateways bool
+	gwInstanceType string
+	gateways       int
 )
 
 const DefaultDedicatedGateways = 1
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/cloud/prepare/aws.go
```diff
@@ -65,16 +56,28 @@ func prepareAws(cmd *cobra.Command, args []string) {
 		},
 	}
 
+	// For load-banalced gateways we want these ports open internally to facilitate private-ip to pivate-ip gateways communications.
+	if gateways == 0 {
+		input.InternalPorts = append(input.InternalPorts,
+			api.PortSpec{Port: nattPort, Protocol: "udp"},
+			api.PortSpec{Port: natDiscoveryPort, Protocol: "udp"},
```
mangelajo:
```suggestion
			api.PortSpec{Port: natDiscoveryPort, Protocol: "udp"},
			// ESP & AH protocols are used for private-ip to private-ip gateway communications
			{Port: 0, Protocol: "50"},
			{Port: 0, Protocol: "51"},
```

mangelajo: Here it's also necessary, if private-ip to private-ip connectivity is possible, and no nat is detected, the gateway will attempt to use ESP/AH

pkg/subctl/cmd/cloud/prepare/aws.go
```diff
@@ -65,16 +56,28 @@ func prepareAws(cmd *cobra.Command, args []string) {
 		},
 	}
 
+	// For load-banalced gateways we want these ports open internally to facilitate private-ip to pivate-ip gateways communications.
```
mangelajo:
```suggestion
	// For load-balanced gateways we want these ports open internally to facilitate private-ip to pivate-ip gateways communications.
```

mangelajo (changes requested):
small comments, linters seem to be upset 

Thanks for handling this.

pkg/subctl/cmd/cloud/prepare/aws.go
mkolesnik: these are internal ports between the submariner gateways inside the same cluster, is ESP/AH expected in this scenario? i thought only for gateway-to-gateway external communications

mkolesnik: ah nvm i see your point

mangelajo (approved):

mangelajo:
:+1:

tpantelis (approved):

PR #1511, opened 2021-07-23
Move gather.go to gather pkg

Jaanki:


pkg/subctl/cmd/gather/gather.go
```diff
@@ -185,59 +179,59 @@ func gatherDataByCluster(restConfig restconfig.RestConfig, directory string) {
...
 		return false
 	}
 
 	switch dataType {
 	case Logs:
-		gather.GatewayPodLogs(info)
-		gather.RouteAgentPodLogs(info)
-		gather.GlobalnetPodLogs(info)
-		gather.NetworkPluginSyncerPodLogs(info)
+		GatewayPodLogs(info)
```
tpantelis, line 193: Referencing  as `GatewayPodLogs` instead of `gather.GatewayPodLogs` seems awkward. I think we should prepend `gather` to each function name. However I think it's fine to keep this file where it is.

tpantelis (changes requested):

pkg/subctl/cmd/gather/gather.go
Jaanki: Agree on prepending `gather` to function name and also making the functions private.

Just trying to make dir structure clean and consistent by this change.

Jaanki:
bump

pkg/subctl/cmd/gather/gather.go
skitt: I think it’s interesting to move commands to their own packages: currently the whole `cmd` package is a bit of a mess, and various commands reference each others’ package variables (and sometimes miss important aspects, as happened at one point with contexts etc.). Moving `gather` to its own package is a step in the right direction IMO.

Jaanki:
This PR somehow lost track of its fork counterpart. Closing this in favor of https://github.com/submariner-io/submariner-operator/pull/1554

PR #1512, opened 2021-07-27
Mount xtables.lock file from host in gw pod

sridhargaddam:
Libreswan process seems to require iptables utility and since this was
missing, it was reporting errors in the logs.

...
/usr/libexec/ipsec/_stackmanager: line 111: pidof: command not found
Initializing NSS database
/usr/sbin/ipsec: line 175: iptables: command not found
nflog ipsec capture disabled
...


tpantelis (approved):

sridhargaddam:
This PR is good for review. I'm just trying to understand why libreswan requires iptables. So until then marking it as Draft.

dfarrell07 (approved):

mangelajo (approved):

sridhargaddam:
Thanks for reviewing this PR, but after further analysis its seen that iptables is used only for debugging via nflog which we are not using. So we no longer need it and the following PR will address the errors we noticed from the gateway pod - https://github.com/submariner-io/submariner/pull/1505

PR #1517, opened 2021-07-30
Add additional parameters to subctl benchmark throughput

mangelajo:
This provides additional flexibility to throughput testing,
since not two networks are the same (MTU, capacity, latency), and
sometimes we want to measure the CPU on the gateways.
```
   --parallel -P number of parallel client streams to run
   --window   -w set TCP window size in kilobytes
   --set-mss  -M set TCP MSS in bytes
   --time     -t test time in seconds
```
Depends On submariner-io/shipyard#616

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/submariner-io/admiral v0.10.0-rc0.0.20210716074658-80e46214671b
 	github.com/submariner-io/cloud-prepare v0.10.0-rc0.0.20210727150757-5e29b6472b99
 	github.com/submariner-io/lighthouse v0.10.0-rc0.0.20210713134647-2739f14330cd
-	github.com/submariner-io/shipyard v0.10.0-rc0.0.20210719115546-4e6aefed5ecd
+	github.com/submariner-io/shipyard v0.10.0-rc0.0.20210730104545-356f61a119a0
 	github.com/submariner-io/submariner v0.10.0-rc0.0.20210721111558-ee4aa093b719
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/uw-labs/lichen v0.1.4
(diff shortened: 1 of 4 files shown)


sridhargaddam:
> This provides additional flexibility to throughput testing,
> since not two networks are the same (MTU, capacity, latency), and
> sometimes we want to measure the CPU on the gateways.
> 
> ```
>    --parallel -P number of parallel client streams to run
>    --window   -w set TCP window size in kilobytes
>    --set-mss  -M set TCP MSS in bytes
>    --time     -t test time in seconds
> ```

Thank you @mangelajo for pushing the PR. But I have a feeling that if we provide these options today and in the future if we want to change the backend from iperf3 to something else, we might have issues if the new tool does not support those params. 
To make it clear, I'm not against adding these params but just wanted to discuss other options before we finalize this approach. I remember @mkolesnik had some suggestions during the initial design, so looping him in.

One other option is to take "--server-opts" and "--client-opts" from the users as part of "subctl benchmark throughput ..." and pass them as-is to the server/client. But this is not very user-friendly.

mkolesnik:
> Thank you @mangelajo for pushing the PR. But I have a feeling that if we provide these options today and in the future if we want to change the backend from iperf3 to something else, we might have issues if the new tool does not support those params.
> To make it clear, I'm not against adding these params but just wanted to discuss other options before we finalize this approach. I remember @mkolesnik had some suggestions during the initial design, so looping him in.

Yes it does indeed create a leaky abstraction which would be a problem if we want to change to a different benchmarking tool (or tools)
 
> One other option is to take "--server-opts" and "--client-opts" from the users as part of "subctl benchmark throughput ..." and pass them as-is to the server/client. But this is not very user-friendly.

We could possibly do this, which would be easier for us (but still the user would need to know it's `iperf3` behind the scenes).

Another option is to add the options but with a prefix `iperf` and document that they only work as long as `iperf` is used for benchmarking (currently that's the only option, but in the future maybe we'll have an option to choose a different backend).

sridhargaddam:
> Another option is to add the options but with a prefix `iperf` and document that they only work as long as `iperf` is used for benchmarking (currently that's the only option, but in the future maybe we'll have an option to choose a different backend).

Yeah, that's a possibility too. Another option is to support "--backend-tool" along with "--server-opts/--client-opts". In this manner, we can default to iperf3 but can continue to support additional backends in the future. The same server/client-opts params can be re-used and can be passed to the respective backend-tool.

mangelajo:
I think it's ok to leak this and stop abstracting too much. The problem with benchmarking is that there are too many network topologies, CPU types, encryptions , they will always behave differently under different types of network patterns.

In the future if we have to use other tools, and that's not supported we can deprecate, and throw a warning for those flags explaining that those are being ignored.

We could also support multiple tools if necessary.

mangelajo:
I like the "--benchpark-tool" "--server-opts" "--client-opts" , what do you guys think? @sridhargaddam @mkolesnik

sridhargaddam:
> I like the "--benchpark-tool" "--server-opts" "--client-opts" , what do you guys think? @sridhargaddam @mkolesnik

Yes, it LGTM.

maayanf24:
@sridhargaddam @mkolesnik @mangelajo - do we want to reopen this?

sridhargaddam:
I think we have to revisit this and implement the suggested approach* either in this PR or a separate PR 

[*]  "--benchmark-tool" "--server-opts" "--client-opts"

PR #1518, opened 2021-07-30
Limit cluster-id to 63 chars while joining a cluster

sridhargaddam:
ClusterID is now stored as a label on the endpoint object when its
pushed to the Broker. K8s restricts the label values to 63 chars.
So adding the corresponding check.

### DIFF
--- a/pkg/subctl/cmd/join.go
+++ b/pkg/subctl/cmd/join.go
@@ -417,6 +417,12 @@ func isValidClusterID(clusterID string) (bool, error) {
 			"'.' or '-' (and the first and last characters must be alphanumerics).\n"+
 			"%s doesn't meet these requirements", clusterID)
 	}
+
+	if len(clusterID) > 63 {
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -417,6 +417,12 @@ func isValidClusterID(clusterID string) (bool, error) {
 			"'.' or '-' (and the first and last characters must be alphanumerics).\n"+
 			"%s doesn't meet these requirements", clusterID)
 	}
+
+	if len(clusterID) > 63 {
+		return false, fmt.Errorf("Cluster ID %q has a length of %d, which is longer then the max supported"+
+			" value of 63 chars", clusterID, len(clusterID))
```
```suggestion
		return false, fmt.Errorf("The cluster ID %q has a length of %d characters which exceeds the maximum supported length of 63", clusterID, len(clusterID))
```

tpantelis (changes requested):

skitt (approved):

tpantelis (approved):

PR #1526, opened 2021-08-05
Use SHAs for all GHA versions

dfarrell07:
Per GitHub's security guidelines, GHAs should be pinned using full
length commit SHAs instead of tags.

The SHAs are of the commits currently resolved by the versions.

Even "trusted" GHAs from GitHub developers are pinned because it's
possible their repo right could be compromised and a malicious GHA
published. These core repos are not frequently substantially updated.

Submariner-internal GHAs are left pinned at devel because we want
automatic updates from Shipyard's shared tooling.

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
       - name: Delete current CODEOWNERS file
         run: rm CODEOWNERS
       - name: Run gen-codeowners to rebuild CODEOWNERS file
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 
       - name: Build all supported binaries
         run: make build-cross
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -25,7 +25,7 @@ jobs:
     if: github.repository_owner == 'submariner-io'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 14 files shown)


dfarrell07:
It seems like the update of yamllint has highlighted new issues. 

https://github.com/submariner-io/submariner-operator/pull/1526/checks?check_run_id=3256447545

skitt:
I think the switch to SHAs shouldn’t bump the versions, *i.e.* the SHA used here should be the one matching the tag. Bumping should be done separately.

skitt:
> It seems like the update of yamllint has highlighted new issues.
> 
> https://github.com/submariner-io/submariner-operator/pull/1526/checks?check_run_id=3256447545

That one’s weird, yamllint isn’t upgraded in this PR, is it?

dfarrell07:
> I think the switch to SHAs shouldn’t bump the versions, _i.e._ the SHA used here should be the one matching the tag. Bumping should be done separately.

Makes sense, I've updated the PR to the SHAs resolved by the current versions (accounting for tags like v2 resolving to 2.*). I'll send different commits/PRs to update the versions.

dfarrell07:
> > It seems like the update of yamllint has highlighted new issues.
> > https://github.com/submariner-io/submariner-operator/pull/1526/checks?check_run_id=3256447545
> 
> That one’s weird, yamllint isn’t upgraded in this PR, is it?

Yeah, good point. I can replicate this locally on devel.

dfarrell07:
> > > It seems like the update of yamllint has highlighted new issues.
> > > https://github.com/submariner-io/submariner-operator/pull/1526/checks?check_run_id=3256447545
> > 
> > 
> > That one’s weird, yamllint isn’t upgraded in this PR, is it?
> 
> Yeah, good point. I can replicate this locally on devel.

This is really weird. I sent a fix, but see https://github.com/submariner-io/submariner-operator/pull/1527#issuecomment-894408408.

dfarrell07:
> > I think the switch to SHAs shouldn’t bump the versions, _i.e._ the SHA used here should be the one matching the tag. Bumping should be done separately.
> 
> Makes sense, I've updated the PR to the SHAs resolved by the current versions. I'll send different commits/PRs to update the versions.

I updated actions/upload-artifact here: https://github.com/submariner-io/submariner-operator/pull/1529

It's actually the only GHA dependency here that needs an update I think. The others are either the latest commit (for not very active projects) or recent/latest release (for active projects, where we can expect regular releases).

skitt (approved):

tpantelis (approved):

PR #1543, opened 2021-08-13
Rename release subctl devel job

### LINKED ISSUE #201: Invalid job name in release_subctl_devel.yml when branching
When a release branch is created in the `submariner-operator` repo, it creates a PR to update the job name in _release_subctl_devel.yml_ by appending the release tag name. But this contains a '.' which GH complains about when the job attempts to run:

`The identifier 'release-subctl-release-0.10' is invalid. IDs may only contain alphanumeric characters, '', and '-'. IDs must start with a letter or '' and and must be less than 100 characters.`

We need to change the '.' to '-'.

tpantelis:
When the release process replaces `devel` with the branch name, the job fails b/c the branch name contains a '.' which is an invalid char for a GH job name. So use `cutting-edge` in the job name instead of `devel`. Also renamed the yaml file accordingly.

Fixes https://github.com/submariner-io/releases/issues/201


.github/workflows/release_subctl_devel.yml
```diff
@@ -1,14 +1,13 @@
 ---
-name: Release Subctl Devel
+name: Release Subctl Cutting Edge
 
 on:
   push:
     branches:
       - devel
-      - release-*
```
tpantelis, line 8: This doesn't seem necessary since `devel` is replaced by the actual release branch name when the branch is created.

dfarrell07 (changes requested):
I don't understand why this fixes the issue of the stable branch name having a dot. The issue isn't "devel" I think, it's what we're replacing it with ("release-0.10"). As I understand it, this would just make the release automation fail to rename the job (it would look for devel, but now it's cutting-edge).

In general, I think we should be trying to standardize on naming things "devel" instead of mix of mostly-devel and a few "cutting-edge".

tpantelis:
> I don't understand why this fixes the issue of the stable branch name having a dot. The issue isn't "devel" I think, it's what we're replacing it with ("release-0.10"). 

Yes

> As I understand it, this would just make the release automation fail to rename the job (it would look for devel, but now it's cutting-edge).
> In general, I think we should be trying to standardize on naming things "devel" instead of mix of mostly-devel and a few "cutting-edge".

It does a simple search and replace of "devel" with the release branch name throughout the entire _.github/workflow_ directory. The resulting invalid name for this job is a side-effect of that. One option is to modify the release automation so it doesn't replace this instance of "devel". But, AFAIK, the name of the GH job doesn't really matter and doesn't need to reflect the actual branch name so the simplest thing to do is just to remove "devel" from the job name. To me, `cutting-edge` seems appropriate since the job creates the "Cutting Edge" release of the branch. Also the workflow name contains "Devel" which isn't replaced which is inconsistent.

dfarrell07:
If the goal is to make this workflow unrelated to devel, something that applies to all branches without a rename, then let's call it something like "release-subctl-update". The actual difference between this workflow and the "release-subctl" workflow is that this one updates an existing release vs creating a new one.

tpantelis:
> If the goal is to make this workflow unrelated to devel, something that applies to all branches without a rename, then let's call it something like "release-subctl-update". The actual difference between this workflow and the "release-subctl" workflow is that this one updates an existing release vs creating a new one.

The goal is to fix the issue of the invalid job name by renaming it to not contain "devel". I'm not picky as to what we name it - "update" sounds fine to me. Let's see what others  (@skitt @mkolesnik ) think.

.github/workflows/release_subctl_cutting_edge.yml
```diff
@@ -1,14 +1,13 @@
 ---
-name: Release Subctl Devel
+name: Release Subctl Cutting Edge
```
skitt:
```suggestion
name: Release Subctl devel
```

.github/workflows/release_subctl_cutting_edge.yml
```diff
@@ -1,14 +1,13 @@
...
 
 on:
   push:
     branches:
       - devel
-      - release-*
 
 jobs:
-  release-subctl-devel:
+  release-subctl-cutting-edge:
```
skitt:
```suggestion
  release-subctl-on-push:
```

skitt (changes requested):

dfarrell07 (approved):

mangelajo (approved):

skitt (approved):

PR #1546, opened 2021-08-17
Verify that requested CRDs are generated

skitt:
controller-gen can complete successfully without generating the files
we expect, e.g. if the CRD type definitions are moved out of the
execution scope. This results in failure further down the build;
instead of waiting for that, this patch adds explicit checks for the
existence of the CRD we expect, and fails if it's missing.


tpantelis (approved):

sridhargaddam (approved):

dfarrell07 (approved):

dfarrell07:
So this is catching an existing issue?

```
test -f deploy/crds/submariner.io_servicediscoveries.yaml
make: *** [Makefile:161: deploy/crds/submariner.io_servicediscoveries.yaml] Error 1
```

mangelajo (approved):

PR #1558, opened 2021-09-07
build(bundle): use default values

SteveMattar:
VERSION, FROM_VERSION and CHANNEL will have default values calculated
from git tags when invalid or not specified.


dfarrell07 (approved):
LGTM, thanks @SteveMattar!

mangelajo (approved):

PR #1560, opened 2021-09-07
Move broker ns creation to apis

Jaanki:
....so that submariner-addon can use it.

Signed-off-by: Jaanki <jchhatba@redhat.com>

### DIFF
--- a/apis/submariner/manifests/namespace.go
+++ b/apis/submariner/manifests/namespace.go
@@ -16,21 +16,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package broker
+package manifests
 
 import (
+	"context"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-const (
-	SubmarinerBrokerNamespace = "submariner-k8s-broker"
-)
+func CreateNewBrokerNamespace(clientset *kubernetes.Clientset, namespace string) (brokernamespace *v1.Namespace, err error) {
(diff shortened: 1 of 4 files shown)


apis/submariner/manifests/namespace.go
```diff
@@ -16,21 +16,24 @@ See the License for the specific language governing permissions and
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-const (
-	SubmarinerBrokerNamespace = "submariner-k8s-broker"
-)
+func CreateNewBrokerNamespace(clientset *kubernetes.Clientset, namespace string) (brokernamespace *v1.Namespace, err error) {
+	return clientset.CoreV1().Namespaces().Create(context.TODO(), NewBrokerNamespace(namespace), metav1.CreateOptions{})
+}
```
tpantelis, line 31: This is simply a direct wrapper for the passed clientset,.  I don't think this function is really necessary and certainly  not as API to import from the add-on.

tpantelis (changes requested):

skitt (changes requested):
There already is a public API for broker creation: the `Broker` CRD. To move broker namespace creation to the operator, we shouldn’t need any changes to the operator itself; the addon can be changed to create a `Broker` object as appropriate.

Jaanki:
> There already is a public API for broker creation: the `Broker` CRD. To move broker namespace creation to the operator, we shouldn’t need any changes to the operator itself; the addon can be changed to create a `Broker` object as appropriate.

That would require installing submariner-operator on hub cluster. Are we okay with that dependency? Currently submariner-addon installs submariner-operator on managedclusters but not on the hub cluster.

PR #1603, opened 2021-10-11
test(bundle): fix periodic test

SteveMattar:
The default broker's client SA is being shared across all
the connected clusters when deploying using the bundle method.
Unlike subctl the bundle creates all the resources during deployment
and before the join command, therefore the process cannot predict
the creation of a SA per connected cluster.
This should be handled by the operator.
Meanwhile, this specific validation test will be disabled.
The subctl gather test requires lighthouse.

Depends on submariner-io/shipyard/pull/647

### DIFF
--- a/scripts/kind-e2e/lib_operator_verify_subm.sh
+++ b/scripts/kind-e2e/lib_operator_verify_subm.sh
@@ -197,8 +197,8 @@ function verify_subm_cr() {
   validate_equals '.kind' 'Submariner'
   validate_equals '.metadata.name' $deployment_name
   validate_equals '.spec.brokerK8sApiServer' $SUBMARINER_BROKER_URL
-  # every cluster must have it's own token / SA
-  validate_not_equals '.spec.brokerK8sApiServerToken' $SUBMARINER_BROKER_TOKEN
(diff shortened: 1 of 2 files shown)


skitt (approved):

dfarrell07 (approved):

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -197,8 +197,8 @@ function verify_subm_cr() {
   validate_equals '.kind' 'Submariner'
   validate_equals '.metadata.name' $deployment_name
   validate_equals '.spec.brokerK8sApiServer' $SUBMARINER_BROKER_URL
-  # every cluster must have it's own token / SA
-  validate_not_equals '.spec.brokerK8sApiServerToken' $SUBMARINER_BROKER_TOKEN
+  # TODO: every cluster must have it's own token / SA (not working when using bundle/acm)
```
dfarrell07, line 200: Should we create an issue to track this technical debt?

SteveMattar: yes I will create

PR #1627, opened 2021-10-19
Update K8s versions: drop 1.17/1.18, add 1.21/1.22

skitt:
1.17 and 1.18 have reached EOL.


Jaanki (approved):

tpantelis (approved):

.github/workflows/e2e-full.yml
```diff
@@ -25,7 +25,9 @@ jobs:
             globalnet: 'globalnet'
         include:
           # This is the oldest K8s version we try to support
-          - k8s_version: '1.17'
+          - k8s_version: '1.19'
```
mkolesnik, line 29: I thought line 18 was the newest?

.github/workflows/flake_finder.yml
```diff
@@ -16,18 +16,18 @@ jobs:
...
-        k8s_version: ['1.17']
+        k8s_version: ['1.19']
         lighthouse: ['', 'lighthouse']
         include:
-          - k8s_version: '1.18'
-          - k8s_version: '1.18'
-            globalnet: 'globalnet'
           - k8s_version: '1.19'
           - k8s_version: '1.19'
             globalnet: 'globalnet'
```
mkolesnik:
```suggestion
```

These lines are redundant now

mkolesnik (changes requested):

.github/workflows/e2e-full.yml
skitt: Isn’t that the default?

dfarrell07 (approved):

.github/workflows/e2e-full.yml
dfarrell07: This does highlight that commit/PR titles are a bit confusing, as the E2E default is actually unchanged at 1.20. We're also adding coverage for both 1.21 and 1.22. Something like "Add E2E for K8s 1.21-22, drop 1.17-18" might be more clear.

mkolesnik: I think if you just list all the version here explicitly, GHA is smart enough not to run the "deault" twice.

dfarrell07: I may not understand what you're saying @mkolesnik. Adding K8s versions in `include`s gives jobs with only the K8s version non-default. The default K8s version is used in all the jobs defined by the matrix of the other params. Adding another `1.20` include would duplicate the job in the main matrix that uses all defaults.

mkolesnik: > Adding another `1.20` include would duplicate the job in the main matrix that uses all defaults.

That's what I mean, it won't. GH is smart enough not to duplicate it.

However if we have all versions listed explicitly, we won't have to edit this list (or forget to) when we change the `1.20` default to some other version.

Hope this is clearer :)

dfarrell07: Ah okay, I see the underlying idea now. I hear you regarding that helping maintenance in a way, but I also (as you see) find it confusing to define duplicate jobs but rely on GHAs to not run them multiple times. An inline comment could mitigate that, though.

If we do this strategy, I guess it would be least-confusing to do it consistently, so for now I think it shouldn't be a part of this PR.

mkolesnik (approved):

.github/workflows/e2e-full.yml
mkolesnik: Yes but it's not "rely on GHAs to not run them multiple times" but rather the way GHA is meant to work. You can read it on their manual.

You can add extra parameters this way to an existing job, it won't duplicate it.

In our case we're not adding anything so it just keeps that one job there as is, it won't duplicate it either way.

PR #1672, opened 2021-11-22
Support configuring insecure TLS to the broker

### LINKED ISSUE #1671: The operator should be capable of configuring insecure access to the broker
See https://github.com/submariner-io/submariner-charts/issues/185 for the use-case.

Admiral supports insecure TLS, but the operator can’t configure it.

skitt:
Fixes: #1671

### DIFF
--- a/api/submariner/v1alpha1/submariner_types.go
+++ b/api/submariner/v1alpha1/submariner_types.go
@@ -58,6 +58,7 @@ type SubmarinerSpec struct {
 	NatEnabled               bool                 `json:"natEnabled"`
 	LoadBalancerEnabled      bool                 `json:"loadBalancerEnabled,omitempty"`
 	ServiceDiscoveryEnabled  bool                 `json:"serviceDiscoveryEnabled,omitempty"`
+	BrokerK8sInsecure        bool                 `json:"brokerK8sInsecure,omitempty"`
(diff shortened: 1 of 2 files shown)


tpantelis (approved):

dfarrell07 (approved):

api/submariner/v1alpha1/submariner_types.go
```diff
@@ -58,6 +58,7 @@ type SubmarinerSpec struct {
 	NatEnabled               bool                 `json:"natEnabled"`
 	LoadBalancerEnabled      bool                 `json:"loadBalancerEnabled,omitempty"`
 	ServiceDiscoveryEnabled  bool                 `json:"serviceDiscoveryEnabled,omitempty"`
+	BrokerK8sInsecure        bool                 `json:"brokerK8sInsecure"`
```
tpantelis: Perhaps this is why the upgrade test is failing?

```suggestion
	BrokerK8sInsecure        bool                 `json:"brokerK8sInsecure, omitempty"`
```

tpantelis (changes requested):

api/submariner/v1alpha1/submariner_types.go
skitt: Yes, I wanted to see how the default was handled without that...

tpantelis (approved):

PR #1674, opened 2021-11-25
Eliminate extraneous reconciliation controllers

tpantelis:
The `SubmarinerReconciler` and `ServiceDiscoveryReconciler` create a controller instance via `controller.New` and another extraneous incomplete instance via `ctrl.NewControllerManagedBy` in `SetupWithManager`. The latter was added by [this commit](https://github.com/submariner-io/submariner-operator/pull/724/commits/a9004d71da4e981264efcf77d8b54a90ab95b608) during a refactor - it seems like it was a mistake. The `ctrl.NewControllerManagedBy` API is a simpler builder wrapper for the `controller.New` API so let's use the former and, more importantly, create only one controller instance.


skitt (approved):

SteveMattar (approved):

tpantelis:
@skitt I think this explains the strange intermittent behavior we've seen, like conflicting updates, that indicated there were 2 controllers running.

PR #1680, opened 2021-11-29
Centralise RestConfig handling

skitt:
This moves the configuration-handling specifics into restconfig, which
will make it easier to switch to clientcmd while preserving
backwards-compatibility.

Depends on #1681

### DIFF
--- a/cmd/subctl/deploybroker.go
+++ b/cmd/subctl/deploybroker.go
@@ -39,14 +39,14 @@ var deployBroker = &cobra.Command{
 	Short: "Deploys the broker",
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println("deployBroker called")
-		err := deploy.Broker(&deployflags, kubeConfig, kubeContext)
+		err := deploy.Broker(&deployflags, restConfigProducer)
 		exit.OnError("Error deploying Broker", err)
 	},
 }
 
 func init() {
 	addDeployBrokerFlags()
-	addKubeContextFlag(deployBroker)
+	restConfigProducer.AddKubeContextFlag(deployBroker)
 	rootCmd.AddCommand(deployBroker)
 }
 
--- a/cmd/subctl/root.go
+++ b/cmd/subctl/root.go
@@ -18,12 +18,10 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/submariner-io/submariner-operator/internal/restconfig"
 )
 
-var (
-	kubeConfig  string
-	kubeContext string
-)
+var restConfigProducer = restconfig.NewProducer()
 
 // rootCmd represents the base command when called without any subcommands.
 var rootCmd = &cobra.Command{
@@ -38,15 +36,4 @@ func Execute() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-}
-
-// addKubeConfigFlag adds a "kubeconfig" flag.
-func addKubeConfigFlag(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVar(&kubeConfig, "kubeconfig", "", "absolute path(s) to the kubeconfig file(s)")
-}
-
-// addKubeContextFlag adds a "kubeconfig" flag and a single "kubecontext" flag that can be used once and only once.
-func addKubeContextFlag(cmd *cobra.Command) {
-	addKubeConfigFlag(cmd)
-	cmd.PersistentFlags().StringVar(&kubeContext, "kubecontext", "", "kubeconfig context to use")
 }
--- a/internal/restconfig/restconfig.go
+++ b/internal/restconfig/restconfig.go
@@ -21,10 +21,15 @@ package restconfig
 import (
 	"fmt"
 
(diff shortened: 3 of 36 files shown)


dfarrell07 (approved):

pkg/subctl/cmd/utils/restconfig/restconfig.go
```diff
@@ -40,24 +43,82 @@ type RestConfig struct {
 	ClusterName string
 }
 
-func MustGetForClusters(kubeConfigPath string, kubeContexts []string) []RestConfig {
-	configs, err := ForClusters(kubeConfigPath, kubeContexts)
+type RestConfigProducer struct {
```
tpantelis: Since "_restconfig_" is the package name.

```suggestion
type Producer struct {
```

pkg/subctl/cmd/utils/restconfig/restconfig.go
```diff
@@ -40,24 +43,82 @@ type RestConfig struct {
...
+	kubeConfig   string
+	kubeContext  string
+	kubeContexts []string
+}
+
+func NewRestConfigProducer() RestConfigProducer {
```
tpantelis:
```suggestion
func NewProducer() RestConfigProducer {
```

pkg/subctl/cmd/utils/restconfig/restconfig.go
```diff
@@ -40,24 +43,82 @@ type RestConfig struct {
...
+	return RestConfigProducer{}
+}
+
+func NewFromKubeConfig(kubeConfig, kubeContext string) RestConfigProducer {
```
tpantelis:
```suggestion
func NewProducerFrom(kubeConfig, kubeContext string) RestConfigProducer {
```

tpantelis (changes requested):

tpantelis (approved):

dfarrell07:
Do we want this rebased on top of https://github.com/submariner-io/submariner-operator/pull/1681? Seems accidental?

skitt:
> Do we want this rebased on top of #1681? Seems accidental?

Earlier versions of this PR broke some `subctl` commands, I want to be as confident as possible that the merged version doesn’t; that’s why I added #1681 and made sure the tests are run on this PR as well.

tpantelis:
@skitt I rebased and fixed conflicts. I **think** I did it all correctly but please confirm.

pkg/subctl/cmd/utils/restconfig/restconfig.go
```diff
@@ -39,24 +44,85 @@ type RestConfig struct {
 	ClusterName string
 }
 
-func MustGetForClusters(kubeConfigPath string, kubeContexts []string) []RestConfig {
-	configs, err := ForClusters(kubeConfigPath, kubeContexts)
+type Producer struct {
```
tpantelis, line 47: With @Jaanki efforts to restructure/refactor the code base, perhaps we should move/put this under _pkg/restconfig_ (or _cmd/subctl/restconfig_  since this seems specific to CLI) to avoid having to refactor it later.

skitt: I moved it to `internal/restconfig`.

tpantelis: The code references `cobra.Command` which I think seems more suited under _cmd_  but it's fine either way.

skitt: AIUI, `cmd` is supposed to be kept as small as possible, with all the actual code under `internal` or `pkg`.

tpantelis: yeah it's reasonable to keep `cmd` relatively small with the non-CLI related elsewhere so it can be reused otherwise. But the `Producer` is tied to `kubeConfig` and `kubeContext` and references `cobra.Command` which seem pertinent to CLI and leans me towards _cmd/restconfig_.

On a side note, we talked about using this for `deploy.Broker` under _pkg/deploy/broker_  but, to make it generally reusable (eg for the **submariner-addon**) and also "unit-testable", we should be passing around client set interfaces instead.

skitt: See https://github.com/golang-standards/project-layout#cmd for where this is coming from. `cmd/restconfig` would suggest we’re building a `restconfig` command, which isn’t the case here.

Jaanki: I agree. This should not be in `cmd/subctl` but rather `internal/restconfig` seems a good fit.

tpantelis: OK.  Do we anticipate external apps using this in which case it should be under _pkg/restconfig_?

skitt: This should stay confined to `subctl` for now IMO.

tpantelis: OK - _internal/restconfig_ it is. We can always move it later if need be.

PR #1686, opened 2021-12-03
Support insecure connections to the broker in SD

skitt:
Propagate support for insecure connections in service discovery.

### DIFF
--- a/api/submariner/v1alpha1/servicediscovery_types.go
+++ b/api/submariner/v1alpha1/servicediscovery_types.go
@@ -40,6 +40,7 @@ type ServiceDiscoverySpec struct {
 	Version                  string               `json:"version,omitempty"`
 	Debug                    bool                 `json:"debug"`
 	GlobalnetEnabled         bool                 `json:"globalnetEnabled,omitempty"`
+	BrokerK8sInsecure        bool                 `json:"brokerK8sInsecure,omitempty"`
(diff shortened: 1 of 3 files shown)


tpantelis (approved):

Jaanki (approved):

sridhargaddam (approved):

m4ce:
@skitt @sridhargaddam  - we tried this change but it's not working. Where is this option actually used in the code?

m4ce:
should it not be supported in

```func ForBroker(submariner *v1alpha1.Submariner, serviceDisc *v1alpha1.ServiceDiscovery) (*rest.Config, string, error)``` in ```submariner.Spec.BrokerK8sCA, &rest.TLSClientConfig{}```?

skitt:
The option is used in Admiral to set up the Broker syncer, but I messed it up and the setting is passed incorrectly, sorry. I’m fixing that right now.

The `ForBroker` function is used in `subctl` only; it shouldn’t need an “insecure” flag (but I’m working on a separate set of changes which will allow that too).

skitt:
Fixed with https://github.com/submariner-io/submariner-operator/pull/1692 (at least, the setting will be passed correctly).

m4ce:
Thanks @skitt - let us know when it's ready to be tested. We are quite eager to try it out asap.

Hitesh-Agrawal:
@skitt Waiting to test the same. Pls share the details on which submariner-operator helm chart to use and with which docker image.

PR #1695, opened 2021-12-08
Added support for weights annotations for subctl export command

danibachar:
Signed-off-by: Daniel Bachar <danibachar89@gmail.com>

After debating a bit in the group I have decided to remove the need for a new CRD and support weights distribution through the ServiceExport annotations as was originally suggested.

This PR is directly related to - https://github.com/submariner-io/lighthouse/pull/652. While it does not depend on the former, it allows extending the `subsctl export` command to support the extended API in https://github.com/submariner-io/lighthouse/pull/652

The explanation is described in the EP - submariner-io/enhancements#2
Link to the explanation - https://github.com/danibachar/enhancements/blob/feature/lighthouse/lighthouse/service-selection/multi-cluster-service-selection.md#serviceweightpolicy-distribution

### DIFF
--- a/pkg/subctl/cmd/export.go
+++ b/pkg/subctl/cmd/export.go
@@ -23,6 +23,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 	autil "github.com/submariner-io/admiral/pkg/util"
@@ -49,6 +51,7 @@ var (
 		Run:     exportService,
 	}
 	serviceNamespace string
+	rawWeightsString string
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/subctl/cmd/export.go
```diff
@@ -114,3 +122,28 @@ func validateArguments(args []string) error {
...
+		}
+		clusterID := weightClusterTuple[0]
+		weightStr := weightClusterTuple[1]
+		// Validating integer support only
+		if _, err := strconv.ParseInt(weightStr, 0, 64); err != nil {
+			err = fmt.Errorf("Error: %v parsing the weight %q for cluster %q, it is probably not of intenger format for the tuple %q ", err, clusterID, weightStr, weightClusterString)
+			break
+		}
+
+		annotations["lighthouse-load-balancer.submariner.io/weight"+"/"+clusterID] = weightStr
```
tpantelis, line 145: To be consistent with other LH annotations:

```suggestion
		annotations["lighthouse.submariner.io/load-balancer-weights"+"/"+clusterID] = weightStr
```

pkg/subctl/cmd/export.go
```diff
@@ -60,6 +63,7 @@ func init() {
 
 func addServiceExportFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "namespace of the service to be exported")
+	cmd.Flags().StringVarP(&rawWeightsString, "clustersWeight", "cw", "", "weights for each source cluster, supports integr weights only. Formatted as <cluster:weight,cluster:weight> for example cluster1:10,cluster2:90")
```
tpantelis, line 66: Rather than a comma-separated list in a single param, how about allowing multiple params, eq

`--cluster-weight east:10 --cluster-weight west:20`

tpantelis (changes requested):

PR #1702, opened 2021-12-11
feat(bundle): support related-images

### LINKED ISSUE #1427: Overriding images using Environment variables
Today we support overriding images during build and runtime but not during the installation process.
I propose adding an environment variable for each component in the deployment YAML which contains an image url,
The operator will then use it to override the default image.

The priority will be in this order:

1. `image-override` in Submariner CR
2. The deployment environment variable
3. Using the `version` & `repo` values in  Submariner CR
4. The default image

**Why is this needed**:
When building the bundle as part of the CI. I would like to pin the image digest for all components belong to the same build.
This way I can make sure the bundle does not use the latest images according to the regular floating tags but uses the images that I expect. In the end we would like to make sure what QA tests is what will be released.

SteveMattar:
**Note**: the file `config/manager/patches/related-images.deployment.config.yaml` should have different content in the downstream environment.

Closes: #1427 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,7 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) kustomization
 	(cd config/bundle && $(KUSTOMIZE) edit add resource ../../bundle/manifests/submariner.clusterserviceversion.yaml && cd ../../) && \
 	$(KUSTOMIZE) build config/bundle/ --load_restrictor=LoadRestrictionsNone --output bundle/manifests/submariner.clusterserviceversion.yaml && \
 	sed -i -e 's/$$(SHORT_VERSION)/$(SHORT_VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml && \
+	sed -i -e 's/$$(VERSION)/$(VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml && \
 	$(OPERATOR_SDK) bundle validate ./bundle
 
 # Generate package manifests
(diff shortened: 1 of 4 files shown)


dfarrell07 (approved):

skitt (approved):

Makefile
```diff
@@ -268,6 +268,7 @@ bundle: $(KUSTOMIZE) $(OPERATOR_SDK) kustomization
 	(cd config/bundle && $(KUSTOMIZE) edit add resource ../../bundle/manifests/submariner.clusterserviceversion.yaml && cd ../../) && \
 	$(KUSTOMIZE) build config/bundle/ --load_restrictor=LoadRestrictionsNone --output bundle/manifests/submariner.clusterserviceversion.yaml && \
 	sed -i -e 's/$$(SHORT_VERSION)/$(SHORT_VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml && \
+	sed -i -e 's/$$(VERSION)/$(VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml && \
```
mkolesnik, line 271:
```suggestion
	sed -i -e 's/$${VERSION}/$(VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml && \
```

Did you mean this? I suspect it'll fail otherwise

Makefile
```diff
@@ -277,6 +278,7 @@ packagemanifests: $(OPERATOR_SDK) $(KUSTOMIZE) kustomization
 	(cd config/bundle && $(KUSTOMIZE) edit add resource ../../packagemanifests/$(BUNDLE_VERSION)/submariner.clusterserviceversion.yaml && cd ../../) && \
 	$(KUSTOMIZE) build config/bundle/ --load_restrictor=LoadRestrictionsNone --output packagemanifests/$(BUNDLE_VERSION)/submariner.clusterserviceversion.yaml && \
 	sed -i -e 's/$$(SHORT_VERSION)/$(SHORT_VERSION)/g' packagemanifests/$(BUNDLE_VERSION)/submariner.clusterserviceversion.yaml && \
+	sed -i -e 's/$$(VERSION)/$(VERSION)/g' packagemanifests/$(BUNDLE_VERSION)/submariner.clusterserviceversion.yaml && \
```
mkolesnik, line 281:
```suggestion
	sed -i -e 's/$${VERSION}/$(VERSION)/g' packagemanifests/$(BUNDLE_VERSION)/submariner.clusterserviceversion.yaml && \
```

Same here I guess?

mkolesnik (changes requested):
Couple of places I suspect there might be an issue, unless I'm missing something?

Makefile
skitt: `$$(VERSION)` is correct, the goal is to replace the token (which is `$(VERSION)`) in the YAML files, not replace a given value with another value.

mkolesnik (approved):

nyechiel:
@SteveMattar can you help me understand where this change should be documented? Thanks!

SteveMattar:
@nyechiel this is only relevant when deploying using the bundle. Nothing needs to be done by the user.
The related images are pinned automatically in the CSV.
This change is mostly relevant to the downstream environment.

PR #1706, opened 2021-12-13
Add client Producer interface and implementations

tpantelis:
...that yields various client set interfaces used throughout the code base. This provides a layer of abstraction in lieu of passing around `rest.Config` instances to enable easier reuse for external apps, eg the **submariner-addon**, and to facilitate unit testing.

### DIFF
--- a/pkg/client/default_producer.go
+++ b/pkg/client/default_producer.go
@@ -1,0 +1,49 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dfarrell07 (approved):

tpantelis:
@dfarrell07 Thanks for the prompt review.  + @skitt @Jaanki  This is a quick proposal but, looking at the code some more, most `Ensure` functions use only one client set, a few  use two, and I believe only one needs all four. So I think this abstraction may be a bit overkill, ie I think we can just pass the pertinent client set(s) to each function - in fact some lower level functions already take client sets. This also avoids having to instantiate client sets that may not be used.

tpantelis:
> @dfarrell07 Thanks for the prompt review. + @skitt @Jaanki This is a quick proposal but, looking at the code some more, most `Ensure` functions use only one client set, a few use two, and I believe only one needs all four. So I think this abstraction may be a bit overkill, ie I think we can just pass the pertinent client set(s) to each function - in fact some lower level functions already take client sets. This also avoids having to instantiate client sets that may not be used.

Exploring further, I think this abstraction would be useful at a higher lever, eg `deploy.Broker` which needs all the client sets.  From there, the lower level `Ensure` functions can take the specific client set type that they need (again most need only one). I'll push another PR to demonstrate this with `deploy.Broker`.

skitt (approved):

PR #1710, opened 2021-12-16
Copy the local build setup from Submariner

skitt:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,17 @@ DEFAULT_IMAGE_VERSION ?= $(BASE_BRANCH)
 export BASE_BRANCH
 export DEFAULT_IMAGE_VERSION
 
+# Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

tpantelis (approved):

mkolesnik:
@skitt wouldn't it be better to add this to a shared place, e.g. `Makefile.dapper` in Shipyard (yes I know it's a bit ironic, but it fits the case), instead of copying it all over?

skitt:
> @skitt wouldn't it be better to add this to a shared place, e.g. `Makefile.dapper` in Shipyard (yes I know it's a bit ironic, but it fits the case), instead of copying it all over?

@mkolesnik it would, yes, and I was hoping to find a nice way (which is why I haven’t done my usual scripted set of PRs against all the repos). If we move the `Makefile.dapper` inclusion to the top of the main `Makefile` it might just work, I’ll give that a shot.

PR #1713, opened 2021-12-16
Use order-only prerequisites for tools

skitt:
Currently, the tools used during the build (e.g. controller-gen) are
rebuilt if their dependencies change, and then the files generated by
the tools are rebuilt because the tool was rebuilt. This isn't
desirable in day-to-day use; to avoid having tool rebuilds cause
regeneration of all the files and thus recompilation of everything in
most cases, this patch changes concrete targets to specify the
required tools using order-only prerequisites.


dfarrell07 (approved):
If anyone else is wondering: https://stackoverflow.com/questions/58039810/makefiles-what-is-an-order-only-prerequisite

dfarrell07:
This won't somehow hide updates that _should_ happen when prerequisites change, right?

skitt:
> This won't somehow hide updates that _should_ happen when prerequisites change, right?

I hesitated to discuss that in the PR; for CI builds this isn’t an issue, since the tools and generated files are always rebuilt. But when building locally, yes, this could hide updates that should happen; but if that causes problems, they’ll end up being caught by CI.

I should add a `clean` target which removes all the locally-installed tools and generated files.

mkolesnik (approved):

tpantelis (approved):

PR #1717, opened 2021-12-20
Refactor deploy.Broker to use the client.Producer

tpantelis:
This requires changing the underlying `Ensure` functions to take client interfaces instead of `rest.Config` and moving other components under _pkg/internal_.  See individual commits for details.

Depends on #1705

### DIFF
--- a/cmd/subctl/deploybroker.go
+++ b/cmd/subctl/deploybroker.go
@@ -23,15 +23,15 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/submariner-io/submariner-operator/internal/cli"
+	"github.com/submariner-io/submariner-operator/internal/component"
 	"github.com/submariner-io/submariner-operator/internal/constants"
 	"github.com/submariner-io/submariner-operator/internal/exit"
 	"github.com/submariner-io/submariner-operator/pkg/deploy"
-	"github.com/submariner-io/submariner-operator/pkg/subctl/components"
 )
 
 var (
 	deployflags       deploy.BrokerOptions
-	defaultComponents = []string{components.ServiceDiscovery, components.Connectivity}
+	defaultComponents = []string{component.ServiceDiscovery, component.Connectivity}
 )
 
 // deployBroker represents the deployBroker command.
--- a/cmd/subctl/deploybroker.go
+++ b/cmd/subctl/deploybroker.go
@@ -43,20 +43,20 @@ var deployBroker = &cobra.Command{
 	Use:   "deploy-broker",
 	Short: "Deploys the broker",
 	Run: func(cmd *cobra.Command, args []string) {
+		status := cli.NewReporter()
+
 		config, err := restConfigProducer.ForCluster()
-		exit.OnError("Error creating REST config", err)
+		exit.OnError(status.Error(err, "Error creating REST config"))
 
 		clientProducer, err := client.NewProducerFromRestConfig(config)
-		exit.OnError("Error creating client producer", err)
+		exit.OnError(status.Error(err, "Error creating client producer"))
 
-		status := cli.NewReporter()
 		err = deploy.Broker(&deployflags, clientProducer, status)
-		if err == nil {
-			err = broker.WriteInfoToFile(config, deployflags.BrokerNamespace, ipsecSubmFile,
-				stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
-		}
+		exit.OnError(err)
 
-		exit.OnError("", err)
+		err = broker.WriteInfoToFile(config, deployflags.BrokerNamespace, ipsecSubmFile,
+			stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
+		exit.OnError(err)
 	},
 }
 
--- a/cmd/subctl/deploybroker.go
+++ b/cmd/subctl/deploybroker.go
@@ -22,15 +22,18 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/submariner-operator/internal/cli"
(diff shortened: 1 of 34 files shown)


skitt (approved):

pkg/broker/file.go
```diff
@@ -0,0 +1,136 @@
...
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		// TODO return reporter.Error(err, "error creating Kubernetes client")
+		return errors.Wrap(err, "error creating Kubernetes client")
+	}
+
+	data, err := newDataFrom(kubeClient, brokerNamespace, ipsecFile)
+	if err != nil {
+		// TODO return reporter.Error(err, "error initializing broker info")
```
Jaanki, line 51: Remove this TODO

pkg/broker/file.go
```diff
@@ -0,0 +1,136 @@
...
+		return err
+	}
+
+	data.BrokerURL = restConfig.Host + restConfig.APIPath
+
+	newFilename, err := backupIfExists(InfoFileName)
+	if err != nil {
+		// TODO return reporter.Error(err, "error backing up the broker file")
```
Jaanki: Ditto

cmd/subctl/deploybroker.go
```diff
@@ -39,7 +42,14 @@ var deployBroker = &cobra.Command{
 	Use:   "deploy-broker",
 	Short: "Deploys the broker",
 	Run: func(cmd *cobra.Command, args []string) {
-		err := deploy.Broker(&deployflags, restConfigProducer, cli.NewReporter())
+		status := cli.NewReporter()
+		err := deploy.Broker(&deployflags, restConfigProducer, status)
+		if err == nil {
```
Jaanki: reworking the logic to:

```
err := deploy.Broker(&deployflags, restConfigProducer, status)
exit.OnError("Error deploying Broker", err)

configs, _ := restConfigProducer.ForCluster()
			err = broker.WriteInfoToFile(configs, deployflags.BrokerNamespace, ipsecSubmFile,
				stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
		}
```

pkg/deploy/broker.go
```diff
@@ -51,50 +50,45 @@ var ValidComponents = []string{component.ServiceDiscovery, component.Connectivit
 // Ignoring th cyclic complexity of Broker function because it is being refactored in
 // https://github.com/submariner-io/submariner-operator/pull/1717.
 //gocyclo:ignore
```
Jaanki: unignore this check

Jaanki (changes requested):

Jaanki (commented):
For the commit `Modify cli.Status to log directly if status not set` I don't think this was necessary because there are cases where the status.Start/End happens in the calling functions.

pkg/broker/file.go
tpantelis: yeah I thought I had done that

tpantelis:
> For the commit `Modify cli.Status to log directly if status not set` I don't think this was necessary because there are cases where the status.Start/End happens in the calling functions.

There are places in `deploy.Broker` where we return an error w/o invoking the reporter and let the caller log it via `exit.OnError.` I think we should be consistent and always notify the reporter. Also in those cases and others, we don't really need a Start/End block. This provides the flexibility to log a series of operations via a Start/End block or just log a single operation.

pkg/deploy/broker.go
tpantelis: yeah I meant to do that.

cmd/subctl/deploybroker.go
tpantelis: Restructured it.

tpantelis (commented):
Addressed comments with https://github.com/submariner-io/submariner-operator/pull/1717/commits/54415d5770860cb66a5c2bfee57591786e5e282c.

internal/exit/exit.go
```diff
@@ -24,23 +24,23 @@ import (
 	"github.com/submariner-io/submariner-operator/pkg/version"
 )
 
-// OnError will print your error nicely and exit in case of error.
-func OnError(message string, err error) {
+// OnError exits in case of error.
+func OnError(err error) {
```
Jaanki, line 28: We do want to print what the err is

cmd/subctl/deploybroker.go
```diff
@@ -43,20 +43,20 @@ var deployBroker = &cobra.Command{
...
-		exit.OnError("Error creating client producer", err)
+		exit.OnError(status.Error(err, "Error creating client producer"))
 
-		status := cli.NewReporter()
 		err = deploy.Broker(&deployflags, clientProducer, status)
-		if err == nil {
-			err = broker.WriteInfoToFile(config, deployflags.BrokerNamespace, ipsecSubmFile,
-				stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
-		}
+		exit.OnError(err)
```
Jaanki, line 55: exit.OnError(status.Error(.....))?

Jaanki (changes requested):

internal/exit/exit.go
tpantelis: With the current usages the error is already printed - here we provide a convenience to exit if non-nil. If the caller wants to print the error they can call `WithMessage`.

cmd/subctl/deploybroker.go
tpantelis: `deploy.Broker` has already logged the error via the `status`. Here we just want to exit with a non-zero code if `err` is  non-nil.

Jaanki: On second thought, `status` should end in the `pkg` itself and not propagate to `cmd/subctl`. This way any user using `broker` pkg will have a complete status implementation.

Jaanki (changes requested):

Jaanki (approved):

PR #1739, opened 2022-01-05
Add manifests for 0.12.0

### LINKED ISSUE #1716: Broken bundle detected by CI
# Broken bundle

Periodic operator bundle CI detected a broken bundle. Please see the [periodic job
results](https://github.com/submariner-io/submariner-operator/actions?query=workflow%3APeriodic) for details.

Depends on https://github.com/submariner-io/enhancements/issues/81

skitt:
Fixes: #1716


skitt:
This gets us past the bundle validation errors seen in the periodic check, but the deployment itself fails.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,57 @@
+---
```
mkolesnik, line 1: Why are you adding this entire job matrix? It seems quite excessive and most of it (if not all) should be covered by the E2E

config/manifests/kustomization.yaml
```diff
@@ -9,6 +9,6 @@ resources:
 images:
 - name: controller
   newName: quay.io/submariner/submariner-operator
-  newTag: 0.11.0
+  newTag: devel-efc6ef9e
```
mkolesnik, line 12: Is it correct to reference devel?

skitt: No, and I think this is contributing to the failures of the bundle test jobs; the published bundles now store the hash of the operator repository, rather than the relevant tag.

.github/workflows/deploy.yml
skitt: It’s covered in Shipyard, but not here. There are _no_ tests of the bundle in the operator, which is unfortunate since that’s where it gets changed. Catching failures in Shipyard doesn’t prevent merging the PR causing them in the operator.

dfarrell07:
This depends on other bundle work, closing for now.

PR #1742, opened 2022-01-05
Sync the broker secret from the broker

skitt:
This ensures that the broker secret, used to access the broker, is
kept up-to-date with token updates and secret changes in the service
account.

### DIFF
--- a/controllers/submariner/submariner_controller.go
+++ b/controllers/submariner/submariner_controller.go
@@ -21,10 +21,17 @@ package submariner
 import (
 	"context"
 	"reflect"
+	"sync"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/federate"
+	level "github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/syncer"
+	"github.com/submariner-io/admiral/pkg/util"
 	submopv1a1 "github.com/submariner-io/submariner-operator/api/submariner/v1alpha1"
+	"github.com/submariner-io/submariner-operator/pkg/broker"
 	submarinerclientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
 	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
 	"github.com/submariner-io/submariner-operator/pkg/gateway"
@@ -35,6 +42,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


dfarrell07 (approved):

pkg/broker/rbac.go
```diff
@@ -101,6 +105,11 @@ func NewBrokerClusterRole() *rbacv1.Role {
 				APIGroups: []string{"discovery.k8s.io"},
 				Resources: []string{"endpointslices", "endpointslices/restricted"},
 			},
+			{
+				Verbs:     []string{"get", "list"},
```
tpantelis, line 109: ?

```suggestion
				Verbs:     []string{"get", "list", "watch"},
```

controllers/submariner/submariner_controller.go
```diff
@@ -116,6 +140,69 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+				})
+			if err != nil {
+				return reconcile.Result{}, errors.Wrap(err, "error building a resource syncer for secrets")
+			}
+			// TODO skitt handle stopping this when we delete the resource
+			secretSyncer.Start(ctx.Done())
+
+			r.secretSyncers[instance.Spec.BrokerK8sSecret] = secretSyncer
+		}
+	}
```
tpantelis, line 150: Can we move this into a separate function since this function is already large? (actually I'm surprised this didn't trip the gocyclo linter threshold)

controllers/submariner/submariner_controller.go
```diff
@@ -116,6 +140,69 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+				&syncer.ResourceSyncerConfig{
+					Name:            "Broker secret syncer",
+					ResourceType:    &corev1.Secret{},
+					SourceClient:    brokerClient,
+					SourceNamespace: instance.Spec.BrokerK8sRemoteNamespace,
+					Direction:       syncer.RemoteToLocal,
+					RestMapper:      r.config.Client.RESTMapper(),
+					Scheme:          r.config.Scheme,
+					Federator: federate.NewCreateOrUpdateFederator(
+						r.config.DynClient, r.config.Client.RESTMapper(), request.Namespace, instance.ClusterName),
```
tpantelis: Passing the cluster name will cause it to add the `submariner-io/clusterID` label which I don't believe you need/want?

```suggestion
						r.config.DynClient, r.config.Client.RESTMapper(), request.Namespace, ""),
```

controllers/submariner/submariner_controller.go
```diff
@@ -116,6 +140,69 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+				return reconcile.Result{}, errors.Wrap(err, "error building a dynamic client for the broker")
+			}
+
+			secretSyncer, err := syncer.NewResourceSyncer(
+				&syncer.ResourceSyncerConfig{
+					Name:            "Broker secret syncer",
+					ResourceType:    &corev1.Secret{},
+					SourceClient:    brokerClient,
+					SourceNamespace: instance.Spec.BrokerK8sRemoteNamespace,
+					Direction:       syncer.RemoteToLocal,
```
tpantelis:
```suggestion
					Direction:       syncer.None,
```

tpantelis (changes requested):

pkg/broker/rbac.go
skitt: I had that initially but the resulting RBAC fails, I’ll try again.

controllers/submariner/submariner_controller.go
skitt: Why don’t we want `RemoteToLocal`? I want the syncer to update the local copy if the remote changes.

pkg/broker/rbac.go
skitt:
```
[cluster1] Error creating SA for cluster: error binding sa to cluster role: rolebindings.rbac.authorization.k8s.io "cluster-cluster1-submariner-k8s-broker-cluster" is forbidden: user "system:serviceaccount:submariner-k8s-broker:submariner-k8s-broker-admin" (groups=["system:serviceaccounts" "system:serviceaccounts:submariner-k8s-broker" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
[cluster1] {APIGroups:[""], Resources:["secrets"], Verbs:["watch"]}
```

controllers/submariner/submariner_controller.go
tpantelis: It will.  This setting is really for 2-way syncing to avoid endless update loops by inspecting the cluster ID label. Eg, we don't want to sync local resources from the remote source nor sync remote resources back to the remote source.

pkg/broker/rbac.go
tpantelis: I thought we needed it b/c we're watching `Secrets`?

skitt: So did I, but it seems to work without it...

tpantelis (approved):

PR #1744, opened 2022-01-07
Determine the real "from version" for bundles

skitt:
Currently, the "from version" for bundles is determined by listing all
the tags in the repository, sorting them in reverse order, excluding
the current version, and picking the first tag left over. This works
on devel, but fails in any release branch since there will be tags
newer than the release tag.

Instead, list all the tags, sort them in reverse order, look for the
current tag, and output the tag after that, which will be the previous
tag in the current release branch.


tpantelis (approved):

sridhargaddam (approved):

SteveMattar:
good catch @skitt 😉

PR #1745, opened 2022-01-07
Avoid running tests on every PR edit

skitt:
Using the "edited" type for pull request filters causes the
corresponding jobs to run on every PR edit, which isn't desirable.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -3,7 +3,7 @@ name: Cross-Build Checks
 
 on:
   pull_request:
-    types: [ready_for_review, opened, edited, reopened, synchronize, converted_to_draft, labeled]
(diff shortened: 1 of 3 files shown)


skitt:
I noticed this on https://github.com/submariner-io/submariner-operator/pull/1744 — check the “upgrade” job there, it’s run three times already even though the PR hasn’t had any change which would warrant it.

tpantelis (approved):

Jaanki (approved):

mkolesnik:
Not sure what `edited` means, according to [this post](https://frontside.com/blog/2020-05-26-github-actions-pull_request/) it's "title, body, or the base branch of the PR is modified"

If so, for title and body it does seem redundant, but if the base branch changed and the event is 'edited' that might be a problem

PR #1770, opened 2022-01-12
Use reporter.Interface in cmd/cloud

Jaanki:


pkg/subctl/cmd/cloud/utils/reporter.go
```diff
@@ -19,36 +19,30 @@ limitations under the License.
...
 }
 
-func (r *cliReporter) Succeeded(message string, args ...interface{}) {
-	if message != "" {
-		r.status.QueueSuccessMessage(fmt.Sprintf(message, args...))
-	}
-
-	r.status.EndWith(cli.Success)
+func (s *statusReporter) Succeeded(message string, args ...interface{}) {
+	s.status.Success(message, args...)
```
skitt, line 40: Please don’t drop the check against `message`’s contents.

pkg/subctl/cmd/cloud/utils/reporter.go
```diff
@@ -19,36 +19,30 @@ limitations under the License.
...
 }
 
-func (r *cliReporter) Failed(err ...error) {
-	if len(err) > 0 {
-		r.status.QueueFailureMessage(err[0].Error())
-	}
-
-	r.status.EndWith(cli.Failure)
+func (s *statusReporter) Failed(err ...error) {
+	s.status.Failure(err[0].Error())
```
skitt: Same here, please don’t drop the check against `err`’s contents.

skitt (changes requested):

pkg/subctl/cmd/cloud/utils/reporter.go
Jaanki: `status.Success` checks for nil message https://github.com/submariner-io/submariner-operator/blob/devel/internal/cli/status.go#L227 so removed the check here.

pkg/subctl/cmd/cloud/utils/reporter.go
Jaanki: Ditto https://github.com/submariner-io/submariner-operator/blob/devel/internal/cli/status.go#L213

skitt: Not quite, if `err` is empty here, `err[0].Error()` will panic.

skitt (changes requested):

skitt (approved):

tpantelis (approved):

PR #1772, opened 2022-01-12
Refactor benchmark cmd

Jaanki:
`benchmark` package is under `/internal` at the moment because it replies on shipyard test Framework which uses a lot of globals and isn't concurrency-safe.

### DIFF
--- a/cmd/subctl/root.go
+++ b/cmd/subctl/root.go
@@ -19,11 +19,20 @@ limitations under the License.
 package subctl
 
 import (
+	"bytes"
+	"context"
 	"fmt"
 	"os"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+	"github.com/submariner-io/submariner-operator/internal/constants"
+	"github.com/submariner-io/submariner-operator/internal/exit"
 	"github.com/submariner-io/submariner-operator/internal/restconfig"
+	"github.com/submariner-io/submariner-operator/pkg/client"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var restConfigProducer = restconfig.NewProducer()
@@ -41,4 +50,37 @@ func Execute() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
+}
+
+func detectGlobalnet() {
+	clientProducer, err := client.NewProducerFromRestConfig(framework.RestConfigs[framework.ClusterA])
+	exit.OnErrorWithMessage(err, "Error creating client producer")
+
+	operatorClient := clientProducer.ForOperator()
+
+	submariner, err := operatorClient.SubmarinerV1alpha1().Submariners(constants.OperatorNamespace).Get(
+		context.TODO(), constants.SubmarinerName, v1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
+		exit.WithMessage("The Submariner resource was not found. Either submariner has not" +
+			"been deployed in this cluster or was deployed using helm. This command only supports submariner deployed" +
+			" using the operator via 'subctl join'.")
+	}
+
+	exit.OnErrorWithMessage(err, "Error obtaining Submariner resource")
+
+	framework.TestContext.GlobalnetEnabled = submariner.Spec.GlobalCIDR != ""
+}
+
+func compareFiles(file1, file2 string) (bool, error) {
+	first, err := os.ReadFile(file1)
+	if err != nil {
+		return false, errors.Wrapf(err, "error reading file %q", file1)
+	}
+
+	second, err := os.ReadFile(file2)
+	if err != nil {
+		return false, errors.Wrapf(err, "error reading file %q", file2)
+	}
+
+	return bytes.Equal(first, second), nil
 }
--- a/.github/workflows/e2e-new-subctl.yml
+++ b/.github/workflows/e2e-new-subctl.yml
@@ -46,3 +46,13 @@ jobs:
             --servicecidr 100.1.0.0/16 --clustercidr 10.1.0.0/16
           cmd/bin/subctl join --kubeconfig output/kubeconfigs/kind-config-cluster2 broker-info.subm --clusterid cluster2 --natt=false \
             --servicecidr 100.2.0.0/16 --clustercidr 10.2.0.0/16
+
+      - name: Test benchmark
+        run: |
+          cmd/bin/subctl benchmark latency --kubeconfig output/kubeconfigs/kind-config-cluster1:output/kubeconfigs/kind-config-cluster2 \
+          	--kubecontexts cluster1,cluster2 --verbose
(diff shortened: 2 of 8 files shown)


pkg/tests/framework.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	operationTimeout                = 240
+	connectionTimeout               = 60
+	connectionAttempts              = 2
+	reportDirectory                 = "."
+	verboseConnectivityVerification = false
+)
+
+func ConfigureFramework(args []string, restConfigProducer restconfig.Producer) error {
```
Jaanki: How about moving this to restconfig?

skitt (dismissed):

cmd/subctl/root.go
```diff
@@ -42,3 +52,41 @@ func Execute() {
...
+
+	clientProducer, err := client.NewProducerFromRestConfig(framework.RestConfigs[framework.ClusterA])
+	exit.OnError(status.Error(err, "Error creating client producer"))
+
+	operatorClient := clientProducer.ForOperator()
+
+	submariner, err := operatorClient.SubmarinerV1alpha1().Submariners(constants.OperatorNamespace).Get(
+		context.TODO(), constants.SubmarinerName, v1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
+		exit.OnError(status.Error(err, "The Submariner resource was not found. Either submariner has not"+
```
tpantelis: We don't really need to output the K8s error here.

```suggestion
		exit.WithMessage("The Submariner resource was not found. Either submariner has not"+
```

cmd/subctl/root.go
```diff
@@ -42,3 +52,41 @@ func Execute() {
...
+			"been deployed in this cluster or was deployed using helm. This command only supports submariner deployed"+
+			" using the operator via 'subctl join'."))
+	}
+
+	if err != nil {
+		exit.OnError(status.Error(err, "Error obtaining Submariner resource: %v"))
```
tpantelis:
```suggestion
		exit.OnError(status.Error(err, "Error obtaining Submariner resource"))
```

pkg/tests/framework.go
Jaanki: P.S. I just realised the comment was not pushed until now.

pkg/tests/framework.go
```diff
@@ -0,0 +1,86 @@
...
+	// Legacy handling: if arguments are provided and are files, assume they are kubeconfigs;
+	// otherwise, use contexts from --kubecontexts
+	// This is shared by verify and benchmark
+	if len(args) > 0 {
```
tpantelis: This is legacy behavior pertaining to `subctl` so shouldn't be in the library. We also shouldn't pass CLI args to the library. I think we should keep this under _cmd/subctl_ or perhaps we can remove support for this legacy behavior altogether.

pkg/tests/framework.go
tpantelis: The `restconfig.Producer` is really tied to the subctl CLI. Also it is in the _internal_ package so can't be referenced by users outside of the repo which means this function can only be called internally.  But it's only used to call `PopulateTestFramework` which the cmd front-end can do.

pkg/benchmark/throughput.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/onsi/gomega"
+	"github.com/submariner-io/shipyard/test/e2e/framework"
+	"github.com/submariner-io/submariner-operator/internal/restconfig"
+	"github.com/submariner-io/submariner-operator/pkg/tests"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var Verbose bool
```
tpantelis: This should be passed in instead of global.

pkg/benchmark/throughput.go
```diff
@@ -0,0 +1,191 @@
...
+		fmt.Printf("Performing throughput tests from Non-Gateway pod to Gateway pod on cluster %q\n", clusterAName)
+		runThroughputTest(f, testIntraClusterParams)
+	}
+
+	cleanupFramework(f)
+}
+
+var By = func(str string) {
+	if Verbose {
+		fmt.Println(str)
```
tpantelis: I think we should let the caller pass in an `io.Writer` and use `fmt.Fprintln` here instead of assuming `os.Stdout`.

tpantelis (changes requested):
I'm wondering if we should put the benchmark code under _internal_ instead of exposing it publicly under  _pkg_ at least for now. One issue is that is it uses the shipyard test Framework which uses a lot of globals and isn't concurrency-safe. To make it more generally reusable as a library, we should eliminate the globals. @skitt

pkg/benchmark/throughput.go
Jaanki: All `subctl` output is `os.Stdout`. We can add a flag to the whole `benchmark` cmd if we want to store the output someplace else.

pkg/benchmark/throughput.go
```diff
@@ -0,0 +1,187 @@
...
+	cleanupFramework(f)
+}
+
+var By = func(str string) {
+	fmt.Println(str)
+}
+
+func initFramework(baseName string) *framework.Framework {
+	f := framework.NewBareFramework(baseName)
+	framework.SetStatusFunction(By)
```
tpantelis: Instead of making `By` global and checking the `verbose` flag everywhere, you can define the function here and check the verbose flag.

```suggestion
	framework.SetStatusFunction(func(str string) {
	     if verbose {
	         fmt.Println(str)
	     }
         })
```

tpantelis (changes requested):

pkg/benchmark/throughput.go
tpantelis: I was referring to making the output destination configurable in the library by passing an `io.Writer`. But that would only redirect calls to `By` - we'd also have to configure the gomega fail handler. We can do that later if need be.

skitt: @Jaanki, since this is not under `cmd` we can’t consider it to be `subctl`-specific, so the same rules apply as elsewhere: no output unless to something controller-provided (a reporter or a writer), in fact no I/O in general unless to something controller-provided.

skitt:
> I'm wondering if we should put the benchmark code under _internal_ instead of exposing it publicly under _pkg_ at least for now. One issue is that is it uses the shipyard test Framework which uses a lot of globals and isn't concurrency-safe. To make it more generally reusable as a library, we should eliminate the globals. @skitt

Yes, that would make sense, with a note in the commit message or the code explaining why it’s `internal` instead of publicly-consumable.

cmd/subctl/root.go
```diff
@@ -42,3 +52,41 @@ func Execute() {
 		os.Exit(1)
 	}
 }
+
+func detectGlobalnet() {
+	status := cli.NewReporter()
+	status.Start("Trying to detect Globalnet settings")
```
tpantelis: I don't think we really need to use the status spinner here since we didn't before and we don't use one for the rest of the command output.

tpantelis (approved):

skitt (approved):
The message explaining `internal` needs to go in the commit message, not (just) the PR description; I’ll fix that.

PR #1775, opened 2022-01-13
Deprecate colour codes, stop sending them to Submariner

skitt:
These aren't used, let's drop them.

Depends on #1726.

### DIFF
--- a/api/submariner/v1alpha1/submariner_types.go
+++ b/api/submariner/v1alpha1/submariner_types.go
@@ -124,7 +124,6 @@ type (
 )
 
 const (
-	DefaultColorCode                     = "blue"
 	K8s                   KubernetesType = "k8s"
 	OCP                                  = "ocp"
 	EKS                                  = "eks"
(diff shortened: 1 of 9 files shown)


dfarrell07:
This might address part of a requirement for CNCF Incubation, btw.

> Specifications must have at least one public reference implementation.

Since color codes might could be considered a spec, and they aren't really implemented (or at least we don't know if they work).

https://github.com/submariner-io/enhancements/issues/58

dfarrell07 (approved):

dfarrell07:
> pkg/subctl/cmd/join.go:97:28: Error return value of `(*github.com/spf13/pflag.FlagSet).MarkDeprecated` is not checked (errcheck)
	cmd.Flags().MarkDeprecated("colorcodes", "--colorcodes has no effect and is deprecated")

tpantelis:
https://github.com/submariner-io/submariner-operator/pull/1726 is also in flight.

tpantelis (approved):

PR #1791, opened 2022-01-18
Add E2E GHA to test all K8s versions

dfarrell07:
Add GHA to run the E2E tests for the normal matrix but with every
relevant K8s version. This includes the currently supported versions, as
well as the oldest version thought to work with SubM.

Triggered once a week or manually with the "e2e-all-k8s" label.

### DIFF
--- a/.github/workflows/e2e-all-k8s.yml
+++ b/.github/workflows/e2e-all-k8s.yml
@@ -1,0 +1,40 @@
+---
+name: End to End All K8s
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened]
+  schedule:
+    - cron: "0 0 * * 6"
+
+jobs:
+  e2e:
+    name: E2E
+    if: contains(github.event.pull_request.labels.*.name, 'e2e-all-k8s')
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        globalnet: ['', 'globalnet']
+        k8s_version: ['1.17', '1.20', '1.21', '1.22', '1.23']
+        lighthouse: ['', 'lighthouse']
+        ovn: ['', 'ovn']
+        exclude:
+          - ovn: 'ovn'
+            lighthouse: 'lighthouse'
+          - ovn: 'ovn'
+            globalnet: 'globalnet'
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+
+      - name: Run E2E deployment and tests
+        uses: submariner-io/shipyard/gh-actions/e2e@devel
+        with:
+          k8s_version: ${{ matrix.k8s_version }}
+          using: ${{ matrix.globalnet }} ${{ matrix.lighthouse }} ${{ matrix.ovn }}
+
+      - name: Post mortem
+        if: failure()
+        uses: submariner-io/shipyard/gh-actions/post-mortem@devel


.github/workflows/e2e-all-k8s.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: End to End All K8s
+
+on:
```
Jaanki, line 4: I think this should be general job and not a PR job which runs weekly/twice a week on the devel branch.

.github/workflows/e2e-all-k8s.yml
```diff
@@ -0,0 +1,40 @@
...
+  e2e:
+    name: E2E
+    if: contains(github.event.pull_request.labels.*.name, 'e2e-all-k8s')
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        globalnet: ['', 'globalnet']
+        k8s_version: ['1.17', '1.20', '1.21', '1.22', '1.23']
```
Jaanki, line 20: Should we drop 1.17?

.github/workflows/e2e-all-k8s.yml
dfarrell07: Do you mean this should run on every PR? There is other CI for that, see: https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/e2e-full.yml#L33

This is one of the jobs discussed as a part of https://github.com/submariner-io/submariner-website/issues/624, for the use-case of wanting to run the full E2E matrix against every K8s version when developing something that is expected to have an impact, periodically, and on releases.

.github/workflows/e2e-all-k8s.yml
dfarrell07: No, 1.17 is currently the "last thought to be working version", which we want to keep testing to maintain https://github.com/submariner-io/submariner-operator/blob/devel/pkg/version/version.go#L34. See the proposed docs about all of this: https://github.com/submariner-io/submariner-website/pull/643

.github/workflows/e2e-all-k8s.yml
Jaanki: I meant to just keep the `schedule` section and remove `pull_request` section but then saw your PR in `release` repo https://github.com/submariner-io/releases/pull/298 and understood the intent.

Jaanki (approved):
Don't forget to create the label `e2e-all-k8s` on relevant repos.

Jaanki:
I would also add the `e2e-all-k8s` label to this PR to test the GHA once the PR has enough votes and doesn't require a rework.

.github/workflows/e2e-all-k8s.yml
mkolesnik: This isn't aligned with the docs though

> Submariner's policy is to support all versions upstream-Kubernetes supports and no EOL versions.

Also didn't we agree not to to officially test these old versions, so that we don't give a false semblance of supporting them?

skitt (approved):

.github/workflows/e2e-all-k8s.yml
Jaanki: > This isn't aligned with the docs though
> 
> > Submariner's policy is to support all versions upstream-Kubernetes supports and no EOL versions.
> 
> Also didn't we agree not to to officially test these old versions, so that we don't give a false semblance of supporting them?

I thought so too There was an issue as well (can't find it right now though) to track this ans was discussed in one of the meetings.

skitt: It is aligned with what was decided in the weekly call; see the minutes: “Keep a test in the operator to make sure that our “breaking” threshold on installation (currently 1.17) is accurate”.

skitt: IIRC from the discussions, we wanted to distinguish the minimum version we *support*, and the version below which we prevent installation. There’s a difference between “we’ll let you install this, but we won’t provide support” (for some value of “support” anyway in an open source project), and “we know this will break, so we’ll prevent you from doing it” (unless you provide the appropriate flag).

dfarrell07: @mkolesnik I'd like to go ahead and merge this, so we don't have to rebase and run all the jobs again. Does Stephen's answer clarify things?

dfarrell07:
I see Tom has rebased this, and I'm fairly sure we're doing the right thing here as Stephen explained, so I'd like to go ahead and merge this to prevent running 52 jobs yet again.

Update: If we have more feedback to address, we can do it in a follow-up PR without all the testing labels.

PR #1795, opened 2022-01-20
Move verify to cmd/subctl

Jaanki:

### DIFF
--- a/cmd/subctl/verify.go
+++ b/cmd/subctl/verify.go
@@ -16,11 +16,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cmd
+package subctl
 
 // nolint:revive // Blank imports below are intentional.
 import (
-	"context"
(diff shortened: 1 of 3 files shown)


cmd/subctl/verify.go
```diff
@@ -0,0 +1,286 @@
...
+limitations under the License.
+*/
+
+package subctl
+
+// nolint:revive // Blank imports below are intentional.
+import (
+	"errors"
+	"fmt"
+	"github.com/submariner-io/submariner-operator/internal/exit"
```
skitt: This needs to be moved into the second set of imports.

skitt (changes requested):

cmd/subctl/verify.go
Jaanki: Done

skitt (approved):

tpantelis (approved):

PR #1800, opened 2022-01-24
Replace utils.ExitOnError->exit.OnErrorWithMessage

Jaanki:

### DIFF
--- a/internal/restconfig/restconfig.go
+++ b/internal/restconfig/restconfig.go
@@ -29,6 +29,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/stringset"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
 	"github.com/submariner-io/submariner-operator/api/submariner/v1alpha1"
+	"github.com/submariner-io/submariner-operator/internal/exit"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd/utils"
 	"github.com/submariner-io/submariner-operator/pkg/version"
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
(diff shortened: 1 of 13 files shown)


internal/exit/exit.go
```diff
@@ -40,7 +40,7 @@ func WithMessage(message string) {
 }
 
 // OnErrorWithMessage will print the message and quit the program with an error code.
-func OnErrorWithMessage(err error, message string) {
+func OnErrorWithMessage(message string, err error) {
```
tpantelis: Why was the signature changed? I prefer the previous ordering which is consistent with `errors.Wrap`, `errors.WithMessage` et al.

internal/restconfig/restconfig.go
```diff
@@ -20,6 +20,8 @@ package restconfig
 
 import (
 	"fmt"
+	"github.com/submariner-io/submariner-operator/internal/exit"
+	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd/utils"
```
tpantelis: `utils` is no longer referenced.

tpantelis (changes requested):
We should be able to remove `utils.ExitOnError`.

internal/exit/exit.go
Jaanki: It was changed to match the output message `custom message: err`.

internal/restconfig/restconfig.go
Jaanki: L302 still uses `utils.GetSubmarinerResource(config)`

internal/exit/exit.go
tpantelis: But the runtime output produced isn't related to the param ordering. The param ordering should reflect the function name. Please keep the ordering as it was for consistency.

tpantelis (approved):

skitt (approved):
This is much clearer!

sridhargaddam (approved):

PR #1801, opened 2022-01-25
Use reporter interface & clientproducer for gather

Jaanki:

### DIFF
--- a/pkg/subctl/cmd/gather/gather.go
+++ b/pkg/subctl/cmd/gather/gather.go
@@ -31,10 +31,9 @@ import (
 	"github.com/submariner-io/submariner-operator/internal/exit"
 	"github.com/submariner-io/submariner-operator/internal/restconfig"
 	"github.com/submariner-io/submariner-operator/pkg/brokercr"
-	subOperatorClientset "github.com/submariner-io/submariner-operator/pkg/client/clientset/versioned"
+	"github.com/submariner-io/submariner-operator/pkg/client"
 	"github.com/submariner-io/submariner-operator/pkg/names"
 	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd"
-	"github.com/submariner-io/submariner-operator/pkg/subctl/operator/submarinercr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -126,6 +125,14 @@ func gatherData(cluster *cmd.Cluster) bool {
 func gatherDataByCluster(cluster *cmd.Cluster, directory string) {
 	var err error
 	clusterName := cluster.Name
+	status := cli.NewReporter()
+
+	clientProducer, err := client.NewProducerFromRestConfig(cluster.Config)
+	if err != nil {
+		status.Failure("Error creating client producer")
+
+		return
+	}
 
 	fmt.Printf("Gathering information from cluster %q\n", clusterName)
 
@@ -135,35 +142,17 @@ func gatherDataByCluster(cluster *cmd.Cluster, directory string) {
 		DirName:              directory,
 		IncludeSensitiveData: includeSensitiveData,
 		Summary:              &Summary{},
+		ClientProducer:       clientProducer,
+		Submariner:           cluster.Submariner,
 	}
 
-	info.DynClient = cluster.DynClient
-	info.ClientSet = cluster.KubeClient
-
-	submarinerClient, err := subOperatorClientset.NewForConfig(cluster.Config)
-	if err != nil {
-		fmt.Printf("Error getting Submariner client: %s\n", err)
-		return
-	}
-
-	info.Submariner, err = submarinerClient.SubmarinerV1alpha1().Submariners(cmd.OperatorNamespace).
-		Get(context.TODO(), submarinercr.SubmarinerName, metav1.GetOptions{})
-	if err != nil {
-		info.Submariner = nil
-
-		if !apierrors.IsNotFound(err) {
-			fmt.Printf("Error getting Submariner resource: %s\n", err)
-			return
-		}
-	}
-
-	info.ServiceDiscovery, err = submarinerClient.SubmarinerV1alpha1().ServiceDiscoveries(cmd.OperatorNamespace).
+	info.ServiceDiscovery, err = clientProducer.ForOperator().SubmarinerV1alpha1().ServiceDiscoveries(cmd.OperatorNamespace).
 		Get(context.TODO(), names.ServiceDiscoveryCrName, metav1.GetOptions{})
 	if err != nil {
 		info.ServiceDiscovery = nil
 
 		if !apierrors.IsNotFound(err) {
-			fmt.Printf("Error getting ServiceDiscovery resource: %s\n", err)
+			status.Failure("Error getting ServiceDiscovery resource: %s", err)
 			return
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/subctl/cmd/gather/gather.go
```diff
@@ -116,54 +116,43 @@ func gatherData(cluster *cmd.Cluster) bool {
...
+func gatherDataByCluster(cluster *cmd.Cluster, directory string, status reporter.Interface) {
 	var err error
 	clusterName := cluster.Name
 
-	fmt.Printf("Gathering information from cluster %q\n", clusterName)
+	clientProducer, err := client.NewProducerFromRestConfig(cluster.Config)
+	if err != nil {
+		status.Failure("Error creating client producer")
+		status.End()
+		return
```
tpantelis:
```suggestion
		return status.Error(err, "Error creating client producer")
```

pkg/subctl/cmd/gather/gather.go
```diff
@@ -65,7 +65,7 @@ var gatherTypeFlags = map[string]bool{
 	"resources": false,
 }
 
-var gatherFuncs = map[string]func(string, Info) bool{
+var gatherFuncs = map[string]func(string, Info) {
```
tpantelis: Why was the return value removed? The reason for it is to avoid outputting any status if the module/data type doesn't apply.

tpantelis (changes requested):

pkg/subctl/cmd/gather/gather.go
skitt: The logging as it stands isn’t balanced: we always call `info.Status.Start()`, but the end function only gets called if the function returns true. Shouldn’t the end function be called in all cases? Then the return value is no longer needed, and the status is only reflected in the calls to the status functions...

skitt (changes requested):

pkg/subctl/cmd/gather/gather.go
```diff
@@ -175,9 +164,8 @@ func gatherDataByCluster(cluster *cmd.Cluster, directory string) {
 					info.Status = cli.NewStatus()
 					info.Status.Start(fmt.Sprintf("Gathering %s %s", module, dataType))
 
-					if gatherFuncs[module](dataType, info) {
-						info.Status.EndWith(info.Status.ResultFromMessages())
-					}
+					gatherFuncs[module](dataType, info)
+					status.End()
```
skitt: This should be `info.Status.End()`.

skitt (changes requested):

pkg/subctl/cmd/gather/gather.go
tpantelis: The idea was to avoid outputting anything if the module/data type doesn't apply (like for `gatherBroker`, _Logs_ doesn't apply), taking advantage of the fact that the CLI status doesn't output anything unless `End` is called.  Either way though I think changing this function signature is really outside the scope of this PR.

Jaanki: Changing the signature, I think is a side effect of balancing the status start and end.

tpantelis: Then how do we handle the case where the module/data type doesn't apply? Either way this doesn't pertain to the stated purpose of this PR.

pkg/subctl/cmd/gather/gather.go
Jaanki: Using `sttaus.Error` means I will have to change `gatherDataByCluster` to return error and check the error at the caller.

```
gatherDataByCluster(cluster, directory, status)

    status.Success("Files are stored under directory %q\n", directory)
    status.End()
```

Checking for error here is not logical because there is no further operation happening based on the error. More logical approach is to log the error via `status.Failure` and move to `status.Success("Files are stored under directory %q\n", directory)`

pkg/subctl/cmd/gather/gather.go
Jaanki: well, I have reverted the change for now. Lets discuss this offline.

pkg/subctl/cmd/gather/gather.go
```diff
@@ -116,54 +116,46 @@ func gatherData(cluster *cmd.Cluster) bool {
...
 	}
 
-	gatherDataByCluster(cluster, directory)
+	status := cli.NewReporter()
+	status.Start("Gathering information from cluster %q\n", cluster.Name)
+	gatherDataByCluster(cluster, directory, status)
 
-	fmt.Printf("Files are stored under directory %q\n", directory)
+	status.Success("Files are stored under directory %q\n", directory)
+	status.End()
```
tpantelis: I don't think any of this needs to change. L123 is an informational message and does not need to be deemed as Success so leave it as `fmt.Printf`.  Same with L120 - since we do Start/End for each module/type below we can keep this as informational (ie `fmt.Printf`) and keep it in `gatherDataByCluster`. There is no need to create a status here and pass into `gatherDataByCluster`.

pkg/subctl/cmd/gather/gather.go
```diff
@@ -116,54 +116,46 @@ func gatherData(cluster *cmd.Cluster) bool {
...
-func gatherDataByCluster(cluster *cmd.Cluster, directory string) {
+func gatherDataByCluster(cluster *cmd.Cluster, directory string, status reporter.Interface) {
 	var err error
 	clusterName := cluster.Name
 
-	fmt.Printf("Gathering information from cluster %q\n", clusterName)
+	clientProducer, err := client.NewProducerFromRestConfig(cluster.Config)
+	if err != nil {
+		status.Failure("Error creating client producer")
+		status.End()
```
tpantelis: We want to log the error here. The previous code used `fmt.Printf` but we  can use a status instance to get the red X decoration. 

```suggestion
	        status.Failure("Error creating client producer: %v", err)
```

pkg/subctl/cmd/gather/gather.go
```diff
@@ -172,11 +164,10 @@ func gatherDataByCluster(cluster *cmd.Cluster, directory string) {
 		if ok {
 			for dataType, ok := range gatherTypeFlags {
 				if ok {
-					info.Status = cli.NewStatus()
-					info.Status.Start(fmt.Sprintf("Gathering %s %s", module, dataType))
+					info.Status.Start("Gathering %s %s", module, dataType)
```
tpantelis, line 165: Since we may not `End` the status below, we need to create a new instance as before. Whether this code should be refactored to balance the Start/End I think is out-of-scope for this PR.

tpantelis (changes requested):

pkg/subctl/cmd/gather/gather.go
```diff
@@ -126,44 +125,32 @@ func gatherData(cluster *cmd.Cluster) bool {
...
-
-	info.ServiceDiscovery, err = submarinerClient.SubmarinerV1alpha1().ServiceDiscoveries(cmd.OperatorNamespace).
+	info.ServiceDiscovery, err = clientProducer.ForOperator().SubmarinerV1alpha1().ServiceDiscoveries(cmd.OperatorNamespace).
 		Get(context.TODO(), names.ServiceDiscoveryCrName, metav1.GetOptions{})
 	if err != nil {
 		info.ServiceDiscovery = nil
 
 		if !apierrors.IsNotFound(err) {
-			fmt.Printf("Error getting ServiceDiscovery resource: %s\n", err)
+			info.Status.Failure("Error getting ServiceDiscovery resource: %s", err)
```
tpantelis: We haven't created `info.Status` yet.

```suggestion
			status.Failure("Error getting ServiceDiscovery resource: %s", err)
```

tpantelis (approved):

pkg/subctl/cmd/gather/gather.go
```diff
@@ -246,33 +233,27 @@ func gatherBroker(dataType string, info Info) bool {
...
 
 		if brokerRestConfig != nil {
 			info.RestConfig = brokerRestConfig
 
-			info.DynClient, info.ClientSet, err = restconfig.Clients(brokerRestConfig)
-			if err != nil {
-				info.Status.QueueFailureMessage(fmt.Sprintf("Error getting the broker client: %s", err))
-				return true
-			}
-		} else {
```
skitt, line 261: I think that dropping this `else` is what’s causing your problems retrieving `Broker` objects.

Jaanki: yes, I missed that the clients are created by using different restconfigs in the if-else loop.

tpantelis (approved):

skitt (approved):

PR #1802, opened 2022-01-25
Netwok-discovery: handle networkType==Calico in OCP

### LINKED ISSUE #1852: submariner should accept "Calico" (with capital C) as network plugin
My OpenShift cluster is configured to use Calico as CNI network plugin instead of the default OpenShiftSDN:
```
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  name: cluster
spec:
  clusterNetwork:
    - cidr: 172.17.0.0/18
      hostPrefix: 23
  externalIP:
    policy: {}
  networkType: Calico
  serviceNetwork:
    - 172.21.0.0/16
```
unfortunately, submariner does not recognize "Calico" as a supported CNI network plugin because of the capital C:
```
# subctl diagnose all
...
 • Checking Submariner support for the CNI network plugin  ...
 ✗ Checking Submariner support for the CNI network plugin
 ✗ The detected CNI network plugin ("Calico") is not supported by Submariner. Supported network plugins: [generic canal-flannel weave-net OpenShiftSDN OVNKubernetes calico]
```

**What happened**:
`submariner-route-agent` does not create the required iptables rules because it does not recognize the "Calico" network plugin.

submariner should support "Calico" as this is the suggested name to use in the calico documentation:
https://projectcalico.docs.tigera.io/getting-started/openshift/installation

> $ sed -i 's/OpenShiftSDN/Calico/' install-config.yaml

Create an OpenShift cluster using "Calico" as network plugin.

**Environment**:

yboaron:
When calico is used as third-party plugin in OCP, networkType value (of Network.config.openshift.io/v1 CR named cluster) is set to 'Calico'.
With this PR, network-discovery sets NetworkPlugin to constants.NetworkPluginCalico (=='calico') when calico is used as plugin in OCP.

Signed-off-by: yboaron <yboaron@redhat.com>


yboaron:
This PR is instead https://github.com/submariner-io/submariner/pull/1666

pkg/discovery/network/openshift4.go
```diff
@@ -90,13 +95,17 @@ func parseOS4Network(cr *unstructured.Unstructured) (*ClusterNetwork, error) {
...
-
+	ocpNetworkType, found, err := unstructured.NestedString(cr.Object, "spec", "networkType")
 	if err != nil {
 		return nil, errors.Wrap(err, "error retrieving spec.networkType field")
 	} else if !found {
 		return nil, fmt.Errorf("field .spec.networkType expected, but not found in Network resource: %v", cr.Object)
 	}
-
+	// calico plugin named 'Calico' in networkType
+	if ocpNetworkType == NetworkTypeCalico {
```
sridhargaddam: What if some users configure "calico" instead of "Calico"? Can we do case insensitive comparision here for Calico and configure the `result.NetworkPlugin = constants.NetworkPluginCalico`

yboaron: If some user will configure "calico" I think we should be OK (we'll fail to the else statement).

This extra code should map OCP NetworkType to the correct NetworkPlugin,
so maybe we should at least raise an error if network-discovery for OCP sets result.NetworkPlugin to value != one of the supported plugins, what do you think?

sridhargaddam: > If some user will configure "calico" I think we should be OK (we'll fail to the else statement).

Oh yes, I missed it.

> This extra code should map OCP NetworkType to the correct NetworkPlugin, so maybe we should at least raise an error if network-discovery for OCP sets result.NetworkPlugin to value != one of the supported plugins, what do you think?

Such validations will be done using `subctl diagnose ...` tool

sridhargaddam:
Thanks for the PR @yboaron. There seems to be some linting errors, please take a look.

pkg/discovery/network/openshift4.go
```diff
@@ -90,13 +95,18 @@ func parseOS4Network(cr *unstructured.Unstructured) (*ClusterNetwork, error) {
...
 
-	result.NetworkPlugin, found, err = unstructured.NestedString(cr.Object, "spec", "networkType")
-
+	ocpNetworkType, found, err := unstructured.NestedString(cr.Object, "spec", "networkType")
 	if err != nil {
 		return nil, errors.Wrap(err, "error retrieving spec.networkType field")
 	} else if !found {
 		return nil, fmt.Errorf("field .spec.networkType expected, but not found in Network resource: %v", cr.Object)
 	}
+	// calico plugin named 'Calico' in networkType
```
dfarrell07: I think we probably shouldn't link comments to values of variables like this, to me it seems more likely to cause maintenance issues than to be helpful.

yboaron: :+1: 
Comment has been deleted.

pkg/discovery/network/openshift4.go
```diff
@@ -23,13 +23,18 @@ import (
...
+	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 )
 
+const (
+	NetworkTypeCalico = "Calico"
```
Jaanki:
```suggestion
	CalicoNetwork = "Calico"
```
seems a better constant name

pkg/discovery/network/openshift4.go
```diff
@@ -90,13 +95,18 @@ func parseOS4Network(cr *unstructured.Unstructured) (*ClusterNetwork, error) {
...
-
+	ocpNetworkType, found, err := unstructured.NestedString(cr.Object, "spec", "networkType")
 	if err != nil {
 		return nil, errors.Wrap(err, "error retrieving spec.networkType field")
 	} else if !found {
 		return nil, fmt.Errorf("field .spec.networkType expected, but not found in Network resource: %v", cr.Object)
 	}
 
+	if ocpNetworkType == NetworkTypeCalico {
+		result.NetworkPlugin = constants.NetworkPluginCalico
```
Jaanki, line 102: Instead of importing this constant from `submariner` repo, we can create a new constant `CalicoNetworkPlugin` in `internal/constant/constants.go`

yboaron: We use the submariner-imported network plugins constants for all plugins discovery code in the operator, I adopted the same approach also for CalicoNetworkPlugin.

If we want to maintain a separate plugins constant list in the operator's repo, it makes sense to do that for all plugins and not just CalicoNetworkPlugin.

IMO, we should maintain the constants in one place in order to avoid confusion and inconsistency. So, the current implementation seems right to me.

Please, let me know if you still think the code needs to be changed

Jaanki: Lets keep it as is then. Thanks for the explanation.

pkg/discovery/network/openshift4.go
yboaron: I'm totally open to changing the constant name, but since
this constant represents the value of 'networkType' field of OCP Network.config CR in case CNI is Calico, I think it makes sense to include NetworkType in the constant name somehow.  

Maybe 'CalicoNetworkType' ?

skitt: Is it even worth having a constant for this, let alone an exported constant? It’s only used once, and I don’t think it’s going to be used anywhere else, is it?

yboaron: Yep, 'NetworkTypeCalico' should be used only here and I don't expect it to be exported.
I think we can delete the constant definition, if in the future we'll need to support  more  networkType 2 Plugin mapping,we might consider using non exported constant values.

Jaanki (approved):

dfarrell07 (approved):

skitt (approved):

Jaanki (approved):

PR #1806, opened 2022-01-28
Make `ForCluster` return cluster config and name

Jaanki:
This is needed:

1. because new [`cluster.NewInfo`](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/cluster/info.go#L39) takes in a `clusterName`. For commands working on a single cluster and needing to create new cluster, `clusterName` is needed
2. uninstall cmd also needs a clustername
3. To make it consistent with [`ForClusters()`](https://github.com/submariner-io/submariner-operator/blob/devel/internal/restconfig/restconfig.go#L122)

### DIFF
--- a/cmd/subctl/deploybroker.go
+++ b/cmd/subctl/deploybroker.go
@@ -49,13 +49,13 @@ var deployBroker = &cobra.Command{
 		config, err := restConfigProducer.ForCluster()
 		exit.OnError(status.Error(err, "Error creating REST config"))
 
-		clientProducer, err := client.NewProducerFromRestConfig(config)
+		clientProducer, err := client.NewProducerFromRestConfig(config.Config)
 		exit.OnError(status.Error(err, "Error creating client producer"))
 
 		err = deploy.Broker(&deployflags, clientProducer, status)
 		exit.OnError(err)
 
-		err = broker.WriteInfoToFile(config, deployflags.BrokerNamespace, ipsecSubmFile,
+		err = broker.WriteInfoToFile(config.Config, deployflags.BrokerNamespace, ipsecSubmFile,
 			stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
 		exit.OnError(err)
 	},
--- a/cmd/subctl/export.go
+++ b/cmd/subctl/export.go
@@ -58,7 +58,7 @@ var (
 				exit.OnErrorWithMessage(err, "Error getting service Namespace")
 			}
 
-			clientProducer, err := client.NewProducerFromRestConfig(config)
+			clientProducer, err := client.NewProducerFromRestConfig(config.Config)
 			exit.OnError(status.Error(err, "Error creating client producer"))
 
 			err = service.Export(clientProducer, serviceNamespace, args[0], status)
--- a/cmd/subctl/join.go
+++ b/cmd/subctl/join.go
@@ -63,7 +63,7 @@ var joinCmd = &cobra.Command{
 		clientConfig, err := restConfigProducer.ForCluster()
 		exit.OnError(status.Error(err, "Error creating the REST config"))
 
-		clientProducer, err := client.NewProducerFromRestConfig(clientConfig)
+		clientProducer, err := client.NewProducerFromRestConfig(clientConfig.Config)
 		exit.OnError(status.Error(err, "Error creating the client producer"))
 
 		networkDetails := getNetworkDetails(clientProducer, status)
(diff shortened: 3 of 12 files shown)


tpantelis:
Can you explain in the PR description why this change in needed?

tpantelis:
> 1. because new [`cluster.NewInfo`](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/cluster/info.go#L39) takes in a `clusterName`. For commands working on a single cluster and needing to create new cluster, `clusterName` is needed

This PR significantly changes the implementation of `ForCluster`.  It's unclear to me if the new implementation is equivalent to the previous as I'm not all that familiar with the code (I'll defer to @skitt). But the `restconfig` code is internal to `subctl` and no command prior to the restructure needed this change so why would we need it now?  The _Cluster_  concept is coupled with `ExecuteMultiCluster`  which naturally utilizes `ForClusters`. The cluster name is only used for informational output and is really only needed in a multi-cluster context.

Jaanki:
> > 1. because new [`cluster.NewInfo`](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/cluster/info.go#L39) takes in a `clusterName`. For commands working on a single cluster and needing to create new cluster, `clusterName` is needed
> 
> This PR significantly changes the implementation of `ForCluster`. It's unclear to me if the new implementation is equivalent to the previous as I'm not all that familiar with the code (I'll defer to @skitt). But the `restconfig` code is internal to `subctl` and no command prior to the restructure needed this change so why would we need it now? The _Cluster_ concept is coupled with `ExecuteMultiCluster` which naturally utilizes `ForClusters`. The cluster name is only used for informational output and is really only needed in a multi-cluster context.

`diagnose firewall tunnel` command utilises `ForCluster`. (see https://github.com/submariner-io/submariner-operator/pull/1790/commits/6c45128cc7cc074c3d7fd73aeffcac3487699bd5#diff-0af38e4d61ea571330e9ed8c153f21fee397ae876a37a5cfb1010d4282084b57R177) It also needs `cluster` to be created from `cluster.NewInfo`. `cluster.NewInfo` takes in a `clusterName`. This `clusterName` with this change will come from `ForCluster` like it does with `ForClusters`. Currently, `diagnose firewall tunnel` is getting `clusterName` from [submariner.spec.clusterID](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/subctl/cmd/diagnose/firewall_tunnel.go#L86). But this is a more logical/consistent and global approach.

skitt:
> > 1. because new [`cluster.NewInfo`](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/cluster/info.go#L39) takes in a `clusterName`. For commands working on a single cluster and needing to create new cluster, `clusterName` is needed
> 
> This PR significantly changes the implementation of `ForCluster`. It's unclear to me if the new implementation is equivalent to the previous as I'm not all that familiar with the code (I'll defer to @skitt). But the `restconfig` code is internal to `subctl` and no command prior to the restructure needed this change so why would we need it now? The _Cluster_ concept is coupled with `ExecuteMultiCluster` which naturally utilizes `ForClusters`. The cluster name is only used for informational output and is really only needed in a multi-cluster context.

The new implementation is equivalent to the previous, and as a nice side-effect it adds support for in-cluster configuration (which is useful for `subctl diagnose` for ACM, and also for end-users running `subctl` in pods — which would apparently be useful to some).

tpantelis:
> diagnose firewall tunnel command utilises ForCluster. (see https://github.com/submariner-io/submariner-operator/commit/6c45128cc7cc074c3d7fd73aeffcac3487699bd5#diff-0af38e4d61ea571330e9ed8c153f21fee397ae876a37a5cfb1010d4282084b57R177) It also needs cluster to be created from > cluster.NewInfo. cluster.NewInfo takes in a clusterName. This clusterName with this change will come from ForCluster like > it does with ForClusters. Currently, diagnose firewall tunnel is getting clusterName from [submariner.spec.clusterID](https://github.com/submariner-io/submariner-operator/blob/devel/pkg/subctl/cmd/diagnose/firewall_tunnel.go#L86). But 
 this is a more logical/consistent and global approach.

As @skitt pointed out, this adds support for in-cluster configuration for single cluster commands which is useful. The only thing is that the cluster name in that case would be "_in-cluster_" so perhaps `diagnose firewall tunnel ` et al would still want to use `submariner.spec.clusterID` anyway.

skitt:
So this doesn’t address the initial problem of giving a name to the firewall diagnostics, but it does seem useful for the in-cluster scenarios. Let’s keep this open and revisit it for 0.13.

skitt (approved):

tpantelis (approved):

PR #1807, opened 2022-01-29
Refactor/simplify globalnet APIs

tpantelis:
...to make them easier to consume for third party apps. See individual commits for details.


pkg/join/join.go
```diff
@@ -68,17 +68,15 @@ func ClusterToBroker(brokerInfo *broker.Info, options *Options, clientProducer c
...
-	if options.GlobalnetEnabled {
-		status.Start("Discovering multi cluster details")
+	status.End()
 
-		err = globalnet.AllocateAndUpdateGlobalCIDRConfigMap(options.ClusterID, brokerAdminClientset, brokerNamespace, &netconfig)
+	if options.GlobalnetEnabled {
+		err = globalnet.AllocateAndUpdateGlobalCIDRConfigMap(brokerAdminClientset, brokerNamespace, &netconfig, status)
 		if err != nil {
-			return status.Error(err, "Error Discovering multi cluster details")
+			return err // nolint:wrapcheck // No need to wrap here
```
Jaanki: I think a high-level error message is good to have here. and then `globalnet.go` has detailed error specified.

pkg/reporter/klog.go
```diff
@@ -0,0 +1,50 @@
+/*
```
Jaanki: commit https://github.com/submariner-io/submariner-operator/pull/1807/commits/038af7cb542948d281f5323fc4654ff26aa002ff seems unrelated to the globalnet change and should be part of a separate PR.

Jaanki (changes requested):

pkg/reporter/klog.go
tpantelis: yeah I included these so the add-on has some choice of implementations to use instead of spinning it's own when the globanet APIs are used. But I'll move to another PR.

pkg/join/join.go
tpantelis: I'm not sure I follow. Do you mean wrap the return error or report it to the status? For the latter, `globalnet.AllocateAndUpdateGlobalCIDRConfigMap` already does that.

tpantelis: I assume you meant the former - it's now wrapped.

dfarrell07 (approved):

pkg/join/join.go
Jaanki: I meant to keep the writeup `"Error Discovering multi cluster details"` and not delete it.

tpantelis: Frankly, I don't think that message really makes sense in this context. I changed it to "_unable to determine the global CIDR_" which is what this code actually does.

sridhargaddam (approved):
Thanks @tpantelis

Jaanki (approved):

PR #1812, opened 2022-01-31
Add image and rbac for Diagnose

vthapar:
* Add Dockerfile to create `submariner-diagnose` image
* Add RBACs for `submariner-diagnose`

Refer: https://github.com/submariner-io/enhancements/pull/60


Makefile
```diff
@@ -161,6 +161,8 @@ package/Dockerfile.submariner-operator: bin/submariner-operator
 
 package/Dockerfile.submariner-operator-index: packagemanifests
 
+package/Dockerfile.submariner-debug: bin/subctl
```
skitt:
```suggestion
package/Dockerfile.submariner-diagnose: bin/subctl
```

config/rbac/submariner-diagnose/cluster_role.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - update
+      - delete
+      - watch
+  - apiGroups:  # pods, services and nodes are looked up to figure out network settings
+      - ""
+    resources:
+      - pods
+      - services
+      - nodes
+    verbs
```
skitt:
```suggestion
    verbs:
```

skitt (changes requested):

Makefile
```diff
@@ -161,6 +161,8 @@ package/Dockerfile.submariner-operator: bin/submariner-operator
 
 package/Dockerfile.submariner-operator-index: packagemanifests
 
+package/Dockerfile.submariner-diagnose: bin/subctl
+
```
skitt: This shouldn’t be necessary, `docker_deps` in Shipyard’s `Makefile.images` extracts `COPY` statements to calculate prerequisites.

skitt (changes requested):

Makefile
vthapar: > This shouldn’t be necessary, `docker_deps` in Shipyard’s `Makefile.images` extracts `COPY` statements to calculate prerequisites.

I get following error without this:

```
Dockerfile.submariner-diagnose:6
--------------------
   4 |     
   5 |     # install subctl binary
   6 | >>> COPY bin/subctl /subctl
   7 |     
   8 |     USER ${USER_UID}
--------------------
error: failed to solve: failed to compute cache key: "/bin/subctl": not found
make: *** [/opt/shipyard/Makefile.images:24: package/.image.submariner-diagnose] Error 1
```
Maybe the extraction logic in shipyard isn't working? Or if it is a recent, my base shipyard image is stale?

skitt (approved):

package/Dockerfile.submariner-diagnose
```diff
@@ -0,0 +1,8 @@
+FROM scratch
+
+ENV USER_UID=1001 PATH=/
+
+# install subctl binary
+COPY bin/subctl /subctl
```
mkolesnik: Since this is essentially packaging `subctl` shouldn't the image be just named `submariner-subctl` or event just `subctl`?

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
```diff
@@ -84,10 +85,16 @@ func ensureServiceAccounts(kubeClient kubernetes.Interface, namespace string) (b
 		return false, errors.Wrap(err, "error provisioning globalnet ServiceAccount resource")
 	}
 
+	createdDiagnoseSA, err := serviceaccount.EnsureFromYAML(kubeClient, namespace,
```
mkolesnik, line 88: I'm not sure why this needs to be the responsibility of submariner-operator?

Since this a standalone capability, shouldn't the "diagnose operator" be the one ensuring this service account?

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
```diff
@@ -112,9 +119,14 @@ func ensureClusterRoles(kubeClient kubernetes.Interface) (bool, error) {
 		return false, errors.Wrap(err, "error provisioning globalnet ClusterRole resource")
 	}
 
+	createdDiagnoseCR, err := clusterrole.EnsureFromYAML(kubeClient, embeddedyamls.Config_rbac_submariner_diagnose_cluster_role_yaml)
```
mkolesnik, line 122: Same thing about the roles

mkolesnik (dismissed):
Some comments, also shouldn't the EP be merged before implementation? https://github.com/submariner-io/enhancements/pull/60

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
vthapar: We don't have any diagnose operator as diagnose doesn't need anything other than relevant RBAC for it. So added to `submariner-operator` as it was the best fit. Creating a whole new operator just to deploy RBAC seems to be overkill. Note that we do deploy globalnet SAs etc. even on deployments without globalnet, so we have a precedent.

package/Dockerfile.submariner-diagnose
vthapar: Yes, I was going to name it `submariner-subctl` too but wasn't sure at the time if we will need anything more than subctl going forward for diagnostics. Am okay with either, we can always revisit this if need be in future. `submariner-subctl` or `subctl` ?

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
mkolesnik: Ah from the original EP I thought that there's going to be an operator, because of:

> * A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
>  to trigger running diagnose from OCM UI.`
> * A new `SubmarnerDiagnoseController` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose

So if that all lives in OCM this part can just as well live there, and the proposed controller will make sure it has the correct RBAC.
I guess it's more relevant to that specific use case?

Also the fact that globalnet RBAC is deployed even without globalnet is sloppy and probably wrong? But that's not an excuse IMHO

package/Dockerfile.submariner-diagnose
mkolesnik: IDK we already have `nettest` I think in that light we can call it simply `subctl` but I'm fine either way, whatever seems most logical to everyone else

skitt: I vote for `subctl`, it would end up being `quay.io/submariner/subctl`.

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
vthapar: @mkolesnik This will also allow non OCM deployments to run diagnostics by just creating the Job. This enables use cases like creating a CronJob to run diagnostics periodically, run Diagnostics using golang instead shell scripts, run Diagnose as part of our E2E, run it as part of our operator deployment to make sure everything deployed correctly etc. Permissions required for diagnose are very closely tied to internals of diagnose. Having it in operator will help make sure we're updating RBAC when we add new features to diagnose too.

Agree on Globalnet RBAC, it is just bad security practice.

mkolesnik: I agree this can be useful for running diagnose by just creating the job, but I don't think this will be done much if at all.

Is `subctl diagnose` going to run the job and collect the result?

Also I'm still uneasy with just deploying these permissions even though they might never be used.
I would rather they be deployed on a need-for basis, e.g. we could have a flag in the `Submariner` CR indicating this capability is enabled and that would deploy these (or make sure they're deleted if changed from enabled to disabled).

Consider from a security standpoint a model of "less attack surface" in this context, the less permissions we're giving out needlessly, the more secure the system is.

Makefile
mkolesnik: We should probably fix this on Shipyard then, have you opened an issue there?

mkolesnik:
> shouldn't the EP be merged before implementation

@vthapar you missed my general comment re the EP ^

pkg/subctl/operator/submarinerop/serviceaccount/ensure.go
skitt: The new SA can’t be used unless you have the corresponding token; and if anyone is in a position to get the token, we’ve lost anyway (they can get the operator token, which is far more useful).

I do think the roles currently grant permissions that aren’t needed for `diagnose`, it would be worth checking which ones can be removed (especially those granting the ability to make changes).

Makefile
skitt: https://github.com/submariner-io/shipyard/issues/727

vthapar:
@skitt @mkolesnik Plan was always to reduce the permissions to bare minimum in a subsequent PR. I started with same as operator coz had those working and needed that to work on addon changes. With the PR delayed, I have minimal permissions patch ready and pushed as part of this PR itself.

config/rbac/submariner-diagnose/service_account.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: submariner-diagnose
```
mkolesnik, line 5: Should it be all named `submariner-diagnose` or maybe `submariner-subctl` if we want to enable to run arbitrary subctl commands?

config/rbac/submariner-diagnose/cluster_role.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - nodes
+    verbs:
+      - get
+      - list
+  - apiGroups:  # pods are created to run firewall diagnostics
+      - ""
+    resources:
+      - pods
+    verbs:
+      - create
```
mkolesnik, line 22: Is it necessary to create pods on the `ClusterRole`? You already grant create pods on the `Role`

mkolesnik (commented):
Was wondering about the RBAC being "targeted" towards `subctl diagnose` but you're generally adding a capability to run any `subctl` commands, so shouldn't RBAC be more general or do we intend to create separate RBAC for each command under `subctl`?

config/rbac/submariner-diagnose/cluster_role.yaml
vthapar: Yes. Firewall test creates pods in "default" namespace. Permissions in `Role` aren't enough.

config/rbac/submariner-diagnose/service_account.yaml
vthapar: Permissions are very specific to run `diagnose`. e.g. `gather` or `validate` will not work.

vthapar:
> Was wondering about the RBAC being "targeted" towards `subctl diagnose` but you're generally adding a capability to run any `subctl` commands, so shouldn't RBAC be more general or do we intend to create separate RBAC for each command under `subctl`?

While we can run any, for now only use case is diagnose and permissions are very specific to it. We can revisit this when we add option to run other commands. Also, for some of other commands `submariner-operator` would suffice. Diagnose needed permissions not present in `submariner-operator` else could've just reused it too.

config/rbac/submariner-diagnose/cluster_role.yaml
mkolesnik: I wonder if its possible to bind the existing `Role` to the `default` namespace instead of granting cluster wide `pod create` permissions?

mkolesnik:
@skitt WDYT

config/rbac/submariner-diagnose/cluster_role.yaml
skitt: That should be possible, yes.

skitt: https://github.com/submariner-io/submariner-operator/issues/1885

skitt (approved):

config/rbac/submariner-diagnose/cluster_role.yaml
vthapar: @skitt Is it possible to bind role to two namespaces? We also need pod creation in `submariner-operator` namespace to allow Job to run the pods. It was failing for me without pod create permission in the role. Or create two roles? Alternately, why not modify `diagnose` to create pods in operator namespace?

skitt: No, a role is a namespaced object; so you’d have to create it in all namespaces where it’s needed, and bind the SA to all the corresponding roles. We could use a single namespace for everything instead.

config/rbac/submariner-diagnose/cluster_role.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: submariner-diagnose
+rules:
+  # submariner-diagnose runs subctl diagnose to troubleshoot submariner
+  # dpeloyment
```
dfarrell07, line 8:
```suggestion
  # deployments
```

dfarrell07 (approved):
LGTM, typo is non-blocker I think

config/rbac/submariner-diagnose/cluster_role.yaml
mkolesnik: So it should be possible to create the `Role` and `RoleBinding` for that in the default namespace, or just use the single namespace for diagnosis (which makes sense anyhow I guess).

mkolesnik (approved):

PR #1815, opened 2022-01-31
Ship our embedded YAMLs and check they're up-to-date

skitt:
We would like dependent projects to be able to use (indirectly) our
CRDs in particular. Doing so requires shipping them as code, so we
need to publish yamls.go.

The linting jobs are updated to check that yamls.go is up-to-date.


skitt:
The GHA is verified by https://github.com/submariner-io/submariner-operator/runs/5008922791?check_suite_focus=true (which correctly failed).

dfarrell07 (approved):

pkg/embeddedyamls/yamls.go
```diff
@@ -0,0 +1,3151 @@
+/*
```
Jaanki, line 1: now that this file is available, we can remove the logic to generate them on various `make` invocations.

skitt: We still need to ensure it’s always up-to-date; I’d rather leave the prerequisites as-is in the Makefile.

Jaanki (approved):

tpantelis (approved):

PR #1817, opened 2022-02-01
Copy export to its own package

Jaanki:
To better understand the changes, review commit https://github.com/submariner-io/submariner-operator/pull/1817/commits/c822c36584e4f6764c3866c1d0c14899ada70c3e against https://github.com/submariner-io/submariner-operator/pull/1817/commits/2dc7317355d236e23c518b3301dbf5fe439076a7

### DIFF
--- a/cmd/subctl/export.go
+++ b/cmd/subctl/export.go
@@ -1,0 +1,91 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subctl
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/submariner-io/submariner-operator/internal/cli"
+	"github.com/submariner-io/submariner-operator/internal/exit"
+	"github.com/submariner-io/submariner-operator/pkg/client"
+	"github.com/submariner-io/submariner-operator/pkg/service"
+	"k8s.io/client-go/kubernetes/scheme"
+	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+)
+
+var (
+	exportCmd = &cobra.Command{
+		Use:   "export",
+		Short: "Exports a resource to other clusters",
+		Long:  "This command exports a resource so it is accessible to other clusters",
+	}
+	exportServiceCmd = &cobra.Command{
+		Use:   "service <serviceName>",
+		Short: "Exports a Service to other clusters",
+		Long: "This command creates a ServiceExport resource with the given name which causes the Service of the same name to be accessible" +
+			" to other clusters",
+		PreRunE: restConfigProducer.CheckVersionMismatch,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := validateArguments(args)
+			exit.OnErrorWithMessage(err, "Insufficient arguments")
+
+			status := cli.NewReporter()
+
+			config, err := restConfigProducer.ForCluster()
+			exit.OnError(status.Error(err, "Error creating REST config"))
+
+			clientConfig := restConfigProducer.ClientConfig()
+			if serviceNamespace == "" {
+				if serviceNamespace, _, err = clientConfig.Namespace(); err != nil {
+					serviceNamespace = "default"
+				}
+				exit.OnErrorWithMessage(err, "Error getting service Namespace")
+			}
+
+			clientProducer, err := client.NewProducerFromRestConfig(config)
+			exit.OnError(status.Error(err, "Error creating client producer"))
+
+			err = service.Export(clientProducer, serviceNamespace, args[0], status)
+			exit.OnErrorWithMessage(err, "Failed to export Service")
+		},
+	}
+	serviceNamespace string
+)
+
+func init() {
+	err := mcsv1a1.Install(scheme.Scheme)
+	exit.OnErrorWithMessage(err, "Failed to add to scheme")
+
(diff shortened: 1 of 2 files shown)


pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+	resourceServiceExport, gvr, err := autil.ToUnstructuredResource(mcsServiceExport, restMapper)
+	if err != nil {
+		return status.Error(err, "Failed to convert to Unstructured")
+	}
+
+	_, err = clientProducer.ForDynamic().Resource(*gvr).Namespace(serviceNamespace).
+		Create(context.TODO(), resourceServiceExport, metav1.CreateOptions{})
+	if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			return errors.New("Service already exported")
```
tpantelis: This isn't an error. 

```suggestion
                         status.Success("Service already exported")
			return nil
```

pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/submariner-io/submariner-operator/internal/cli"
+	"github.com/submariner-io/submariner-operator/internal/restconfig"
+	"github.com/submariner-io/submariner-operator/pkg/client"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+)
+
+func Export(restConfigProducer restconfig.Producer, serviceNamespace, svcName string) error {
```
tpantelis: We  can't pass `restconfig.Producer` to the public lib functions since it's internal. Pass a `client.Producer` instead.

pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+	status := cli.NewReporter()
```
tpantelis: Pass in the status `reporter.Interface` to the lib functions.

pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+	if serviceNamespace == "" {
+		if serviceNamespace, _, err = clientConfig.Namespace(); err != nil {
+			serviceNamespace = "default"
+		}
+	}
+
+	err = mcsv1a1.AddToScheme(scheme.Scheme)
+	if err != nil {
+		return status.Error(err, "Failed to add to scheme")
+	}
```
tpantelis: This only needs to be done once so I think we should do it in an `init` function and panic on error.

pkg/service/export.go
skitt: This should be provided as an argument to `Export()` (so that non-CLI users can use the function).

pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+limitations under the License.
+*/
+
+package service
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	autil "github.com/submariner-io/admiral/pkg/util"
```
skitt: This import alias isn’t necessary, there are no longer any other `util` packages used here.

pkg/service/export.go
```diff
@@ -0,0 +1,96 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      svcName,
+			Namespace: serviceNamespace,
+		},
+	}
+
+	resourceServiceExport, gvr, err := autil.ToUnstructuredResource(mcsServiceExport, restMapper)
+	if err != nil {
+		return status.Error(err, "Failed to convert to Unstructured")
+	}
```
tpantelis: We don't really need to dynamically find the GVR since we know that info directly. Hence we don't need to build a `RESTMapper` and thus we don't need a `rest.Config` which simplifies the code. 

```suggestion
        import ""github.com/submariner-io/admiral/pkg/resource"
        ...
	resourceServiceExport, err := resource.ToUnstructured(mcsServiceExport)
	if err != nil {
		return status.Error(err, "Failed to convert to Unstructured")
	}
	
	gvr := &schema.GroupVersionResource{
		Group:    mcsv1a1.GroupVersion.Group,
		Version:  mcsv1a1.GroupVersion.Version,
		Resource: "serviceexports",
	}
```

skitt (changes requested):

tpantelis (changes requested):

cmd/subctl/export.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	}
+	serviceNamespace string
+)
+
+func init() {
+	// Using Install as AddToScheme is deprecated.
+	// https://pkg.go.dev/sigs.k8s.io/mcs-api/pkg/apis/v1alpha1#pkg-variables
+	err := mcsv1a1.Install(scheme.Scheme)
+    exit.AndPanicOnError(err)
```
tpantelis: My previous comment suggested to panic but we really don't need to - we can just do `exit.OnErrorWithMessage`.

tpantelis (changes requested):
Looks good - just one more minor comment.

cmd/subctl/export.go
Jaanki: I like `panic` better. But its not used at any other place so removing it and using `exit.OnErrorWithMessage`

cmd/subctl/export.go
```diff
@@ -0,0 +1,93 @@
...
+		},
+	}
+	serviceNamespace string
+)
+
+func init() {
+	// Using Install as AddToScheme is deprecated.
+	// https://pkg.go.dev/sigs.k8s.io/mcs-api/pkg/apis/v1alpha1#pkg-variables
+	err := mcsv1a1.Install(scheme.Scheme)
+	exit.OnError(err)
```
tpantelis:
```suggestion
	exit.OnErrorWithMessage(err)
```

tpantelis (changes requested):

cmd/subctl/export.go
```diff
@@ -0,0 +1,93 @@
...
+
+			err = service.Export(clientProducer, serviceNamespace, args[0], status)
+			exit.OnErrorWithMessage(err, "Failed to export Service")
+		},
+	}
+	serviceNamespace string
+)
+
+func init() {
+	// Using Install as AddToScheme is deprecated.
```
skitt: This isn’t necessary as a code comment; it’s explaining a change so it should go in the commit message.

skitt (changes requested):
Minor change on top of Tom’s request.

tpantelis (approved):

skitt (approved):

PR #1828, opened 2022-02-04
Defer calls to status.End()

skitt:
Instead of calling status.End() explicitly, which involves some
tedious handling around return calls, defer it. Since status.Start()
calls status.End() anyway (in implementations where status.End() does
anything), we can rely on that to avoid calling status.End()
explicitly before status.Start() in function with multiple
status.Start() invocations.

Update the documentation to make the contracts explicit.


pkg/broker/file.go
```diff
@@ -70,14 +71,7 @@ func WriteInfoToFile(restConfig *rest.Config, brokerNamespace, ipsecFile string,
 		data.CustomDomains = &customDomains
 	}
 
-	err = data.writeToFile(InfoFileName)
-	if err != nil {
-		return status.Error(data.writeToFile(InfoFileName), "error saving broker info")
```
skitt, line 75: This also avoids a double call to `data.writeToFile()` on errors.

dfarrell07 (approved):

pkg/broker/file.go
tpantelis: Thanks for the cleanup. I don't know what I was thinking 😄

cmd/subctl/join.go
```diff
@@ -293,8 +289,6 @@ func getNetworkDetails(clientProducer client.Producer, status reporter.Interface
 		status.Warning("No network details discovered")
 	}
 
-	status.End()
```
tpantelis: This actually changes the output behavior, ie the details would now be printed on L299 before the status message which we don't want.

cmd/subctl/join.go
```diff
@@ -159,8 +158,6 @@ func possiblyLabelGateway(kubeClient kubernetes.Interface, status reporter.Inter
 	workerNodes, err := nodes.GetAllWorkerNames(kubeClient)
 	exit.OnError(status.Error(err, "Error listing the worker nodes"))
 
-	status.End()
```
tpantelis: We want to keep this so L165 isn't part of the previous start/end section.

pkg/reporter/reporter.go
```diff
@@ -42,15 +43,17 @@ type Interface interface {
...
 	Error(err error, message string, args ...interface{}) error
 }
 
 func HandleError(reporter Interface, err error, message string, args ...interface{}) error {
-	err = errors.Wrapf(err, message, args...)
 	if err == nil {
 		return nil
 	}
 
+	err = errors.Wrapf(err, message, args...)
```
tpantelis, line 55: Why was this changed since `Wrapf` handles nil?  It's fine either was, just wondering.

tpantelis (changes requested):

pkg/reporter/reporter.go
skitt: It’s just a small optimisation to avoid a function call we know isn’t necessary.

cmd/subctl/join.go
skitt: Ah yes, good point. I actually thought about using `func{}()` constructs, JavaScript-style, but that ends up too cumbersome for a general pattern.

tpantelis (approved):

PR #1832, opened 2022-02-07
Copy pkg/subctl/show to pkg/show

Jaanki:
Decouple subctl part out of pkg/show

and use `reporter.Interface`, `execute.OnMultiCluster`
implementations.

Move `table` pkg into `show` pkg because it is only
used by `show connections` currently.


tpantelis:
The show command seems specific to subctl and thus perhaps should be kept internal. @skitt @sridhargaddam WDYT

skitt (changes requested):
I agree, this should be in an internal package (apart from the `cmd` part).

Issue #1730, opened 2021-12-28
 Issue with discovery of ClusterCIDRs on K3s


ccwalterhk:
**What happened**: Issue with discovery of ClusterCIDRs on K3s.  The clusterCIDRs are configured as 10.44.0.0/16 and 10.144.0.0/16.  However, when submariner does the discovery, it discover /24.

```
For Cluster A:        
    Discovered network details via Submariner:
        Network plugin:  generic
        Service CIDRs:   [10.45.0.0/16]
        Cluster CIDRs:   [10.44.0.0/24]    <---- This is the issue

For Cluster B:
    Discovered network details via Submariner:
        Network plugin:  generic
        Service CIDRs:   [10.145.0.0/16]
        Cluster CIDRs:   [10.144.0.0/24]   <---- This is the issue
```

**What you expected to happen**: Submariner should discover as /16 subnet.

**How to reproduce it (as minimally and precisely as possible)**:
Follow the exact steps in URL https://submariner.io/getting-started/quickstart/k3s/.  below is exact steps to build two clusters

```
Cluster a:
POD_CIDR=10.44.0.0/16
SERVICE_CIDR=10.45.0.0/16
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--cluster-cidr $POD_CIDR --service-cidr $SERVICE_CIDR" sh -s -

sudo snap install yq

sudo su

cp /etc/rancher/k3s/k3s.yaml kubeconfig.cluster-a
export IP=$(hostname -I | awk '{print $1}')
yq -i eval \
'.clusters[].cluster.server |= sub("127.0.0.1", env(IP)) | .contexts[].name = "cluster-a" | .current-context = "cluster-a"' \
kubeconfig.cluster-a

Cluster b:

POD_CIDR=10.144.0.0/16
SERVICE_CIDR=10.145.0.0/16
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--cluster-cidr $POD_CIDR --service-cidr $SERVICE_CIDR" sh -s -

sudo snap install yq

sudo su

cp /etc/rancher/k3s/k3s.yaml kubeconfig.cluster-b
export IP=$(hostname -I | awk '{print $1}')
yq -i eval \
'.clusters[].cluster.server |= sub("127.0.0.1", env(IP)) | .contexts[].name = "cluster-b" | .current-context = "cluster-b"' \
kubeconfig.cluster-b
```

**Anything else we need to know?**:

[...]

PR #1833, opened 2022-02-07
Extend network discovery logging

yboaron:
Currently, we only log the bottom line of network discovery (plugin and cidrs),
in some cases (e.g: generic plugin) it is not clear from this information
how exactly the data was retrieved.

With this PR we should be able to deduce the source of network discovery information
also for cases like generic plugin, this additional information should be useful in cases like #1730 and
submariner-io/submariner-operator#1852   

Depends on #1802

### DIFF
--- a/pkg/discovery/network/canal.go
+++ b/pkg/discovery/network/canal.go
@@ -21,6 +21,7 @@ package network
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/submariner/pkg/routeagent_driver/constants"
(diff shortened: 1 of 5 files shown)


tpantelis (commented):
How about we use the `reporter.Interface` to output additional information?

yboaron:
IIUC, currently both subctl  [1] and operator [2] code don't log network discovery information using 'reporter.interface'.

I assume that we want to output all the network discovery information using 'reporter.interface' and not just the info added in this PR,
 can you please elaborate on how you think we should output only the additional information using reporter.interface ?

[1]
https://github.com/submariner-io/submariner-operator/blob/bce0cc7af103c5033cc82170cc63b362bcfbe667/pkg/discovery/network/network.go#L42
[2] 
https://github.com/submariner-io/submariner-operator/blob/bce0cc7af103c5033cc82170cc63b362bcfbe667/pkg/discovery/network/network.go#L55

dfarrell07:
@yboaron This needs a manual rebase.

nyechiel:
Not urgent for 0.12, moving to the backlog so that we can merge for 0.13.

nyechiel:
@yboaron I am reviewing old issues and came across this one. Is this still relevant?

yboaron:
I'm going to close this PR, in case we'll have new issues relevant to this subject we can address them with updated PR.

PR #1835, opened 2022-02-08
Copy gather to internal/

Jaanki:
and add an option to specify io.Writer for
all printfs defaulting to `os.Stdout` for `subctl.`


pkg/gather/gather.go
```diff
@@ -0,0 +1,256 @@
...
+	component.Connectivity:     gatherConnectivity,
+	component.ServiceDiscovery: gatherDiscovery,
+	component.Broker:           gatherBroker,
+	component.Operator:         gatherOperator,
+}
+
+func Data(clusterInfo *cluster.Info, status reporter.Interface, options Options, writer io.Writer) bool {
+	if options.Directory == "" {
+		options.Directory = "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
+	}
```
tpantelis: The codes creates files so it doesn't seem conducive as a library. I think we should keep all this internal.

pkg/gather/gather.go
```diff
@@ -0,0 +1,256 @@
...
+	_, _ = fmt.Fprintf(writer, "Files are stored under directory %q\n", options.Directory)
+
+	return true
+}
+
+func gatherDataByCluster(clusterInfo *cluster.Info, status reporter.Interface, options Options, writer io.Writer) {
+	var err error
+	clusterName := clusterInfo.Name
+
+	_, _ = fmt.Fprintf(writer, "Gathering information from cluster %q\n", clusterName)
```
tpantelis: Why are you using an `io.Writer` for progress output and not the `reporter.Interface`?

tpantelis (changes requested):

pkg/gather/gather.go
Jaanki: I did use `reporter.Interface` but was advised to not change it [here](https://github.com/submariner-io/submariner-operator/pull/1801#discussion_r794659155). The idea is to give options to users to log these informational info anywhere they want.

tpantelis: Providing two avenues for users to log info would be confusing. This works for subctl b/c it all goes to `fmt.Printf`. It's basically tailored to subctl. This is another reason why I think `gather` is specific to subctl and should not be exposed publicly as a lib.

Jaanki: I think there are plans to use `gather` in ACM UI as well, isn't it @vthapar ?

vthapar: @Jaanki We haven't yet figured out how to make gather work from UI. Note that UI doesn't use Go and can't use Go APIs. Similar to diagnose we will have to run a pod in cluster to run gather and then make results [maybe a tar file] available to UI. Most likely it would be running `subctl gather`, not much different from what we're planning with diagnose.

I kinda agree with @tpantelis There might be some use case to provide API to gather IP tables etc. but for pod logs and resources not much. We can park this requirement unless we come up with actual use case, whether it is from UI or otherwise. Don't want to do some changes which still don't work for UI and end up being dead code.

aswinsuryan:
@Jaanki the lighthouse tests are failing since` framework.ClusterA` is empty string in [1]. Was there any changes in this area?

https://github.com/submariner-io/lighthouse/blob/devel/test/e2e/discovery/statefulsets.go#L183

PR #1838, opened 2022-02-09
Rhos dedicated gw

aswinsuryan:
Add support for dedicated GW node in RHOS clusters

- gateway node of the mentioned type will be created.
- the number of gateway nodes can be configured.

### DIFF
--- a/pkg/subctl/cmd/cloud/prepare/rhos.go
+++ b/pkg/subctl/cmd/cloud/prepare/rhos.go
@@ -38,6 +38,9 @@ func newRHOSPrepareCommand() *cobra.Command {
 	rhos.AddRHOSFlags(cmd)
 	cmd.Flags().IntVar(&gateways, "gateways", DefaultNumGateways,
 		"Number of gateways to deploy")
+	cmd.Flags().StringVar(&rhosGWInstanceType, "gateway-instance", "PnTAE.CPU_16_Memory_32768_Disk_80", "Type of gateway instance machine")
+	cmd.Flags().BoolVar(&dedicatedGateway, "dedicated-gateway", false,
+		"Whether a dedicated gateway node has to be deployed (default false)")
 
 	return cmd
 }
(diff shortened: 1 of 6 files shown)


nyechiel:
@aswinsuryan can you please add more info in the commit message? https://submariner.io/development/code-review/#document-why-in-commit-messages

dfarrell07:
Note that the message-docs should be in the commit itself (https://github.com/submariner-io/submariner-operator/pull/1838/commits/903c53ce05db465ab247e8b9318f09e59af8b558), not just the PR.

pkg/subctl/cmd/cloud/prepare/rhos.go
```diff
@@ -38,6 +38,9 @@ func newRHOSPrepareCommand() *cobra.Command {
 	rhos.AddRHOSFlags(cmd)
 	cmd.Flags().IntVar(&gateways, "gateways", DefaultNumGateways,
 		"Number of gateways to deploy")
+	cmd.Flags().StringVar(&rhosGWInstanceType, "gateway-instance", "PnTAE.CPU_16_Memory_32768_Disk_80", "Type of gateway instance machine")
```
sridhargaddam, line 41: The flavor `PnTAE.CPU_16_Memory_32768_Disk_80` looks like a custom flavor assigned to a specific team. Can we identify an equivalent generic flavor and use it as the default as the current one may not be available in all deployments.

List of flavors:
https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.0/html/instances_and_images_guide/ch-manage_instances#section-flavors

sridhargaddam (commented):
Thanks for the PR @aswinsuryan. One small comment.

pkg/subctl/cmd/cloud/prepare/rhos.go
sridhargaddam: Alternately, the other option would be to leave it empty and pick the same flavor as one of the existing worker nodes of the Cluster.

aswinsuryan: I took this flavor from one of the worker nodes, thought this is openshift specific flavor.

tpantelis (approved):

sridhargaddam (approved):

Issue #1329, opened 2021-05-04
subctl should be refactored to a library, and make subctl the CLI of that library


mangelajo:
**What would you like to be added**:

Splitting subctl in two parts, an interface part, and a library part, the library would contain all the logic
that subctl uses, and a defined Interface that external projects can consume

**Why is this needed**:

In this way, other projects can use the subctl logic to deploy clusters and brokers in an automated
way without calling the subctl logic from the command line.

**Work items:**
- [x] Rename apis -> api
    - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/1651
    - [x] Depends on https://github.com/open-cluster-management/submariner-addon/pull/220
- [x] Move all cobra related stuff to cmd/subctl, Logic called by subctl inside pkg/cmd/subctl for below commands
    - [x] deploy-broker Depends on https://github.com/submariner-io/submariner-operator/pull/1663
    - [x] join Depends on https://github.com/submariner-io/submariner-operator/pull/1726
    - [x] benchmark Depends on https://github.com/submariner-io/submariner-operator/pull/1772
    - [x] verify Depends on https://github.com/submariner-io/submariner-operator/pull/1795
    - [x] version Depends on https://github.com/submariner-io/submariner-operator/pull/1663 
    - [x] diagnose
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/1994
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2015
    - [x] show
      - [x] Changes made in [Print out brokers information on subctl show all](https://github.com/submariner-io/submariner-operator/pull/1924/files#)
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2006
    - [x] gather
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/1801
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2007
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2043
    - [x] cloud 
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/1770
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2013
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2041
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2049
      - [x] Depends on https://github.com/submariner-io/submariner-operator/pull/2048
[...]

PR #1840, opened 2022-02-09
Adding external gateway as a CLI option

sridhargaddam:
Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>
Depends on https://github.com/submariner-io/submariner-operator/issues/1329

### DIFF
--- a/cmd/subctl/join.go
+++ b/cmd/subctl/join.go
@@ -20,6 +20,7 @@ package subctl
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -31,6 +32,7 @@ import (
 	"github.com/submariner-io/submariner-operator/internal/nodes"
 	"github.com/submariner-io/submariner-operator/pkg/broker"
 	"github.com/submariner-io/submariner-operator/pkg/client"
+	"github.com/submariner-io/submariner-operator/pkg/discovery/globalnet"
 	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
 	"github.com/submariner-io/submariner-operator/pkg/join"
 	"github.com/submariner-io/submariner-operator/pkg/reporter"
@@ -67,6 +69,7 @@ var joinCmd = &cobra.Command{
 		networkDetails := getNetworkDetails(clientProducer, status)
 		determinePodCIDR(networkDetails, status)
 		determineServiceCIDR(networkDetails, status)
+		determineExternalCIDR(networkDetails, status)
 
 		if brokerInfo.IsConnectivityEnabled() && labelGateway {
 			possiblyLabelGateway(clientProducer.ForKubernetes(), status)
@@ -91,6 +94,7 @@ func addJoinFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&joinFlags.ClusterID, "clusterid", "", "cluster ID used to identify the tunnels")
 	cmd.Flags().StringVar(&joinFlags.ServiceCIDR, "servicecidr", "", "service CIDR")
 	cmd.Flags().StringVar(&joinFlags.ClusterCIDR, "clustercidr", "", "cluster CIDR")
+	cmd.Flags().StringVar(&joinFlags.ExternalCIDR, "externalcidr", "", "external CIDR")
 	cmd.Flags().StringVar(&joinFlags.Repository, "repository", "", "image repository")
 	cmd.Flags().StringVar(&joinFlags.ImageVersion, "version", "", "image version")
 	cmd.Flags().StringVar(&ignoredColorCodes, "colorcodes", "", "color codes")
@@ -323,4 +327,26 @@ func determineServiceCIDR(nd *network.ClusterNetwork, status reporter.Interface)
 		joinFlags.ServiceCIDR, err = askForCIDR("Service")
 		exit.OnError(status.Error(err, "Error collecting CIDR"))
 	}
+}
+
+// determineExternalCIDR validates and appends the external cidr to cluster cidr.
+func determineExternalCIDR(nd *network.ClusterNetwork, status reporter.Interface) {
+	if joinFlags.ExternalCIDR == "" {
+		return
+	}
+
+	err := globalnet.IsValidCIDR(joinFlags.ExternalCIDR)
+	if err != nil {
+		exit.OnError(status.Error(err, "Error collecting CIDR"))
+	}
+
+	_, externalCIDR, err := net.ParseCIDR(joinFlags.ExternalCIDR)
+	if err != nil {
+		exit.OnError(status.Error(err, "Error collecting CIDR"))
+	}
+
+	joinFlags.ExternalCIDR = externalCIDR.String()
+
+	nd.PodCIDRs = append(nd.PodCIDRs, joinFlags.ExternalCIDR)
+	joinFlags.ClusterCIDR = strings.Join(nd.PodCIDRs, ",")
 }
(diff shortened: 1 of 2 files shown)


pkg/subctl/cmd/join.go
```diff
@@ -378,6 +383,21 @@ func determineServiceCIDR(nd *network.ClusterNetwork) error {
 	return nil
 }
 
+// determineExternalCIDR validates and appends the external cidr to cluster cidr
+func determineExternalCIDR(nd *network.ClusterNetwork) error {
+	if externalCIDR != "" {
+		err := utils.IsValidCIDR(externalCIDR)
+		if err == nil {
+			externalCIDR = utils.GetNetworkAddress(externalCIDR)
+			nd.PodCIDRs = append(nd.PodCIDRs, externalCIDR)
```
sridhargaddam: This alone will not help because nd.PodCIDRs is mainly used to give a warning (line 357) to the user when the user-supplied information does not match with the auto-discovered values.

To get this working, please assign it back to clusterCIDR as shown below at L393
clusterCIDR = strings.Join(nd.PodCIDRs,",")

clusterCIDR is the one that gets added to the submarinerSpec in Line 497
When clusterCIDR is not explicitly specified, the submariner-operator will auto-discover the clusterCIDRs and use it when installing the Submariner Pods.

nerdalert: Done, thanks Sridhar!

nerdalert:
This PR adds the ability for a user to pass in the external cidr network functionality as laid out in the [external network docs (non-kube clustered infra)](https://submariner.io/getting-started/quickstart/external/). 

- The existing implementation of overloading `CLUSTER_CIDR` is not affected by this patch. 
- Since this still overloads `clusterCIDR` in the codebase, there was no need to modify the CRD to implement this. However, as a result of the least invasive approach, external cidr networks will show up in `subctl networks show` as a cluster cidr. As this functionality matures we will revisit making the status reporting a first class citizen.
- How we want to reflect the changes in the docs and e2e is on the agenda for the weekly [modeling non-cluster connectivity](https://calendar.google.com/calendar?cid=NHFuZGVoOGY0bzZ1ajlvZnBsczh1NWNlZ2tAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ)

Ty!

pkg/subctl/cmd/join.go
```diff
@@ -378,6 +383,22 @@ func determineServiceCIDR(nd *network.ClusterNetwork) error {
...
+			clusterCIDR = strings.Join(nd.PodCIDRs, ",")
+		} else {
+			return err
+		}
+	}
+
+	return nil
```
tpantelis: We tend to structure code with early returns to reduce nesting and I think this style is more or less idiomatic in Go.

```suggestion
	if externalCIDR == "" {
	    return nil
	}
	
        err := utils.IsValidCIDR(externalCIDR)
	if err != nil {
	    return err
	}
		
        externalCIDR = utils.GetNetworkAddress(externalCIDR)
        nd.PodCIDRs = append(nd.PodCIDRs, externalCIDR)
        clusterCIDR = strings.Join(nd.PodCIDRs, ",")

	return nil
```

pkg/subctl/cmd/utils/utils.go
```diff
@@ -94,3 +99,32 @@ func GetSubmarinerResource(config *rest.Config) *v1alpha1.Submariner {
 
 	return submariner
 }
+
+// IsValidCIDR validates the supplied string is a valid v4/v6 network address
+func IsValidCIDR(cidr string) error {
```
tpantelis: This function is already defined in _pkg/discovery/globalnet/globalnet.go_.

pkg/subctl/cmd/utils/utils.go
```diff
@@ -94,3 +99,32 @@ func GetSubmarinerResource(config *rest.Config) *v1alpha1.Submariner {
...
+	if ip.IsLinkLocalMulticast() {
+		return fmt.Errorf("%s can't be in link-local multicast range", cidr)
+	}
+
+	return nil
+}
+
+// GetNetworkAddress returns the network address from a given CIDR
+func GetNetworkAddress(str string) string {
+	_, cidr, _ := net.ParseCIDR(str)
```
tpantelis: I don't think we should ignore the error. As such we might as well just inline `net.ParseCIDR` as we do in other places. Either way it's recommended to avoid `utils` packages anyway and this file will be going way once the code restructure is done.

tpantelis (changes requested):
I think we should add this to the new _cmd/subctl_ code since this is scheduled to land in 0.12.0.

pkg/subctl/cmd/utils/utils.go
nerdalert: Thanks Tom, I was thinking moving it to a utils package made sense but didn't realize it was going away. If cool, I will import from `pkg/discovery/globalnet/`.

pkg/subctl/cmd/join.go
nerdalert: ahhh, I thought it was vice versa. will move it over to the `cmd/subctl` package. ty for the info.

pkg/subctl/cmd/utils/utils.go
```diff
@@ -33,6 +33,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	// SubmarinerNamespace default namespace for submariner operations
+	SubmarinerNamespace = "submariner-operator"
+)
```
tpantelis: This change isn't needed.

pkg/subctl/cmd/utils/utils.go
```diff
@@ -72,8 +77,7 @@ func GetSubmarinerResourceWithError(config *rest.Config) (*v1alpha1.Submariner,
 		return nil, errors.Wrap(err, "error creating clientset")
 	}
 
-	// TODO skitt namespace constant
-	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners("submariner-operator").
+	submariner, err := submarinerClient.SubmarinerV1alpha1().Submariners(SubmarinerNamespace).
```
tpantelis: ...nor this.

tpantelis (changes requested):
There's some Go linting errors.

pkg/subctl/cmd/utils/utils.go
nerdalert: I added those as lint was complaining on my first go around but dont see it now ¯\_(ツ)_/¯ 
removed, ty

pkg/subctl/cmd/utils/utils.go
nerdalert: reset to head

nyechiel:
@sridhargaddam is this something we want to have for 0.12? It looks like it's stale

sridhargaddam:
> @sridhargaddam is this something we want to have for 0.12? It looks like it's stale

As Brent mentioned above, this PR improves the readability (by adding new command line flag) while configuring the external-cidrs as part of `subctl join ...` command. So its not critical to have it but its like a good to have feature.

@nerdalert there are some CI failures, can you please look into it and also update the base branch. Thanks.

nerdalert:
@sridhargaddam sorry I missed your comment, rebased and fixed the linting issues. ty!

dfarrell07:
This will need to move to the new subctl repo.

tpantelis:
If the `ClusterCIDR` field in the `Submariner` CRD is empty, the operator will auto-discover it using the underlying CNI. The problem with appending the external CIDR to `ClusterCIDR` in the `subctl` code is that the auto-discovery won't occur unless the user has also specified _--clustercidr_ on the command line. So, to do this right, the external CIDR really needs to be added to the CRD and handled properly by the operator. I think we should park this and revisit if/when the external gateway feature matures and graduates from experimental status.

PR #1854, opened 2022-02-19
Fix GN Pod service account during uninstall

sridhargaddam:
The service-account for Globalnet pod is always named as
submariner-globalnet and this matches with the name of the
Pod. However, when uninstall operation is triggered the
the Globalnet Pod name will be appended with uninstall and
does not match with the service-account. This PR fixes the
issue.

Related to: https://github.com/submariner-io/submariner/issues/1637

### DIFF
--- a/controllers/submariner/globalnet_resources.go
+++ b/controllers/submariner/globalnet_resources.go
@@ -107,7 +107,7 @@ func newGlobalnetDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.DaemonS
 							},
 						},
 					},
-					ServiceAccountName:            name,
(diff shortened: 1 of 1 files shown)


sridhargaddam:
@tpantelis We need this fix to unblock the Globalnet cleanup.

tpantelis (approved):

tpantelis:
Looks like we also need to make a similar change with the [network plugin syncer](https://github.com/submariner-io/submariner-operator/blob/28296f046200f357294bbe067d9891a5e2f368a7/controllers/submariner/np_syncer_resources.go#L101).

Jaanki (approved):

PR #1857, opened 2022-02-20
Fix return value of uninstall.IsSupportedForVersion

sridhargaddam:

### DIFF
--- a/controllers/uninstall/uninstall.go
+++ b/controllers/uninstall/uninstall.go
@@ -325,5 +325,5 @@ func findPodsBySelector(ctx context.Context, clnt client.Reader, namespace strin
 
 func IsSupportedForVersion(version string) bool {
 	semVersion, _ := semver.NewVersion(version)
-	return !(semVersion != nil && !semVersion.LessThan(*minComponentUninstallVersion))
+	return semVersion == nil || !semVersion.LessThan(*minComponentUninstallVersion)
 }
--- a/controllers/servicediscovery/cleanup.go
+++ b/controllers/servicediscovery/cleanup.go
@@ -40,7 +40,7 @@ import (
 )
 
 func (r *Reconciler) doCleanup(ctx context.Context, instance *operatorv1alpha1.ServiceDiscovery) (reconcile.Result, error) {
-	if uninstall.IsSupportedForVersion(instance.Spec.Version) {
+	if !uninstall.IsSupportedForVersion(instance.Spec.Version) {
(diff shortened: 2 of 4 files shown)


controllers/uninstall/uninstall.go
```diff
@@ -325,5 +325,5 @@ func findPodsBySelector(ctx context.Context, clnt client.Reader, namespace strin
 
 func IsSupportedForVersion(version string) bool {
 	semVersion, _ := semver.NewVersion(version)
-	return !(semVersion != nil && !semVersion.LessThan(*minComponentUninstallVersion))
+	return semVersion != nil && !semVersion.LessThan(*minComponentUninstallVersion)
```
sridhargaddam: @tpantelis This PR tries to match the return value with the Function name. 

At the same time, I see that when using KIND deployments, the `submariner` version will be `devel`, so un-installation is not triggered. What were your thoughts about it?

tpantelis: It should trigger for `devel` - this was inadvertently broken by https://github.com/submariner-io/submariner-operator/pull/1843/.  The check should be:

```suggestion
	return semVersion == nil || !semVersion.LessThan(*minComponentUninstallVersion)
```


We should have a test case to cover `devel`. I suggest we change the set up for [this test](https://github.com/submariner-io/submariner-operator/blob/9ed748bcd4db5f639f2c056e875fc9dec882cbf6/controllers/submariner/submariner_controller_test.go#L319) to add `t.submariner.Spec.Version = "devel"`.

sridhargaddam: Done, thanks @tpantelis

controllers/uninstall/uninstall.go
```diff
@@ -325,5 +325,5 @@ func findPodsBySelector(ctx context.Context, clnt client.Reader, namespace strin
 
 func IsSupportedForVersion(version string) bool {
 	semVersion, _ := semver.NewVersion(version)
-	return !(semVersion != nil && !semVersion.LessThan(*minComponentUninstallVersion))
+	return semVersion == nil || !semVersion.LessThan(*minComponentUninstallVersion)
```
skitt, line 328: Does it make sense to return `true` when we don’t know the version?

skitt: Oh, I hadn’t seen the previous comments.

skitt (approved):

Jaanki (approved):

PR #1863, opened 2022-02-22
Add E2E tests for component cleanup

tpantelis:
Delete the `Submariner` resource and verify the uninstall `Daemonset` pods successfully complete. If one doesn't, output its status and pod log. Also verify resources are deleted from the broker.

After the test completes, it re-creates the `Submariner` resource.

Added a similar test for `ServiceDiscovery`.

Depends on https://github.com/submariner-io/lighthouse/pull/691

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,9 +53,9 @@ endif
 
 override CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
-override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG)
+override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG) cluster1 cluster2
 export DEPLOY_ARGS
-override UNIT_TEST_ARGS += cmd internal/cli internal/env internal/log
+override UNIT_TEST_ARGS += test internal/env
 override VALIDATE_ARGS += --skip-dirs pkg/client
 
 # Process extra flags from the `using=a,b,c` optional flag
--- a/controllers/uninstall/uninstall.go
+++ b/controllers/uninstall/uninstall.go
@@ -38,12 +38,23 @@ const (
 	ContainerEnvVar       = "SUBMARINER_UNINSTALL"
 )
 
+type stateType int
+
+const (
+	deleteComponent stateType = iota
+	awaitPodsDeleted
+	createUninstallComponent
+	awaitUninstallComplete
+	uninstallComplete
+)
+
 var minComponentUninstallVersion = semver.New("0.12.0")
 
 type Component struct {
 	Resource          client.Object
 	UninstallResource client.Object
 	CheckInstalled    func() bool
+	state             stateType
 }
 
 type Info struct {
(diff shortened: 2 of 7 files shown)


test/e2e/cleanup/submariner.go
```diff
@@ -0,0 +1,264 @@
...
+
+		submarinerInterface = operatorClient.SubmarinerV1alpha1().Submariners(framework.TestContext.SubmarinerNamespace)
+
+		submariner, err = submarinerInterface.Get(context.TODO(), submarinercr.SubmarinerName, metav1.GetOptions{})
+		Expect(err).To(Succeed())
+
+		brokerRestConfig = getBrokerRestConfig(submariner.Spec.BrokerK8sRemoteNamespace)
+		Expect(brokerRestConfig).ToNot(BeNil(), "No broker located")
+
+		routeAgentPodMonitor = startPodMonitor(names.AppendUninstall(names.RouteAgentComponent), stopCh)
```
sridhargaddam, line 79: Shall we include globalnetPod as well, as it will cover even routeAgent and gatewayPods?

test/e2e/cleanup/submariner.go
```diff
@@ -0,0 +1,264 @@
...
+			_, err := submarinerInterface.Create(context.TODO(), submariner, metav1.CreateOptions{})
+			if err != nil {
+				framework.Errorf("Error re-creating Submariner: %v", err)
+			}
+
+			By("Re-created Submariner resource")
+		}
+	})
+
+	It("should run DaemonSets/Deployments to uninstall components", func() {
```
sridhargaddam, line 112: Both the pods are DaemonSets.

test/e2e/cleanup/submariner.go
```diff
@@ -0,0 +1,264 @@
...
+		// Verify the Cluster and Endpoint resources were removed from the broker.
+
+		submarinerClient, err := submarinerclientset.NewForConfig(brokerRestConfig)
+		Expect(err).To(Succeed())
+
+		clusterID := framework.TestContext.ClusterIDs[framework.ClusterA]
+
+		_, err = submarinerClient.SubmarinerV1().Clusters(framework.TestContext.SubmarinerNamespace).Get(context.TODO(),
+			clusterID, metav1.GetOptions{})
+		Expect(apierrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("Unexpected Cluster %q found on the broker", clusterID))
```
sridhargaddam:
```suggestion
		Expect(apierrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("Unexpected Cluster %q found", clusterID))
```

We are validating in the local cluster (not on broker) right?

test/e2e/cleanup/submariner.go
```diff
@@ -0,0 +1,264 @@
...
+
+		endpoints, err := submarinerClient.SubmarinerV1().Endpoints(framework.TestContext.SubmarinerNamespace).List(
+			context.TODO(), metav1.ListOptions{})
+		Expect(err).To(Succeed())
+
+		for i := range endpoints.Items {
+			Expect(endpoints.Items[i].Spec.ClusterID).ToNot(Equal(clusterID),
+				fmt.Sprintf("Unexpected Endpoint %q found on the broker", endpoints.Items[i].Name))
```
sridhargaddam:
```suggestion
				fmt.Sprintf("Unexpected Endpoint %q found", endpoints.Items[i].Name))
```

sridhargaddam (changes requested):
Thanks for the PR @tpantelis. It catches the gateway pod issue too :D

test/e2e/cleanup/submariner.go
tpantelis: yeah I was going to add globalnet as well.

test/e2e/cleanup/submariner.go
tpantelis: The network plugin syncer is a `Deployment` although we're not checking that right now..

test/e2e/cleanup/submariner.go
tpantelis: It's actually meant to check the broker NS. We really should check both to be complete.

sridhargaddam (approved):

Makefile
```diff
@@ -53,9 +53,9 @@ endif
 
 override CLUSTERS_ARGS += $(CLUSTER_SETTINGS_FLAG)
 override DEPLOY_ARGS += $(CLUSTER_SETTINGS_FLAG)
-override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG)
+override E2E_ARGS += $(CLUSTER_SETTINGS_FLAG) cluster1 cluster2
 export DEPLOY_ARGS
-override UNIT_TEST_ARGS += cmd internal/cli internal/env internal/log
+override UNIT_TEST_ARGS += test internal/env
```
dfarrell07, line 58: These `Makefile` and `e2e.sh` changes are intentional/necessary? They don't seem to be covered by the commit message, or at least it's not clear to me.

tpantelis: They're necessary. Adding "test" to UNIT_TEST_ARGS excludes the E2E tests from being executed by `make unit`. We do this in the other repos. I removed `internal/cli` and `internal/log` from exclusion b/c there are no unit tests in those dirs anyway - I'm unclear why they were added here in first place.

scripts/kind-e2e/e2e.sh
```diff
@@ -69,5 +69,7 @@ ${DAPPER_SOURCE}/bin/subctl benchmark throughput --intra-cluster ${KUBECONFIGS_D
 
 ${DAPPER_SOURCE}/bin/subctl benchmark throughput --verbose ${KUBECONFIGS_DIR}/kind-config-cluster1 ${KUBECONFIGS_DIR}/kind-config-cluster2
 
+${SCRIPTS_DIR}/e2e.sh "$@"
```
tpantelis, line 72: @dfarrell07 This is necessary so the new Go E2E tests in _test/e2e_ are executed via `make e2e`.

dfarrell07: @tpantelis Thanks, that makes a lot of sense. I guess I'm mostly just surprised this didn't exist already.

Makefile
dfarrell07: That makes sense, thanks @tpantelis.

> I'm unclear why they were added here in first place

That was my main concern, didn't want to lose something there. Thanks for clarifying!

controllers/uninstall/uninstall.go
```diff
@@ -38,12 +38,23 @@ const (
 	ContainerEnvVar       = "SUBMARINER_UNINSTALL"
 )
 
+type stateType int
+
+const (
+	deleteComponent stateType = iota
```
dfarrell07, line 44: Nifty! If anyone else is wondering about `iota`: https://golangbyexample.com/iota-in-golang/

test/e2e/cleanup/submariner.go
```diff
@@ -0,0 +1,330 @@
...
+	pods map[string]*podInfo
+}
+
+func startPodMonitor(name string, stopCh <-chan struct{}) *podMonitor {
+	m := &podMonitor{
+		name: name,
+		pods: map[string]*podInfo{},
+	}
+
+	w, err := watcher.New(&watcher.Config{
```
dfarrell07, line 222: This pod monitor watcher pattern is interesting, might be a good topic for a code teaching session.

dfarrell07 (approved):
LGTM

controllers/uninstall/uninstall.go
tpantelis: That's the Go way of creating (sort of) enums.

scripts/kind-e2e/e2e.sh
tpantelis: These are the first Go E2E tests in the repo.

PR #1873, opened 2022-02-23
Added command to unexport services

### LINKED ISSUE #1830: Support subctl unexport

Per a [discussion on Slack](https://kubernetes.slack.com/archives/C010RJV694M/p1642789351043100) with newish-to-SubM devs, it seems like folks rightly-expect subctl to have commands to undo actions it takes. For example, `subctl unexport service`. This can be achieved with `kubectl delete serviceexport`, but that's not a very clear UX.

Another idea is to add a `dry run` mode.

**Why is this needed**:

For better UX.

mkolesnik:
This should be more intuitive to users, instead of having them go delete
the `ServiceExport` CR.

Resolves: #1830

### DIFF
--- a/pkg/subctl/cmd/export.go
+++ b/pkg/subctl/cmd/export.go
@@ -48,18 +48,13 @@ var (
 		PreRunE: restConfigProducer.CheckVersionMismatch,
 		Run:     exportService,
 	}
-	serviceNamespace string
 )
 
 func init() {
 	restConfigProducer.AddKubeConfigFlag(exportCmd)
-	addServiceExportFlags(exportServiceCmd)
+	addNamespaceFlag(exportServiceCmd)
 	exportCmd.AddCommand(exportServiceCmd)
 	rootCmd.AddCommand(exportCmd)
-}
-
-func addServiceExportFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&serviceNamespace, "namespace", "n", "", "namespace of the service to be exported")
 }
 
 func exportService(cmd *cobra.Command, args []string) {
@@ -77,9 +72,9 @@ func exportService(cmd *cobra.Command, args []string) {
 
 	utils.ExitOnError("Error creating RestMapper for ServiceExport", err)
 
-	if serviceNamespace == "" {
-		if serviceNamespace, _, err = clientConfig.Namespace(); err != nil {
-			serviceNamespace = "default"
+	if namespace == "" {
+		if namespace, _, err = clientConfig.Namespace(); err != nil {
+			namespace = "default"
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/subctl/cmd/unexport.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/submariner-io/submariner-operator/pkg/subctl/cmd/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	mcsclient "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1"
+)
+
+var (
+	unexportCmd = &cobra.Command{
+		Use:   "unexport",
+		Short: "Stop a resource from being exported to other clusters",
+		Long:  "This command stops exporting a resource so it isn't accessible to other clusters any more",
```
tpantelis:
```suggestion
		Long:  "This command stops exporting a resource so it is no longer accessible to other clusters",
```

dfarrell07 (approved):
LGTM

tpantelis (dismissed):
One minor thing. @Jaanki We'll have to remember to move this command to the new structure as well. Do we have an epic in which to note this?

aswinsuryan (approved):

pkg/subctl/cmd/unexport.go
```diff
@@ -0,0 +1,85 @@
...
+	client, err := mcsclient.NewForConfig(restConfig)
+	utils.ExitOnError("Error connecting to the target cluster", err)
+
+	if namespace == "" {
+		if namespace, _, err = clientConfig.Namespace(); err != nil {
+			namespace = "default"
+		}
+	}
+
+	err = client.ServiceExports(namespace).Delete(context.TODO(), args[0], metav1.DeleteOptions{})
```
vthapar, line 73: We should ignore IsNotFound() error in case export is already deleted. Or return a non-error msg that ServiceExport was not found. In most unexport cases users won't care if it was already deleted but throwing error may cause problems for their cleanup scripts.

vthapar (changes requested):

Jaanki:
> One minor thing. @Jaanki We'll have to remember to move this command to the new structure as well. Do we have an epic in which to note this?

Here is the issue https://github.com/submariner-io/submariner-operator/issues/1329. Added `unexprt` subitem under item 2 `Move all cobra related stuff to cmd/subctl, Logic called by subctl inside pkg/cmd/subctl for below commands`

pkg/subctl/cmd/unexport.go
mkolesnik: I was following `kubectl` in this regard, as it does throw an error when attempting to delete a resource which isn't found.

In case of scripts an error can be easily avioded.

If we really want to simplify we can add `--ignore-not-found[=false]` flag as exists in `kubectl delete` [1] but I think it's not necessary at this point and can always be added in the future.

[1] https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#delete

pkg/subctl/cmd/unexport.go
Jaanki: along with what @tpantelis suggested,
```suggestion
		Long:  "This command stops exporting a resource so that it isn't accessible to other clusters any more",
```
as per https://learningenglish.voanews.com/a/what-s-the-difference-so-and-so-that-/4627372.html

pkg/subctl/cmd/unexport.go
```diff
@@ -0,0 +1,85 @@
...
+func init() {
+	restConfigProducer.AddKubeContextFlag(unexportCmd)
+	addNamespaceFlag(unexportServiceCmd)
+	unexportCmd.AddCommand(unexportServiceCmd)
+	rootCmd.AddCommand(unexportCmd)
+}
+
+func unexportService(cmd *cobra.Command, args []string) {
+	err := validateUnexportArguments(args)
+	utils.ExitOnError("Insufficient arguments", err)
```
Jaanki: In case you respin, I would recommend using `exit.OnErrorWithMessage` because that is replacing `utils.Exit*` functions.

pkg/subctl/cmd/export.go
```diff
@@ -77,9 +72,9 @@ func exportService(cmd *cobra.Command, args []string) {
 
 	utils.ExitOnError("Error creating RestMapper for ServiceExport", err)
 
-	if serviceNamespace == "" {
-		if serviceNamespace, _, err = clientConfig.Namespace(); err != nil {
-			serviceNamespace = "default"
+	if namespace == "" {
```
Jaanki, line 75: Could also make similar changes to https://github.com/submariner-io/submariner-operator/blob/devel/cmd/subctl/export.go please.

Jaanki (changes requested):

pkg/subctl/cmd/unexport.go
mkolesnik: Is this merged?

mkolesnik: OK found it but the location is weird IMHO, nevertheless I'll update the usages in this file.

You should add a deprecation warning on the utils function though if it's being replaced

tpantelis: We could add deprecation but the `utils` functions will be removed once the restructuring is complete next release.  Even if you don't use the `exit` functions now, they'll be converted once this code is moved under `cmd`.

pkg/subctl/cmd/export.go
tpantelis: We can do that next release when we resume the restructuring.

vthapar (approved):

Jaanki (approved):

sridhargaddam (approved):

sridhargaddam:
@tpantelis PTAL

sridhargaddam:
> @tpantelis PTAL

I see that tom's comments are addressed and also since he is on PTO, go ahead and merging this PR.

PR #1879, opened 2022-02-25
Fix new linter errors from golangci-lint 1.44.2

dfarrell07:

### DIFF
--- a/controllers/submariner/daemonsets.go
+++ b/controllers/submariner/daemonsets.go
@@ -39,8 +39,7 @@ func updateDaemonSetStatus(ctx context.Context, clnt client.Reader, daemonSet *a
 		status.Status = &daemonSet.Status
 		if status.LastResourceVersion != daemonSet.ObjectMeta.ResourceVersion {
 			// The daemonset has changed, check its containers
-			mismatchedContainerImages, nonReadyContainerStates, err :=
-				checkDaemonSetContainers(ctx, clnt, daemonSet, namespace)
+			mismatchedContainerImages, nonReadyContainerStates, err := checkDaemonSetContainers(ctx, clnt, daemonSet, namespace)
 			if err != nil {
 				return err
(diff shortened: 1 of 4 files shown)


dfarrell07:
There are two more issues flagged that I'm not sure how/if we want to fix:

```
pkg/subctl/cmd/diagnose/cni.go:214:17: dynamicFmtString: use errors.New(message) or fmt.Errorf("%s", message) instead (gocritic)
		return false, fmt.Errorf(message)
		              ^
pkg/subctl/cmd/diagnose/firewall.go:111:3: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
		defer sPod.Delete()
		^
```

skitt (approved):

nyechiel (approved):

dfarrell07:
@nyechiel There are still issues flagged by the new version that I wanted to fix here, per https://github.com/submariner-io/submariner-operator/pull/1879#issuecomment-1051356125.

I had this as non-draft to get review, and In Progress because there was still work to be done. That's a valid PR state, we should be careful not to advance PRs past it unless they are ready.

dfarrell07:
> There are two more issues flagged that I'm not sure how/if we want to fix:
> 
> ```
> pkg/subctl/cmd/diagnose/cni.go:214:17: dynamicFmtString: use errors.New(message) or fmt.Errorf("%s", message) instead (gocritic)
> 		return false, fmt.Errorf(message)
> 		              ^
> pkg/subctl/cmd/diagnose/firewall.go:111:3: deferInLoop: Possible resource leak, 'defer' is called in the 'for' loop (gocritic)
> 		defer sPod.Delete()
> 		^
> ```

This is being tracked on https://github.com/submariner-io/shipyard/pull/734#issuecomment-1056072466, so moving this to Done now.

PR #1899, opened 2022-03-04
Fix extraneous uninstall pods present after cleanup complete

tpantelis:
After the finalizer is removed, it can take a little time before K8s deletes the object. In the mean time, the controller may get triggered again in which case it will observe the `DeletionTimeStamp` non-zero and re-create the uninstall daemon sets. To avoid this, elide running cleanup if the finalizer isn't present.

Also increased the requeue interval to 500 ms.

Depends on https://github.com/submariner-io/admiral/pull/327

### DIFF
--- a/controllers/servicediscovery/cleanup.go
+++ b/controllers/servicediscovery/cleanup.go
@@ -40,6 +40,10 @@ import (
 )
 
 func (r *Reconciler) doCleanup(ctx context.Context, instance *operatorv1alpha1.ServiceDiscovery) (reconcile.Result, error) {
+	if !finalizer.IsPresent(instance, constants.CleanupFinalizer) {
+		return reconcile.Result{}, nil
+	}
+
 	if !uninstall.IsSupportedForVersion(instance.Spec.Version) {
 		log.Info("Deleting ServiceDiscovery version does not support uninstall", "version", instance.Spec.Version)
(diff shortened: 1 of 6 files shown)


sridhargaddam:
> Depends on [submariner-io/admiral#327](https://github.com/submariner-io/admiral/pull/327)

It says "Depends on" but i dont see that its using the new API from Admiral.

skitt:
> > Depends on [submariner-io/admiral#327](https://github.com/submariner-io/admiral/pull/327)
> 
> It says "Depends on" but i dont see that its using the new API from Admiral.

I imagine Tom wants to rework this to use the Admiral utility (as mentioned in the `TODO`s).

tpantelis:
> > > Depends on [submariner-io/admiral#327](https://github.com/submariner-io/admiral/pull/327)
> > 
> > 
> > It says "Depends on" but i dont see that its using the new API from Admiral.
> 
> I imagine Tom wants to rework this to use the Admiral utility (as mentioned in the `TODO`s).

Yes and done.

sridhargaddam (approved):
Thanks for pushing the PR @tpantelis.

I verified the use-case switching the cableDrivers and sometimes re-running with the same cableDriver and do not see the issue anymore :)

skitt (approved):

skitt:
Backported in #1912.

PR #1957, opened 2022-03-16
Fix valid ClusterId check

vthapar:
Current check for validating ClusterId allows `.` in clustername.
Instead it should follow Label Names validations as defined in
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

Fixes github.com/submariner-io/lighthouse/issues/707


internal/cluster/cluster.go
```diff
@@ -25,9 +25,9 @@ import (
 
 func IsValidID(clusterID string) error {
 	// Make sure the clusterid is a valid DNS-1123 string
-	if match, _ := regexp.MatchString("^[a-z0-9][a-z0-9.-]*[a-z0-9]$", clusterID); !match {
+	if match, _ := regexp.MatchString("^[a-z0-9][a-z0-9-]*[a-z0-9]$", clusterID); !match {
```
mkolesnik, line 28: Is suggest adding a limit on the length as well, as we discussed https://github.com/submariner-io/lighthouse/issues/707#issuecomment-1068888749

mkolesnik (approved):
We should probably have added test but alas there aren't any tests for subctl :(

sridhargaddam (approved):

tpantelis (approved):

Jaanki (approved):

internal/cluster/cluster.go
skitt, line 28: Can’t we use the same check as in https://github.com/submariner-io/lighthouse/pull/709?

skitt:
> We should probably have added test but alas there aren't any tests for subctl :(

https://github.com/submariner-io/submariner-operator/pull/1958

internal/cluster/cluster.go
mkolesnik: If we do (which we probably should), the main issue would be that multiple errors are returned and would need to be converted into one, where there are two separate checks - this one and the one after that.

Some flattening of the error would be required, as the `survey` expects a single `Error`

mkolesnik:
Ideally we should backport this to the next point release 0.12.1

internal/cluster/cluster.go
skitt: That’s exactly what submariner-io/lighthouse#709 does. See #1958 for the operator version.

skitt:
Merged with #1958.

PR #1959, opened 2022-03-16
Update all GHAs to their latest releases

dfarrell07:


skitt (approved):
Given that you’re taking care of all this, would it be possible to document the specific steps you’re taking? I’ve also been wondering if it would be worth having a fake (always skipped) job whose only purpose is to list our GHA dependencies using tags rather than hashes, so that dependabot would help us by at least letting us know when a new version of a GHA dependency is available...

tpantelis (approved):

PR #1968, opened 2022-03-21
Revert "Drop deprecated --colorcodes option"

skitt:
--colorcodes is used in the upgrade tests, which means we can't remove
it in 0.13. The tests have been fixed, so this will be removable in
0.14.

This reverts commit be8c8e3164d6201ef7d7361c3027e5af0402b174.

### DIFF
--- a/cmd/subctl/join.go
+++ b/cmd/subctl/join.go
@@ -38,8 +38,9 @@ import (
 )
 
 var (
-	joinFlags    join.Options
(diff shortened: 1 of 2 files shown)


skitt:
I haven’t figured out _why_ `--colorcodes` is still being used in the upgrade tests. It shouldn’t be...

skitt:
This doesn’t seem to be necessary.

PR #1980, opened 2022-03-24
Move uninstall command to new subctl implementation

Jaanki:


tpantelis (approved):

cmd/subctl/uninstall.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/submariner-io/submariner-operator/internal/constants"
+	"github.com/submariner-io/submariner-operator/internal/exit"
+	"github.com/submariner-io/submariner-operator/pkg/client"
+	"github.com/submariner-io/submariner-operator/pkg/uninstall"
+)
+
+var (
+	noPrompt  bool
+	namespace string
+)
```
tpantelis: To keep these separate from other command vars, let's encapsulate them into a struct:

```suggestion
var uninstallOptions struct {
	noPrompt bool
	namespace string
}
```

tpantelis (changes requested):

cmd/subctl/uninstall.go
Jaanki: Done

tpantelis (approved):

dfarrell07:
If this "moves" the command, shouldn't it be removed from where it was before? Is that planned for later?

dfarrell07 (approved):

tpantelis:
> If this "moves" the command, shouldn't it be removed from where it was before? Is that planned for later?

yeah it will be once the rest of the commands are "moved".

PR #1983, opened 2022-03-24
Move unexport to new subctl implementation

### LINKED ISSUE #1881: Make changes in export command to match the recent changes
Could also make similar changes to https://github.com/submariner-io/submariner-operator/blob/devel/cmd/subctl/export.go please.

_Originally posted by @Jaanki in https://github.com/submariner-io/submariner-operator/pull/1873#r813689518_

Jaanki:

### DIFF
--- a/cmd/subctl/unexport.go
+++ b/cmd/subctl/unexport.go
@@ -1,0 +1,83 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subctl
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/submariner-io/submariner-operator/internal/cli"
+	"github.com/submariner-io/submariner-operator/internal/exit"
+	"github.com/submariner-io/submariner-operator/pkg/service"
+	mcsclient "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1"
+)
+
+var (
+	unexportOptions struct {
+		namespace string
+	}
+	unexportCmd = &cobra.Command{
+		Use:   "unexport",
+		Short: "Stop a resource from being exported to other clusters",
+		Long:  "This command stops exporting a resource so that it's no longer accessible to other clusters",
+	}
+	unexportServiceCmd = &cobra.Command{
+		Use:   "service <serviceName>",
+		Short: "Stop a Service from being exported to other clusters",
+		Long: "This command removes the ServiceExport resource with the given name which in turn stops the Service " +
+			"of the same name from being exported to other clusters",
+		PreRunE: restConfigProducer.CheckVersionMismatch,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := validateUnexportArguments(args)
+			exit.OnErrorWithMessage(err, "Insufficient arguments")
+
+			status := cli.NewReporter()
+
+			config, err := restConfigProducer.ForCluster()
+			exit.OnError(status.Error(err, "Error creating REST config"))
+
+			clientConfig := restConfigProducer.ClientConfig()
+			if unexportOptions.namespace == "" {
+				if unexportOptions.namespace, _, err = clientConfig.Namespace(); err != nil {
+					unexportOptions.namespace = "default"
+				}
+			}
+
+			client, err := mcsclient.NewForConfig(config.Config)
+			exit.OnError(status.Error(err, "Error creating client"))
+
+			err = service.Unexport(client, unexportOptions.namespace, args[0], status)
+			exit.OnError(err)
+		},
+	}
+)
+
+func init() {
+	restConfigProducer.AddKubeContextFlag(unexportCmd)
+	unexportServiceCmd.Flags().StringVarP(&unexportOptions.namespace, "namespace", "n", "", "namespace of the service to be unexported")
+	unexportCmd.AddCommand(unexportServiceCmd)
+	rootCmd.AddCommand(unexportCmd)
+}
+
+func validateUnexportArguments(args []string) error {
+	if len(args) == 0 || args[0] == "" {
+		return errors.New("name of the Service to be removed must be specified")
+	}
+
+	return nil
+}
--- a/pkg/service/unexport.go
+++ b/pkg/service/unexport.go
@@ -1,0 +1,43 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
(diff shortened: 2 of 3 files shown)


pkg/service/export.go
```diff
@@ -30,16 +30,16 @@ import (
 	mcsv1a1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
-func Export(clientProducer client.Producer, serviceNamespace, svcName string, status reporter.Interface) error {
```
skitt: This should be left as-is — `serviceNamespace` is more explicit than `namespace`.

cmd/subctl/root.go
```diff
@@ -35,7 +35,10 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var restConfigProducer = restconfig.NewProducer()
+var (
+	restConfigProducer = restconfig.NewProducer()
+	namespace          string
```
skitt: This bothers me — we’ve had bugs in the past from global variables which required setup, where code was added which used the variables in code paths where they weren’t set up. I’d rather have a namespace variable specific to each command which can use a namespace (with a shared “namespace flag” function to ensure the help text is consistent); but that’s made complicated by the shared package.

skitt (changes requested):

pkg/service/export.go
Jaanki: This was changed to be in line with the similar changes made in `pkg/subctl/cmd/export.go` in commit https://github.com/submariner-io/submariner-operator/commit/efcc042311cceddb6f1652f58171100d184cfafa

cmd/subctl/unexport.go
```diff
@@ -0,0 +1,80 @@
...
+
+			clientConfig := restConfigProducer.ClientConfig()
+			if namespace == "" {
+				if namespace, _, err = clientConfig.Namespace(); err != nil {
+					namespace = "default"
+				}
+			}
+
+			client, err := mcsclient.NewForConfig(config)
+			exit.OnErrorWithMessage(err, "Error connecting to the target cluster")
```
tpantelis:
```suggestion
                         exit.OnError(status.Error(err, "Error creating client"))
```

pkg/service/export.go
tpantelis: What similar change specifically? That commit didn't touch this file. I agree with @skitt plus this change is extraneous to the purpose of this PR.

pkg/service/unexport.go
```diff
@@ -0,0 +1,44 @@
...
+	mcs "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1"
+)
+
+func Unexport(client *mcs.MulticlusterV1alpha1Client, namespace, svcName string, status reporter.Interface) error {
+	err := client.ServiceExports(namespace).Delete(context.TODO(), svcName, metav1.DeleteOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			status.Success("Service %s not found", svcName)
+			return nil
+		}
```
tpantelis: I don't think this should be reported as success. 

```suggestion
		if k8serrors.IsNotFound(err) {
			status.Failure("Service %s/%s was not previously exported", namespace, svcName)
			return err
		}
```

pkg/service/unexport.go
```diff
@@ -0,0 +1,44 @@
...
+
+		return status.Error(err, "Failed to unexport Service")
+	}
+
+	status.Success("Service unexported successfully")
```
tpantelis:
```suggestion
	status.Success("Service successfully unexported")
```

tpantelis (changes requested):

cmd/subctl/root.go
tpantelis: We can go a step further to isolate global vars by encapsulating them in a struct specific to each command, ie

```
var unexportOptions struct {
    namespace string
}
```

tpantelis (changes requested):

pkg/service/export.go
Jaanki: `serviceNamespace` was changed to `namespace` at https://github.com/submariner-io/submariner-operator/commit/efcc042311cceddb6f1652f58171100d184cfafa#diff-b66e1481a17968a8df33b0001a89072251acecd0f02b1028c262ab39e3d9daf1R75 in file `pkg/subctl/cmd/export.go`. `pkg/service/export.go` is replica of this file and should have same content.

tpantelis: That commit changed a global var to make it reusable. This is a separate function with a local function param. The former code will eventually go away. I don't follow your logic re: the equivalency where they need to be the same. Plus this PR is for changing "unexport" - this change is totally extraneous to that purpose.

cmd/subctl/root.go
Jaanki: Or we can name the vars as `exportNamespace` and `unexportNamespace`?

Jaanki: @tpantelis @skitt WDYT?

skitt: The whole point is to isolate the variables, so we avoid polluting the global namespace; so an `unexportOptions` struct is better than explicitly-named global variables.

cmd/subctl/root.go
```diff
@@ -84,3 +84,7 @@ func compareFiles(file1, file2 string) (bool, error) {
 
 	return bytes.Equal(first, second), nil
 }
+
+func addNamespaceFlag(cmd *cobra.Command, namespace string) {
```
tpantelis: The caller needs to pass the address of `namespace`. However I don't it makes sense to generalize the option name and description, ie we lose specific useful information, eg "namespace of the service to be exported" is better than "namespace to use".  We could pass the name/description but I don't think it's worth it since this function just wraps one lint of code. Therefore, just inline `cmd.Flags().StringVarP(...)`.

tpantelis (changes requested):

tpantelis (approved):

pkg/service/unexport.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/submariner-io/admiral/pkg/reporter"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	mcs "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1"
+)
+
+func Unexport(client *mcs.MulticlusterV1alpha1Client, namespace, svcName string, status reporter.Interface) error {
+	defer status.End()
```
tpantelis: This is only necessary if there's a corresponding `Start`.

pkg/service/unexport.go
```diff
@@ -0,0 +1,47 @@
...
+
+	err := client.ServiceExports(namespace).Delete(context.TODO(), svcName, metav1.DeleteOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			status.Failure("Service %s/%s was not previously exported", namespace, svcName)
+			return errors.Wrap(err)
```
tpantelis: This can be simplified.

```suggestion
			return status.Error(err, "Service %s/%s was not previously exported", namespace, svcName)
```

tpantelis (changes requested):

tpantelis (approved):

skitt (approved):

dfarrell07:
Really weird MD link check error, happed again after re-run:

```
fatal: unsafe repository ('/github/workspace' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /github/workspace
```

PR #1989, opened 2022-03-28
Drop IKE port configuration

skitt:
Submariner no longer uses this configuration setting, drop it from the
operator (but keep it in the CRD to avoid breaking old clients setting
it).

### DIFF
--- a/cmd/subctl/join.go
+++ b/cmd/subctl/join.go
@@ -40,6 +40,9 @@ import (
 var (
 	joinFlags    join.Options
 	labelGateway bool
+
+	// Deprecated: will be removed in 0.14.
+	ignoredIkePort int
 )
 
 var joinCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


vthapar (approved):

cmd/subctl/join.go
```diff
@@ -38,8 +38,9 @@ import (
 )
 
 var (
-	joinFlags    join.Options
-	labelGateway bool
+	joinFlags      join.Options
+	ignoredIkePort int
```
mkolesnik, line 45: Should we add a todo comment here to remove it?

mkolesnik (commented):
Generally we should declare when this is going to be removed, we don;t want to keep it around forever right?

cmd/subctl/join.go
tpantelis: The [recommended way](https://go.dev/blog/godoc) was is to add a doc comment that begins with “Deprecated:”

skitt:
> Generally we should declare when this is going to be removed, we don;t want to keep it around forever right?

Indeed, I’ve added two deprecation comments, and we will have to mention this in the release notes (thanks for the label).

mkolesnik (approved):

tpantelis (approved):

dfarrell07:
```
cmd/subctl/join.go:44:40: Comment should end in a period (godot)
	// Deprecated: will be removed in 0.14
	                                      ^
pkg/subctl/cmd/join.go:81:40: Comment should end in a period (godot)
	// Deprecated: will be removed in 0.14
	                                      ^
```

dfarrell07 (approved):

PR #1993, opened 2022-03-31
Use multi-active-gw dependencies

skitt:
This ensures that GHAs, base branches, and images are all taken from
the multi-active-gw branches.

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -10,5 +10,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check that the PR targets multi-active-gw
-        if: ${{ github.base_ref != 'feature/multi-active-gw' }}
+        if: ${{ github.base_ref != 'feature-multi-active-gw' }}
         run: exit 1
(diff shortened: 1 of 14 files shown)


Billy99 (approved):
Thanks Stephen. This is a lot more work than I thought it would be to add a feature branch. We appreciate it.

skitt:
> Thanks Stephen. This is a lot more work than I thought it would be to add a feature branch. We appreciate it.

You’re welcome! Our build setup is a bit too complex for its own good ;-). Next step: document all this, and perhaps improve some parts of it...

api/submariner/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
dfarrell07: This is intentional?

dfarrell07 (approved):

api/submariner/v1alpha1/zz_generated.deepcopy.go
skitt: Ah no, it gets added automatically when building with Go 1.17. I’ve removed it.

PR #1994, opened 2022-03-31
Refactor pkg/subctl/cmd/diagnose functionality to pkg/diagnose

tpantelis:
See individual commits for details. It's a lot of commits but each is relatively small. There's basically 3 commits per sub-command  functionality:

 - Copy the related file to _pkg/diagnose_ as is
 - Refactor the code to use the new `cluster.Info` and `reporter.Interface` APIs
 - Modify the original code to call the new function


pkg/subctl/cmd/diagnose/cni.go
```diff
@@ -56,168 +41,21 @@ func init() {
...
-	}
-
-	checkGatewaySubnets(gateways, ippools, status)
-
-	result := status.ResultFromMessages()
-	status.EndWith(result)
-
-	return result != cli.Failure
+func checkCNIConfig(c *cmd.Cluster) bool {
+	return diagnose.CNIConfig(clusterInfoFrom(c), cli.NewStatus())
```
Jaanki, line 45: I would prefer using `cli.NewReporter()` because thats what we are using a reporter (which in turn uses status) and also for consistency with other commands. Applies throughout.

pkg/subctl/cmd/diagnose/firewall_tunnel.go
```diff
@@ -90,87 +83,11 @@ func validateTunnelConfigAcrossClusters(localCfg, remoteCfg *rest.Config) bool {
...
-
-	return true
+	return diagnose.TunnelConfigAcrossClusters(clusterInfoFrom(localCluster), clusterInfoFrom(remoteCluster), diagnose.FirewallOptions{
+		ValidationTimeout: validationTimeout,
+		VerboseOutput:     verboseOutput,
+		PodNamespace:      podNamespace,
+	}, cli.NewStatus())
 }
 
 func newCluster(cfg *rest.Config) *cmd.Cluster {
```
Jaanki, line 93: Use `exit.WithMessage` instead of `utils.ExitWithErrorMsg`.

PS. The comment is for L179 and L183 but I am not able to add comment there (I don't know why).

Jaanki (changes requested):
Also add `diagnose.go` with all cobra stuff to `cmd/subctl`.

tpantelis:
> Also add `diagnose.go` with all cobra stuff to `cmd/subctl`.

I have that queued up for another PR. This PR is large enough.

pkg/subctl/cmd/diagnose/cni.go
tpantelis: yeah we can but it doesn't matter as this code will be going away anyway.

pkg/subctl/cmd/diagnose/firewall_tunnel.go
tpantelis: yeah I didn't touch this code so changing it would be out-of-scope for this PR. But again it doesn't matter since this code will go away.

pkg/subctl/cmd/diagnose/cni.go
Jaanki: well, going away yes. But I guess you would be copying this portion to `cmd/subctl`. And you be changing then. I would expect these changes made in this PR (as per the PR description) and simply copy/paste to `cmd/subctl`.

pkg/subctl/cmd/diagnose/firewall_tunnel.go
Jaanki: Ditto as above.

pkg/subctl/cmd/diagnose/cni.go
tpantelis: It's not a copy/paste. I've already done the `cmd/subctl` work. I don't need this changed.

Jaanki (approved):

pkg/subctl/cmd/diagnose/firewall_tunnel.go
tpantelis: This function does not pertain to the new code under `cmd/subctl.`

dfarrell07 (approved):

PR #2005, opened 2022-04-06
Use unique name for all-K8s E2E job

dfarrell07:
The GitHub required job configuration seems to only filter based on the
names of jobs, not the names of workflows and jobs. This now causes the
E2E All K8s job to be required, which would require running way too many
jobs on each PR.

There seems to have been some change to cause this new behavior, as the
same GHA configuration didn't used to result in the all-K8s job being
required.

### DIFF
--- a/.github/workflows/e2e-all-k8s.yml
+++ b/.github/workflows/e2e-all-k8s.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   e2e:
-    name: E2E
+    name: E2E All K8s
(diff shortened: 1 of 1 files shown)


dfarrell07:
Yeah it's no longer required. :heavy_check_mark:

dfarrell07:
We need to be able to pin the version of GitHub we use :wink:

tpantelis (approved):

Jaanki (approved):

PR #2007, opened 2022-04-07
Refactor pkg/subctl/cmd/gather -> internal/gather

Jaanki:
This PR does the following:
1. Adds `config` to `cluster.Info`
2. Uses execute.OnMultiCluster functionality
3. Uses constants defined in internal/constants.go

Related to https://github.com/submariner-io/submariner-operator/issues/1329


tpantelis (commented):
So it looks like you copied code to _internal/gather_  but it's not being used.

Jaanki:
> So it looks like you copied code to _internal/gather_ but it's not being used.

Yes. e2e is failing for some reason so didn't push more commits.

pkg/cluster/info.go
```diff
@@ -28,17 +28,20 @@ import (
...
+	RestConfig     *rest.Config
 	ClientProducer client.Producer
 	Submariner     *v1alpha1.Submariner
 }
 
-func NewInfo(clusterName string, clientProducer client.Producer) (*Info, error) {
+func NewInfo(clusterName string, clientProducer client.Producer, config *rest.Config) (*Info, error) {
 	info := &Info{
 		Name:           clusterName,
+		RestConfig:     config,
```
Jaanki, line 45: `RestConfig` is needed by `gatherCNIResources` -> `execCmdInBash` -> `pods.ExecWithOprions` leading to https://github.com/submariner-io/submariner-operator/blob/devel/internal/pods/exec.go#L90

tpantelis (changes requested):
To maintain git history, we should first move the files to _internal/gather_.

dfarrell07:
> To maintain git history, we should first move the files to _internal/gather_.

We have a pointer for this guidance at https://submariner.io/development/code-review/#rename-and-edit-in-separate-commits FYI. I think it was created because of very similar feedback on similar refactoring a few months ago.

Jaanki:
> To maintain git history, we should first move the files to _internal/gather_.

`git mv` moves the files from `pkg/subctl/cmd/gather/.` to `internal/gather`. I want the files to be **copied** and **not moved.**

tpantelis:
> git mv moves the files from pkg/subctl/cmd/gather/. to internal/gather. I want the files to be copied and not moved.

Why? Copying loses the git history. 

```
git log -- internal/gather/cni.go

commit d11b28efd3454e88ec8a69826270366bb4f24133
Author: Janki Chhatbar <jchhatba@redhat.com>
Date:   Fri Apr 22 11:45:28 2022 +0530

    Copy pkg/subctl/cmd/gather -> internal/gather
    
```

The deceiving thing tho is that the diff in the GtHub UI shows that it was renamed w/o changes.

internal/gather/gather.go
```diff
@@ -0,0 +1,270 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+type Options struct {
+	Directory            string
+	IncludeSensitiveData bool
+	ModuleFlags          map[string]bool
+	TypeFlags            map[string]bool
```
skitt: These should be string slices, not maps; that would make it easier to use them as Cobra flags directly (and we don’t care about the `bool`).

internal/gather/gather.go
```diff
@@ -0,0 +1,270 @@
...
+	Logs      = "logs"
+	Resources = "resources"
+)
+
+var ModuleFlags = map[string]bool{
+	component.Connectivity:     false,
+	component.ServiceDiscovery: false,
+	component.Broker:           false,
+	component.Operator:         false,
+}
```
skitt: This shouldn’t be necessary, the same information is effectively present in `gatherFuncs`; perhaps a `GetSupportedModules() []string` function could be used instead, or an `IsModuleSupported(module string) bool` function.

internal/gather/gather.go
```diff
@@ -0,0 +1,270 @@
...
+
+var TypeFlags = map[string]bool{
+	"logs":      false,
+	"resources": false,
+}
```
skitt: Likewise, this should be hidden behind a function, and maintained as a `[]string`.

skitt (changes requested):

internal/gather/gather.go
tpantelis: Commit https://github.com/submariner-io/submariner-operator/pull/2007/commits/0e5e8a46eeb0ac3155f9f0a85d70626850fd1c56.

internal/gather/gather.go
tpantelis: Made it a set with https://github.com/submariner-io/submariner-operator/pull/2007/commits/0e5e8a46eeb0ac3155f9f0a85d70626850fd1c56.

internal/gather/gather.go
tpantelis: Made it a set with https://github.com/submariner-io/submariner-operator/pull/2007/commits/0e5e8a46eeb0ac3155f9f0a85d70626850fd1c56

tpantelis (approved):

cmd/subctl/diagnose.go
```diff
@@ -210,7 +210,7 @@ func clusterInfoFromKubeConfig(kubeConfig string) *cluster.Info {
 	clientProducer, err := client.NewProducerFromRestConfig(config.Config)
 	exit.OnErrorWithMessage(err, fmt.Sprintf("Error creating client producer for kubeconfig %q", kubeConfig))
 
-	clusterInfo, err := cluster.NewInfo("", clientProducer)
+	clusterInfo, err := cluster.NewInfo("", clientProducer, nil)
```
skitt, line 213: This is unfortunate, but let’s not block the PR — see #2042.

skitt (approved):

cmd/subctl/diagnose.go
tpantelis: Agree.

PR #2015, opened 2022-04-14
Add new subctl diagnose command

tpantelis:
...under _cmd/subctl_ that uses _pkg/diagnose_.

Related to https://github.com/submariner-io/submariner-operator/issues/1329


Jaanki (approved):

mkolesnik (commented):
What about the code in https://github.com/submariner-io/submariner-operator/tree/devel/pkg/subctl/cmd/diagnose ?

We'll be losing all git history from there

tpantelis:
> What about the code in https://github.com/submariner-io/submariner-operator/tree/devel/pkg/subctl/cmd/diagnose ?
> 
> We'll be losing all git history from there

Much that code now resides under _pkg/diagnose_ and was already moved and restructured to a lib by another PR. This PR simply adds new code to use it.

skitt (approved):

mkolesnik:
> > What about the code in https://github.com/submariner-io/submariner-operator/tree/devel/pkg/subctl/cmd/diagnose ?
> > We'll be losing all git history from there
> 
> Much that code now resides under _pkg/diagnose_ and was already moved and restructured to a lib by another PR. This PR simply adds new code to use it.

In fact the link I put is the code after the mentioned refactor, and is basically what your'e adding here so I'm not sure why not `git mv` it here and preserve the history rather than rewrite it?

tpantelis:
> > > What about the code in https://github.com/submariner-io/submariner-operator/tree/devel/pkg/subctl/cmd/diagnose ?
> > > We'll be losing all git history from there
> > 
> > 
> > Much that code now resides under _pkg/diagnose_ and was already moved and restructured to a lib by another PR. This PR simply adds new code to use it.
> 
> In fact the link I put is the code after the mentioned refactor, and is basically what your'e adding here so I'm not sure why not `git mv` it here and preserve the history rather than rewrite it?

This code is a bit different than what's left in _pkg/subctl/cmd/diagnose_  plus the latter will be going away. There's no need to move it as is. As I said the bulk of the code that was in _pkg/subctl/cmd/diagnose_  is now in _pkg//diagnose_ - the files were first copied then modified appropriately in both places (see https://github.com/submariner-io/submariner-operator/pull/1994).  However the files weren't actually moved so the git history wasn't preserved. I supposed they could've been. We could revert that PR and redo it. @skitt @dfarrell07 do you think it's worth doing that at this point?

skitt:
> This code is a bit different than what's left in _pkg/subctl/cmd/diagnose_ plus the latter will be going away. There's no need to move it as is. As I said the bulk of the code that was in _pkg/subctl/cmd/diagnose_ is now in _pkg//diagnose_ - the files were first copied then modified appropriately in both places (see #1994). However the files weren't actually moved so the git history wasn't preserved. I supposed they could've been. We could revert that PR and redo it. @skitt @dfarrell07 do you think it's worth doing that at this point?

It’s too late for that, the whole “new `subctl`” development has involved copying code instead of moving it. It should still be possible to follow the history of changes in code that was copied by using `git blame -CCC`.

tpantelis:
> 
> It’s too late for that, the whole “new `subctl`” development has involved copying code instead of moving it. It should still be possible to follow the history of changes in code that was copied by using `git blame -CCC`.

yeah this effort has involved more than simply moving files to a new location - it also involved restructuring the code such that parts of an original file were copied to other locations. Also we have to keep the current subctl version working in the interim which meant keeping some semblance of the current code in place. Maintaining git history is a challenge in this realm. In hindsight, perhaps things could've been done a little differently to preserve at least some history.

dfarrell07 (approved):

PR #2033, opened 2022-04-28
Update Lighthouse/Shipyard to devel

dfarrell07:
See commit messages for details.

This update should fix Lighthouse's CI (nginx image breakage).

PR #2034, opened 2022-04-28
Update `lighthouse` dependency

mkolesnik:
Update the `go.mod` dependency to consume latest LH test code which uses
a simple server in the `nettest` image to test discovery.


mkolesnik:
I see this is already handled by #2033 so closing this one

PR #2036, opened 2022-05-02
Increase the E2E Full timeout to 45min

skitt:
The Lighthouse jobs are timing out with the current 30min limit. This
increases the timeout to match that used in Submariner.


Jaanki (approved):

dfarrell07 (approved):

dfarrell07:
Related: https://github.com/submariner-io/lighthouse/pull/736

skitt:
We’re hitting the timeouts on the 0.12 branch as well (see https://github.com/submariner-io/submariner-operator/actions/runs/2262447911/attempts/1), this should be backported.

mkolesnik (changes requested):
This is a red herring and wont fix the issue, I'm investigating the failures on 0.12, don't merge for now please

mkolesnik:
We should probably do https://github.com/submariner-io/enhancements/issues/45 instead of this

PR #2039, opened 2022-05-02
Point to submariner@feature-multi-active-gw for updated Endpoint CRD

Billy99:
Override "Point to github.com/astycos/submariner for updated Endpoint CRD"

This overides the temporarily patch which pointed to astoycos version:
 github.com/astycos/submariner@AS-multi-active-gw-cgeip-controller
This now points to:
 github.com/submariner-io/submariner@feature-multi-active-gw

The Submariner Endpoint CRD was updated to include the AllocatedIPs from the
ClusterGlobalEgressIPs. The Endpoint CRD struct is define in the Submariner
repo, so need to point to a version that contains the change.

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,4 +71,5 @@ replace (
 	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.9.0
 )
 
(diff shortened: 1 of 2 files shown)


astoycos:
Just waiting to see how CI works (it should run alright now)

astoycos (approved):

PR #2052, opened 2022-05-05
Bump Lighthouse, drop the CoreDNS dependencies

skitt:
This drops the dependency on CoreDNS and allows us to remove the
pinning on most Kubernetes dependencies. We still need to pin
client-go, because prometheus-operator pulls in v12.


tpantelis:
We need to downgrade `github.com/go-logr/logr` to 0.4.0.

go.mod
```diff
@@ -62,11 +65,6 @@ require (
 // Pinned to kubernetes-1.21.11
 replace (
 	github.com/go-logr/logr => github.com/go-logr/logr v0.4.0
```
tpantelis: I think we can specify this directly on L13 instead of replace?

go.mod
```diff
@@ -62,11 +65,6 @@ require (
...
-	k8s.io/api => k8s.io/api v0.21.11
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.21.11
-	k8s.io/apimachinery => k8s.io/apimachinery v0.21.11
 	k8s.io/client-go => k8s.io/client-go v0.21.11
-	k8s.io/cloud-provider => k8s.io/cloud-provider v0.21.11
 	k8s.io/klog/v2 => k8s.io/klog/v2 v2.9.0
```
tpantelis, line 68: Do we still need this?

skitt: I think so, but let’s try without...

skitt: Yup, we do still need it.

dfarrell07 (approved):

tpantelis (approved):

PR #2053, opened 2022-05-05
Add new subctl aws cloud commands

tpantelis:


skitt (approved):

dfarrell07 (approved):

Jaanki (changes requested):
Also I am thinking if there should be separate packages for `cleanup` and `prepare` under `cmd/subctl/cloud` like how it is currently implemented.

cmd/subctl/aws.go
```diff
@@ -0,0 +1,80 @@
...
+		},
+	}
+
+	awsCleanupCmd = &cobra.Command{
+		Use:   "aws",
+		Short: "Clean up an AWS cloud",
+		Long: "This command cleans up an OpenShift installer-provisioned infrastructure (IPI) on AWS-based" +
+			" cloud after Submariner uninstallation.",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := cleanup.AWS(&restConfigProducer, cli.NewReporter())
```
Jaanki: `cleanup` also needs `infraid` and `region` flags just like `prepare` and those value would need to be passed on.

Jaanki (changes requested):

cmd/subctl/aws.go
tpantelis: Done

tpantelis:
> Also I am thinking if there should be separate packages for `cleanup` and `prepare` under `cmd/subctl/cloud` like how it is currently implemented.

The bulk of the actual code is under `pkg/cloud` so `cmd/subctl/cloud` is basically just setting up cobra commands so I think separate packages would be overkill. Also it would require making some things public and explicitly referencing the packages in `main` so they're initialized which, again, I don't think is worth it.

pkg/subctl/cmd/cloud/cleanup/aws.go
```diff
@@ -40,7 +40,7 @@ func newAWSCleanupCommand(restConfigProducer restconfig.Producer) *cobra.Command
 		Run: func(cmd *cobra.Command, args []string) {
 			status := cli.NewReporter()
 
-			err := cleanup.AWS(&restConfigProducer, status)
+			err := cleanup.AWS(&restConfigProducer, &config, status)
```
Jaanki, line 43: The `config` struct fields `infraID, region, profile, credentialsFile, ocpMetadataFile` need to populated from the flag. And `config.GwInstanceType` to be passed as empty string.

Jaanki (changes requested):

pkg/subctl/cmd/cloud/cleanup/aws.go
tpantelis: That's the case from what I see. What do you see different?

Jaanki: I would have explicitly made `config.GwInstanceType = ""`

cmd/subctl/cloud_cleanup.go
```diff
@@ -0,0 +1,31 @@
+/*
```
Jaanki: How about this is merged with the above file `cloud.go`

cmd/subctl/cloud_prepare.go
```diff
@@ -0,0 +1,45 @@
+/*
```
Jaanki: Ditto

Jaanki (commented):
How about we have 1 file to add cloud, clean and prepare and then individual files for each platform
```
cmd/subctl
  |
  |- cloud.go -> add cloud, prepare and clean command
  |- aws.go
  |- rhos.go
```
and so on

pkg/subctl/cmd/cloud/cleanup/aws.go
tpantelis: No need - empty string is the default.

Jaanki (approved):

PR #2066, opened 2022-05-10
gather: Use libreswan as default if none specified

### LINKED ISSUE #2065: subctl gather fails to gather libreswan data
**What happened**:
`subctl gather` fails to gather libreswan data if `CableDriver` field is missing/empty in `Submariner` Spec.

If `Submariner.Spec.CableDriver` is empty, it should use libreswan as default and gather data for it.

Deploy submariner manually with `Submariner.Spec.CableDriver` field empty.

**Environment**:

subctl version: v0.12.0
Submariner version:  0.12.0
Kubernetes Server version: v1.23.5+9ce5071

vthapar:
If no cable driver is configured in SubmarinerSpec, use
`libreswan` as default when gathering data.

Fixes #2065


internal/gather/cni.go
```diff
@@ -172,7 +172,7 @@ func gatherOVNResources(info *Info, networkPlugin string) {
 
 func gatherCableDriverResources(info *Info, cableDriver string) {
 	logPodInfo(info, "cable driver data", gatewayPodLabel, func(info *Info, pod *v1.Pod) {
-		if cableDriver == libreswan {
+		if cableDriver == libreswan || cableDriver == "" { // If none specified, use libreswan as default
```
skitt, line 175: Instead of hard-coding Libreswan as the default, ask the cable drivers what the default is:

```suggestion
		if cableDriver == libreswan || (cableDriver == "" && cable.GetDefaultCableDriver() == libreswan) {
```

skitt (changes requested):

internal/gather/cni.go
vthapar: Don't think this will work. `GetDefaultCableDriver()` fetches value from a global variable that is set in submariner engine. It won't be available in subctl

skitt: Ah yes, you’re right!

skitt (approved):

dfarrell07 (approved):

tpantelis (approved):

sridhargaddam:
@vthapar I think we should backport this PR. Can you please handle it?

dfarrell07:
PR scrub agrees this should be backported.

PR #2073, opened 2022-05-11
Use == instead of Strings.Compare()

dfarrell07:
Fixes new gocritic errors in golangci-lint 1.46.


tpantelis (approved):

skitt (changes requested):
Please make the commit message describe the change, not what prompted it: “Use == instead of Strings.Compare()” (the gocritic reference can be added in the commit body).

skitt (approved):

PR #2100, opened 2022-06-06
Make `getConfigMap` function public again.

Jaanki:
The function (GetGlobalnetConfigMap -> getConfigMap) was made
private in
https://github.com/submariner-io/submariner-operator/pull/2074/commits/83530c28a3a82c883b8c20e8e4c4e294729742aa#diff-6248f1e67444360455faa328c4ae105a2502bd60cd1db127b5398a7d2af23d7cR137.
This PR makes it publoc because it is used in submariner-addon.
https://github.com/stolostron/submariner-addon/blob/main/pkg/hub/submarineragent/controller.go#L720

### DIFF
--- a/pkg/discovery/globalnet/config_map.go
+++ b/pkg/discovery/globalnet/config_map.go
@@ -134,6 +134,6 @@ func updateConfigMap(k8sClientset kubernetes.Interface, namespace string, config
 }
 
 // nolint:wrapcheck // No need to wrap here
-func getConfigMap(kubeClient kubernetes.Interface, namespace string) (*v1.ConfigMap, error) {
+func GetConfigMap(kubeClient kubernetes.Interface, namespace string) (*v1.ConfigMap, error) {
 	return kubeClient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), globalCIDRConfigMapName, metav1.GetOptions{})
 }
--- a/pkg/discovery/globalnet/config_map.go
+++ b/pkg/discovery/globalnet/config_map.go
@@ -137,3 +137,8 @@ func updateConfigMap(k8sClientset kubernetes.Interface, namespace string, config
 func GetConfigMap(kubeClient kubernetes.Interface, namespace string) (*v1.ConfigMap, error) {
 	return kubeClient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), globalCIDRConfigMapName, metav1.GetOptions{})
 }
+
+// nolint:wrapcheck // No need to wrap here
+func DeleteConfigMap(kubeClient kubernetes.Interface, namespace string) error {
+	return kubeClient.CoreV1().ConfigMaps(namespace).Delete(context.TODO(), globalCIDRConfigMapName, metav1.DeleteOptions{})
+}
--- a/pkg/discovery/globalnet/globalnet.go
+++ b/pkg/discovery/globalnet/globalnet.go
@@ -325,7 +325,7 @@ func ValidateGlobalnetConfiguration(globalnetInfo *Info, netconfig Config, statu
 }
 
 func GetGlobalNetworks(kubeClient kubernetes.Interface, brokerNamespace string) (*Info, *v1.ConfigMap, error) {
-	configMap, err := getConfigMap(kubeClient, brokerNamespace)
+	configMap, err := GetConfigMap(kubeClient, brokerNamespace)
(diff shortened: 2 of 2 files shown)


tpantelis (approved):

dfarrell07:
GitHub is telling me the commit diff you linked to "does not belong to any branch on this repository, and may belong to a fork outside of the repository." I think a6db77301990be56d1e3ad31c992a4e915787568 might be what you mean?

Jaanki:
> GitHub is telling me the commit diff you linked to "does not belong to any branch on this repository, and may belong to a fork outside of the repository." I think [a6db773](https://github.com/submariner-io/submariner-operator/commit/a6db77301990be56d1e3ad31c992a4e915787568) might be what you mean?

@dfarrell07 Are you talking about the commit link in the commit message? Its working for me. But anyways, the intention was to give context about the changes made. The commit you linked also works. https://github.com/submariner-io/submariner-operator/commit/a6db77301990be56d1e3ad31c992a4e915787568#diff-6248f1e67444360455faa328c4ae105a2502bd60cd1db127b5398a7d2af23d7cR137

dfarrell07:
This is what I mean:

commit_not_repo

pkg/discovery/globalnet/config_map.go
```diff
@@ -31,11 +31,11 @@ import (
...
-	globalCIDRConfigMapName     = "submariner-globalnet-info"
-	globalnetEnabledKey         = "globalnetEnabled"
-	clusterInfoKey              = "clusterinfo"
-	globalnetCidrRange          = "globalnetCidrRange"
-	globalnetClusterSize        = "globalnetClusterSize"
+	GlobalCIDRConfigMapName     = "submariner-globalnet-info"
+	GlobalnetEnabledKey         = "globalnetEnabled"
+	ClusterInfoKey              = "clusterinfo"
+	GlobalnetCidrRange          = "globalnetCidrRange"
+	GlobalnetClusterSize        = "globalnetClusterSize"
```
skitt: Why are we exporting these?

skitt (changes requested):

pkg/discovery/globalnet/config_map.go
tpantelis: They're used in unit tests in the [submariner-addon](https://github.com/stolostron/submariner-addon/blob/81e24d31bc6e5ded7fd7b658e43e64b5b37def3e/pkg/hub/submarinerbrokerinfo/brokerinfo_test.go#L102).

skitt: Right, I saw that (which I should have mentioned), but I don’t think it’s great to have UTs in another project relying on internals of the operator and Globalnet.

I really have two questions. (a) Do we want to have a UT for this in the addon? (b) If we do, shouldn’t we encapsulate things better in the operator for the addon? For example, add a constructor.

tpantelis: > I really have two questions. (a) Do we want to have a UT for this in the addon?

I think we want to have the addon test/verify it correctly processed/configured Globalnet. Internally it uses operator functionality that has to be setup in the tests.

> (b) If we do, shouldn’t we encapsulate things better in the operator for the addon? For example, add a constructor.

yeah I think we should be able to do something like that to avoid exposing the internal config map keys. 

@Jaanki I have cycles to look into that if you want.

Jaanki: @tpantelis sure !!

tpantelis: It turns out there's already `CreateGlobalnetConfigMap` and `NewGlobalnetConfigMap` functions that encapsulate 
 the constants. I pushed [this addon PR](https://github.com/stolostron/submariner-addon/pull/384) to use them. The only remaining reference is [here](https://github.com/stolostron/submariner-addon/blob/81e24d31bc6e5ded7fd7b658e43e64b5b37def3e/pkg/hub/submarineragent/controller_test.go#L565) and we can simply add a wrapper function:

```Go
func DeleteConfigMap(kubeClient kubernetes.Interface, namespace string) error {
	return kubeClient.CoreV1().ConfigMaps(namespace).Delete(context.TODO(), globalCIDRConfigMapName, metav1.GetOptions{})
}
```

to take care of that. I'll push a commit to this PR to do that.

skitt (approved):

PR #2104, opened 2022-06-13
Separate system tests out of E2E

### LINKED ISSUE #967: Move "E2E" verification tests to system tests
This was raised in #962 initially but deserves it's own issue.

Currently we have a lot of verifications for subctl and operator happening in https://github.com/submariner-io/submariner-operator/blob/master/scripts/kind-e2e/lib_operator_verify_subm.sh
this is being executed as part of the E2E tests but unlike the E2E tests which are "black box", these tests are very "white box" and check that the deployment happened as expected (e.g. the cable driver was set and all that).

It would be better to move these tests into a kind of "system test" style of testing, probably in Go, where the tests could run subctl and verify that the expected outcome has happened (e.g. submariner CR was created with correct fields) and furthermore test that the operator deploys correctly.

An added benefit of using a proper testing framework would be that we'd have a test report summarizing which tests failed and which passed, as opposed to the current situation in which verification would halt at the first failure.

For this purpose we could utilize an "all in one" kind environment of just 1 cluster with 1 node, since we don't really care to see that submariner is functional but rather that the expected state has been reached (i.e. deployments, daemonsets, and all expected resources have been properly created).

mkolesnik:
This PR separates the system tests out of E2E, making E2E super simple and system tests more focused.

As part of this work the E2E tests have also been improved somewhat.

Please see individual commits for more details.


.github/workflows/system.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: System tests
```
dfarrell07: This ends up showing in the UI of all PRs, and currently it's Leading Caps in all workflows in all repos I think. Would be good to be consistent.

```suggestion
name: System Tests
```

Also, I wonder if anyone has a suggestion for a more descriptive workflow/job name than "System Tests/Test"? I get the idea, but it took some thinking and I authored a lot of these tests. :stuck_out_tongue_winking_eye: A more verbose job name maybe? /me is struggling to think of a good suggestion, will follow up if I think of one

mkolesnik: Updated the caps, I'm not married to the name if you have a better suggestion I don't mind changing it

dfarrell07 (approved):

.github/workflows/system.yml
tpantelis: Unfortunately only the job name is displayed in the branch settings tab so having that more descriptive helps. I suggest job name "_Deployment Test_", so full  name "_System Tests/Deployment Test_".

.github/workflows/system.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: System Tests
+
+on:
+  pull_request:
+
+jobs:
+  system:
+    name: Test
```
tpantelis: Unfortunately only the job name is displayed in the branch settings tab so having that more descriptive helps. I suggest job name "_Deployment Test_", so full name "_System Tests/Deployment Test_".

```suggestion
    name: Deployment Test
```

tpantelis (changes requested):

tpantelis (approved):

dfarrell07:
> Error: two kubecontexts must be specified

mkolesnik:
> > Error: two kubecontexts must be specified

Right, I was foolishly trying to avoid parsing the settings file to get the clusters :)

I amended the necessary commit to add code to parse the settings properly, and while at it squashed another commit (the rename) into the one above it.

The PR is otherwise unchanged.

skitt (approved):

PR #2105, opened 2022-06-13
Fix the case of the multi-arch builds

skitt:


dfarrell07:
I think this is just the original commit, not the commit with the fix (no git diff).

dfarrell07:
Sent https://github.com/submariner-io/submariner-operator/pull/2106 with the desired change.

PR #2126, opened 2022-07-05
Remove uses of the Submariner clientset

skitt:
Controllers are supposed to use the client given to them by the
manager, instead of using generated clientsets. This removes our uses
of the Submariner clientset, except (for now) in tests.

Ultimately this will simplify the operator rebuild, since we won't
have to take care of rebuilding the clientset.

The clientset itself can't be removed yet, subctl imports it.


dfarrell07 (approved):

pkg/discovery/network/network.go
```diff
@@ -63,7 +64,7 @@ func (cn *ClusterNetwork) IsComplete() bool {
 	return cn != nil && len(cn.ServiceCIDRs) > 0 && len(cn.PodCIDRs) > 0
 }
 
-func Discover(dynClient dynamic.Interface, clientSet kubernetes.Interface, operatorClient operatorclientset.Interface,
+func Discover(dynClient dynamic.Interface, clientSet kubernetes.Interface, operatorClient client.Client,
```
tpantelis, line 67: So this introduces yet another client API to callers in addition to the dynamic API and K8s generated API. Perhaps we should just simplify and use the `controller-runtime` API in lieu of the `kubernetes.Interface` as well. I think we still need the dynamic API when the Go type isn't available.

pkg/client/default_producer.go
```diff
@@ -49,7 +47,3 @@ func (p *DefaultProducer) ForDynamic() dynamic.Interface {
 func (p *DefaultProducer) ForSubmariner() submarinerclient.Interface {
 	return p.SubmarinerClient
 }
-
-func (p *DefaultProducer) ForOperator() operatorclient.Interface {
```
tpantelis, line 53: **subctl** uses this.

skitt: Yup, I saw that too :-/.

pkg/discovery/network/network.go
skitt: This is based on the `controller-runtime` pattern (where `Reconciler`s encapsulate a `client.Client` obtained from `mgr`).

tpantelis: Right - I suggest if we're going to switch to use the `controller-runtime` API in lieu of the operator client set then let's do the same for the other generated client sets. Also change the `client.Producer`.

pkg/client/default_producer.go
mkolesnik: Perhaps we need a test for subm-op to make sure were not breaking subctl

Jaanki:
Please also remove the [make target](https://github.com/submariner-io/submariner-operator/blob/devel/Makefile#L183) to generate the clientset.

Jaanki (changes requested):
Remove https://github.com/submariner-io/submariner-operator/blob/devel/Makefile#L183 target.

skitt:
> Please also remove the [make target](https://github.com/submariner-io/submariner-operator/blob/devel/Makefile#L183) to generate the clientset.

I did this, but then reverted it because we still need to be able to create the clientset for `subctl`. I’ll drop the Makefile section when we can actually delete all the clientset code.

pkg/discovery/network/network.go
skitt: I’ve made this PR specifically about the Submariner clientset, we can address the others in different PRs.

tpantelis (approved):

tpantelis:
> > Please also remove the [make target](https://github.com/submariner-io/submariner-operator/blob/devel/Makefile#L183) to generate the clientset.
> 
> I did this, but then reverted it because we still need to be able to create the clientset for `subctl`. I’ll drop the Makefile section when we can actually delete all the clientset code.

We're going to need to adjust **subctl** anyway so I don't think we need to worry about restricting changes. But we can make the changes in multiple PRs then, once done, bump and adjust **subctl**.

Jaanki (approved):

PR #2128, opened 2022-07-06
[Backport] Fix intermittent failure with diagnose metrics cmd

sridhargaddam:
In the latest OCP deployments, it is seen that when the
validate-client pod is scheduled it gets terminated pretty
quickly even before its able to send the required data. We
have addressed similar issue earlier by modifying the client
pod to use HostNetworking enabled. This PR implements a similar
change even for diagnose firewall metrics command.

Fixes: https://github.com/submariner-io/subctl/issues/149


nyechiel (approved):

nyechiel:
@skitt if you can approve this one, we can proceed with releasing 0.12.2

skitt (approved):

PR #2146, opened 2022-07-19
Support for default subnet in weavenet CNI discovery

### LINKED ISSUE #2144: CNI plugin discover does not work when using Weave-net with default subnet
**What happened**:
CNI plugin discovery does not work when using Weave-net with default subnet (10.32.0.0/12), because the **IPALLOC_RANGE** environment variable is not explicitly set (https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/weavenet.go#L38)

CNI plugin discovery should work with Weave-net when using the default subnet

Install Weave-net without explicitly defining the subnet.
I think this maybe easily reproducible on a Kind cluster installing subnet with the following command:
```
kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
```
instead of https://github.com/submariner-io/shipyard/blob/9932f01209df6e092c4d395bf404745a3cb8e588/scripts/shared/clusters.sh#L168
Try to use submariner, you will get the message:
```
✓ Discovering network details 
    Network plugin: generic
```

I think this would be easily solved improving the logic of `submariner-operator/blob/devel/pkg/discovery/network/weavenet.go`:
 - check for the label name=weave-net
     * if IPALLOC_RANGE is defined
       + subnet = IPALLOC_RANGE
     * else             # this means it is using the default weave-net subnet 
       + subnet = 10.32.0.0/12

**Environment**:
subctl version: v0.13.0-rc2

josecastillolema:
Closes https://github.com/submariner-io/submariner-operator/issues/2144

### DIFF
--- a/pkg/discovery/network/weavenet.go
+++ b/pkg/discovery/network/weavenet.go
@@ -23,7 +23,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-// nolint:nilnil // Intentional as the purpose is to discover.
 func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, error) {
 	weaveNetPod, err := FindPod(clientSet, "name=weave-net")
 
@@ -31,23 +30,17 @@ func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, erro
 		return nil, err
 	}
 
-	var clusterNetwork *ClusterNetwork
+	clusterNetwork := &ClusterNetwork{
+		NetworkPlugin: cni.WeaveNet,
+	}
 
 	for i := range weaveNetPod.Spec.Containers {
 		for _, envVar := range weaveNetPod.Spec.Containers[i].Env {
 			if envVar.Name == "IPALLOC_RANGE" {
-				clusterNetwork = &ClusterNetwork{
-					PodCIDRs:      []string{envVar.Value},
-					NetworkPlugin: cni.WeaveNet,
-				}
-
+				clusterNetwork.PodCIDRs = []string{envVar.Value}
 				break
 			}
 		}
-	}
-
-	if clusterNetwork == nil {
-		return nil, nil
 	}
 
 	clusterIPRange, err := findClusterIPRange(clientSet)


pkg/discovery/network/weavenet.go
```diff
@@ -34,16 +34,23 @@ func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, erro
...
 					NetworkPlugin: cni.WeaveNet,
 				}
-
+				found = true
 				break
 			}
 		}
+		if !found {
+			clusterNetwork = &ClusterNetwork{
+				PodCIDRs:      []string{"10.32.0.0/12"},
```
sridhargaddam: Thank you @josecastillolema for the PR. But I'm not sure if this is the approach we want to take as hard-coding the CIDRs could be error prone in future when weave-net decides to modify the default CIDR.

We have options to explicitly specify the `PodCIDR` and `ServiceCIDR` while joining the clusters. Will that option work for your use-case?

```
subctl join --clustercidr .... --servicecidr 
```

sridhargaddam (changes requested):

pkg/discovery/network/weavenet.go
josecastillolema: Imho the problem is not the fact that it not detects automatically the CIDR, is super easy to provide it. The problem is that it fails to detect the CNI.
Would an approach where the **_NetworkPlugin_** was defined but the **_PodCIDRs_** set to nil be prefered? This way, the CNI would be discovered and we would not have any problems in case the default CIDR changes in the future.

skitt: Perhaps another solution would be to retrieve the default from Weave somehow, but so far I haven’t figured out where it’s defined. I can find the unit tests that verify the default, but not the default itself!

skitt: The default is set here: https://github.com/weaveworks/weave/blob/master/weave#L1155 (see also https://github.com/weaveworks/weave/commit/f593c2fbf2c0dbffc77465c7a8acbe74e89f54eb). We can’t extract it from there easily ;-).

tpantelis: It seems reasonable to still report the plugin was detected as @josecastillolema suggests. If the `PodCIDRs` isn't set, subsequent code will try a generic approach to detect it. We can just define `clusterNetwork` on L34:

```Go
clusterNetwork := &ClusterNetwork{
    NetworkPlugin: cni.WeaveNet,
}
```
and then set `PodCIDRs` later.

tpantelis (changes requested):

pkg/discovery/network/weavenet.go
josecastillolema: Agree @tpantelis .
Adjusted as requested.

pkg/discovery/network/weavenet.go
```diff
@@ -34,16 +34,22 @@ func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, erro
...
-
+				found = true
 				break
 			}
 		}
+		if !found {
+			clusterNetwork = &ClusterNetwork{
+				NetworkPlugin: cni.WeaveNet,
+			}
+		}
```
tpantelis: I suggest:

```Go
clusterNetwork := &ClusterNetwork{
    NetworkPlugin: cni.WeaveNet,
}

for i := range weaveNetPod.Spec.Containers {
    for _, envVar := range weaveNetPod.Spec.Containers[i].Env {
        if envVar.Name == "IPALLOC_RANGE" {
	    clusterNetwork.PodCIDRs = []string{envVar.Value}
            break
	}
    }
}
```

Also remove lines 55-57 below.

tpantelis (changes requested):

pkg/discovery/network/weavenet.go
skitt: Since we assume the CNI is Weave if any Weave pod is present, we can simplify this slightly:

```
		clusterNetwork = &ClusterNetwork{
			NetworkPlugin: cni.WeaveNet,
		}
		for _, envVar := range weaveNetPod.Spec.Containers[i].Env {
			if envVar.Name == "IPALLOC_RANGE" {
				clusterNetwork.PodCIDRs = []string{envVar.Value}

				break
			}
		}
```

skitt (changes requested):

pkg/discovery/network/weavenet.go
skitt: Hah, ninja’ed by Tom!

tpantelis: 😆

tpantelis:
The commits need to have `Signed-off-by` to fix the DCO job failure.

josecastillolema:
> The commits need to have `Signed-off-by` to fix the DCO job failure.

Done @tpantelis

tpantelis:
> > The commits need to have `Signed-off-by` to fix the DCO job failure.
> 
> Done @tpantelis

The [original  commit](https://github.com/submariner-io/submariner-operator/pull/2146/commits/20733c6f75869657fce00fc914744fc18dc6f938) is not signed off. I suggest to just squash the commits.

tpantelis (approved):

pkg/discovery/network/weavenet.go
```diff
@@ -33,23 +33,19 @@ func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, erro
 
 	var clusterNetwork *ClusterNetwork
```
skitt: This is no longer necessary.

```suggestion
```

skitt (changes requested):

josecastillolema:
> > > The commits need to have `Signed-off-by` to fix the DCO job failure.
> > 
> > 
> > Done @tpantelis
> 
> The [original commit](https://github.com/submariner-io/submariner-operator/pull/2146/commits/20733c6f75869657fce00fc914744fc18dc6f938) is not signed off. I suggest to just squash the commits.

Commits squashed @tpantelis

sridhargaddam (approved):

pkg/discovery/network/weavenet.go
```diff
@@ -31,25 +31,19 @@ func discoverWeaveNetwork(clientSet kubernetes.Interface) (*ClusterNetwork, erro
 		return nil, err
 	}
 
-	var clusterNetwork *ClusterNetwork
+	clusterNetwork = &ClusterNetwork{
```
skitt: Sorry, with my earlier change request I missed that this should now be

```suggestion
	clusterNetwork := &ClusterNetwork{
```

skitt (changes requested):

skitt (approved):

tpantelis:
One remaining issue - linting error:

```
pkg/discovery/network/weavenet.go:26:1: directive `// nolint:nilnil // Intentional as the purpose is to discover.` is unused for linter "nilnil" (nolintlint)
```

The code no longer returns `nil, nil` so the linting exception needs to be removed.

PR #2147, opened 2022-07-19
Fix KIND deployments using ovn

### LINKED ISSUE #2145: OVN CI Jobs are running with weave instead of ovn-k
**What happened**:
The CI jobs like "[E2E (1.24, ovn)]" ideally should be running with OVN-Kubernetes CNI, but instead use weave giving a false impression that jobs are validated on OVNK. Looks like the "using=ovn" flag is broken.

mkolesnik:
As the `using` flag is needed before it's processed (in Shipyard's
`Makefile.inc`), use the one processed in Shipyard's `Makefile.dapper`.

Resolves:  #2145


dfarrell07 (approved):
This does seem to get us using OVN, not Weave.

https://github.com/submariner-io/submariner-operator/runs/7407262841?check_suite_focus=true

nyechiel (approved):

skitt (approved):

PR #2150, opened 2022-07-20
Add a GHA to test subctl

skitt:
This ensures we know whether changes made to the operator are going to
break subctl, before merging them.

### DIFF
--- a/.github/workflows/subctl-unit.yml
+++ b/.github/workflows/subctl-unit.yml
@@ -1,0 +1,37 @@
+---
+name: Consuming Projects
+
+on:
+  pull_request:
+
+jobs:
+  unit-testing:
+    name: Check subctl
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the submariner-operator repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          path: submariner-operator
+
+      - name: Check out the subctl repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          repository: submariner-io/subctl
+          path: subctl
+
+      - name: Check out the shipyard repository
+        # This is required so that we can run a build involving multiple
+        # repositories (using LOCAL_BUILD=1)
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          repository: submariner-io/shipyard
+          path: shipyard
+
+      - name: Update the subctl build to use the current submariner-operator
+        run: |
+          cd subctl
+          go mod edit -replace=github.com/submariner-io/submariner-operator=../submariner-operator
+
+      - name: Run Go subctl unit tests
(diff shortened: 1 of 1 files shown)


skitt:
See https://github.com/submariner-io/submariner-operator/runs/7433523195?check_suite_focus=true for a correct failure related to breaking changes in https://github.com/submariner-io/submariner-operator/pull/2126

.github/workflows/subctl-unit.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request:
+  push:
+    tags:
+      - 'v**'
+
+jobs:
+  unit-testing:
+    name: subctl Check
```
mkolesnik:
```suggestion
    name: Check subctl
```

.github/workflows/subctl-unit.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Downstream Compatibility Check
+
+on:
+  pull_request:
+  push:
+    tags:
+      - 'v**'
```
mkolesnik: Do we really need this on push to tags?

.github/workflows/subctl-unit.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Downstream Compatibility Check
```
mkolesnik:
```suggestion
name: Consuming Projects
```

IMHO it would be more consistent with similar testing in Shipyard.
Also it's shorter so it renders better on GHA (all other workflow names are rather short)

WDYT?

.github/workflows/subctl-unit.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          path: submariner-operator
+
+      - name: Check out the subctl repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          repository: submariner-io/subctl
+          path: subctl
+
+      - name: Check out the shipyard repository
```
mkolesnik, line 23: Why is Shipyard needed for this one?

mkolesnik (changes requested):
Thanks, some questions and suggestions

.github/workflows/subctl-unit.yml
skitt: Ah, no, we don’t, that was copy-pasta.

.github/workflows/subctl-unit.yml
skitt: That’s a great idea, I like the consistency!

.github/workflows/subctl-unit.yml
skitt: We need to run a “local build” because we need several repositories (since only the repository running `.dapper` is made available inside the container), which means we need a local clone of the Shipyard repository. I’ll add a comment explaining this.

dfarrell07 (approved):

mkolesnik (approved):

PR #2160, opened 2022-08-01
Use controller-runtime client in pkg/discover

tpantelis:
...in lieu of the kube client and dynamic client. See individual commits for details.


skitt (approved):

dfarrell07 (approved):

tpantelis:
@skitt I've run into an issue with using the controller client. The network discovery code needs to be able to list pods in any namespace (eg to find the weave-net pod). However the `Client` instance used in the controller is scoped and restricted to a specific namespace determined by the [WATCH_NAMESPACE env](https://github.com/submariner-io/submariner-operator/blob/6d222218f80b671069267bfa62b19d1e2d3c3e15/main.go#L289) (ie `submariner-operator`). So it cannot be used for the network discovery. We could create an unrestricted controller client to use for the network discovery (call it `GeneralClient`) or just leave it the way it is. The latter would be the easiest but it would be nice to only have to pass one client to the discovery code. WDYT?

skitt:
Ah yes, I’d forgotten about that! The “scoping” of the operator is actually something we’re supposed to track from a security perspective, so it would be helpful to rename our current client to `ScopedClient` and create a `GeneralClient` for network discovery. (Or perhaps `OperatorNamespaceClient` and `AllNamespaceClient` which is more explicit but somewhat unwieldy.)

tpantelis:
> it would be helpful to rename our current client to `ScopedClient` and create a `GeneralClient` for network discovery. 

I was thinking `ScopedClient` and `GeneralClient` as well 😄

dfarrell07 (approved):
> `OperatorNamespaceClient` and `AllNamespaceClient`

As someone less familiar with the code, I find these names provide nice guidance. But yes I guess they are pretty verbose.

PR #2168, opened 2022-08-14
submariner-diagnose-role: add delete permissions for pods.

tomer-shefler:
These are required for kubeproxy and firewall tests

### DIFF
--- a/config/rbac/submariner-diagnose/role.yaml
+++ b/config/rbac/submariner-diagnose/role.yaml
@@ -10,6 +10,7 @@ rules:
     resources:
       - pods
     verbs:
+      - delete
       - create
       - get
       - list
--- a/pkg/embeddedyamls/yamls.go
+++ b/pkg/embeddedyamls/yamls.go
@@ -2721,6 +2721,7 @@ rules:
     resources:
       - pods
     verbs:
+      - delete
       - create
       - get
       - list


tomer-shefler:
related to: https://github.com/submariner-io/subctl/pull/183

pkg/embeddedyamls/yamls.go
```diff
@@ -2783,6 +2783,7 @@ rules:
     resources:
       - pods
     verbs:
+      - delete
```
mkolesnik, line 2724: This file is auto generated, you should look at the original file `config/rbac/submariner-diagnose/role.yaml`

mkolesnik (changes requested):

mkolesnik (approved):
LGTM, please address CI failures

dfarrell07:
[Embedded YAMLs up-to-date](https://github.com/submariner-io/submariner-operator/runs/8074759735?check_suite_focus=true#logs) is failing. Running `make pkg/embeddedyamls/yamls.go`, [as the CI does](https://github.com/submariner-io/submariner-operator/blob/devel/.github/workflows/linting.yml#L174), should make the relevant changes for you.

pkg/embeddedyamls/yamls.go
```diff
@@ -40,7 +40,7 @@ spec:
   - name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: Broker is the Schema for the brokers API.
+        description: Broker is the Schema for the brokers API
```
dfarrell07: Do you mean to remove these periods? If so, maybe that should be a different PR, or at least the reasoning documented in the commit message.

dfarrell07: Also this is the generated version.

config/rbac/submariner-diagnose/role.yaml
```diff
@@ -10,6 +10,7 @@ rules:
     resources:
       - pods
     verbs:
+      - delete
```
dfarrell07, line 13: Can you be more specific about what needs this permission?

pkg/embeddedyamls/yamls.go
tomer-shefler: This dots were added by `make pkg/embeddedyamls/yamls.go`

config/rbac/submariner-diagnose/role.yaml
tomer-shefler: This function deletes the pods.
https://github.com/submariner-io/subctl/blob/aedd7a6e8e43e95b809f0caa96325f45213b46e9/internal/pods/schedule.go#L182
And it's being called by tests. After pods spawned by the test are done, they are deleted by the test the spawned them

tpantelis (approved):

PR #2169, opened 2022-08-14
diagnose clusterrole: remove pods sections:

tomer-shefler:
all pods operations are being done in submariner-operator namespace
and thus these permessions are no longer required on the cluster role
scope

Signed-off-by: Tomer Shefler <sheflertomer@gmail.com>

### DIFF
--- a/pkg/embeddedyamls/yamls.go
+++ b/pkg/embeddedyamls/yamls.go
@@ -2779,14 +2779,6 @@ rules:
     verbs:
       - get
       - list
-  - apiGroups:  # pods are created to run firewall diagnostics
-      - ""
-    resources:
-      - pods
-    verbs:
-      - create
-      - get
-      - list
 `
 	Config_rbac_submariner_diagnose_cluster_role_binding_yaml = `---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/config/rbac/submariner-diagnose/cluster_role.yaml
+++ b/config/rbac/submariner-diagnose/cluster_role.yaml
@@ -14,11 +14,3 @@ rules:
     verbs:
       - get
       - list
-  - apiGroups:  # pods are created to run firewall diagnostics
-      - ""
-    resources:
-      - pods
-    verbs:
-      - create
-      - get
-      - list


tomer-shefler:
solves https://github.com/submariner-io/subctl/issues/48
Is already possible due to https://github.com/submariner-io/subctl/pull/126

pkg/embeddedyamls/yamls.go
```diff
@@ -2837,14 +2837,6 @@ rules:
     verbs:
       - get
       - list
-  - apiGroups:  # pods are created to run firewall diagnostics
```
mkolesnik, line 2782: This file is auto generated, you should be editing the source file `config/rbac/submariner-diagnose/cluster_role.yaml`

mkolesnik (changes requested):

skitt:
Thanks for submitting this! `pkg/embeddedyamls/yamls.go` is generated, please make this change in `config/rbac/submariner-diagnose/cluster_role.yaml` too (and then run `make pkg/embeddedyamls/yamls.go`).

Please fix your commit message too:

> Commit e3ba4af3d1:
> 1: T3 Title has trailing punctuation (:): "diagnose clusterrole: remove pods sections:"
> 2: B4 Second line is not empty: "all pods operations are being done in submariner-operator namespace"

skitt:
This would break `subctl diagnose --namespace …` for any namespace other than the operator namespace (perhaps we should remove that option).

mkolesnik:
> This would break `subctl diagnose --namespace …` for any namespace other than the operator namespace (perhaps we should remove that option).

We can leave the flag but document it, e.g. that when running diagnose the account used needs to have access to creating/deleting pods on the namespace if it's not the default one.

However, I think removing it is simplest as it doesn't add much value.
I think it was added to support custom operator namespace, but a more robust solution is to auto detect the operator namespace and use it for the diagnosis, without bothering the user.

tomer-shefler:
> > This would break `subctl diagnose --namespace …` for any namespace other than the operator namespace (perhaps we should remove that option).
> 
> We can leave the flag but document it, e.g. that when running diagnose the account used needs to have access to creating/deleting pods on the namespace if it's not the default one.
> 
> However, I think removing it is simplest as it doesn't add much value. I think it was added to support custom operator namespace, but a more robust solution is to auto detect the operator namespace and use it for the diagnosis, without bothering the user.

So, the desicion, as I understantd it, is that this namespace option shloud be remained. However a user who wants to use a different namespace should take care of all of the RBACs required to do so?

mkolesnik (approved):
> So, the desicion, as I understantd it, is that this namespace option shloud be remained. However a user who wants to use a different namespace should take care of all of the RBACs required to do so?

For me its fine as long as you add the details to the website documentation.
I'm also open for entirely removing the namespace option.

@skitt - WDYT?

mkolesnik (approved):

mkolesnik:
@tomer-shefler make sure to generate the embedded yamls correctly, currently its failing the CI

tpantelis (approved):

PR #2175, opened 2022-08-16
Use the builtin prometheus service monitors

Jaanki:
and not create them manually


tpantelis:
How are these changes going to play with the [metrics redesign](https://github.com/submariner-io/enhancements/pull/128) or are they not related? @vthapar

Jaanki:
> How are these changes going to play with the [metrics redesign](https://github.com/submariner-io/enhancements/pull/128) or are they not related? @vthapar

This does not affect `globalnet` and `gateway` metrics. This change is about not manually creating `ServiceMonitors` for metrics defined in the [controller](https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/submariner_metrics.go) but let the operator itself create the `ServiceMonitors`. The latest [kubebuilder guide](https://book.kubebuilder.io/reference/metrics.html) explains the intent behind this PR.

skitt:
There are a couple of parts missing for this to work.

The first is that we still need to set up a metrics endpoint. This has been moved into the operator framework library, and https://sdk.operatorframework.io/docs/building-operators/golang/migration/ suggests that metrics are served automatically; in my tests, that doesn’t happen. The controller runtime can start a listener, I’m checking if we’re missing anything in our use of the operator framework or if we need to start the listener manually.

The second is that we also need to set up a metrics service and metrics service monitor. The former _should_ be handled by the framework, and hopefully solving the above will take care of this. The latter is now delegated to `kubebuilder`, but that’s not magic: it populates the service monitor template, but it doesn’t install it. It’s not clear to me whether the operator image which can be built with `kubebuilder` takes care of this, or if we’ll need to keep our own service monitor generation code.

skitt:
The metrics endpoint _is_ automatically set up on port 8383.

Jaanki:
Closing in favor of #2199

PR #2180, opened 2022-08-17
Create all CRDs before initialising the scheme

skitt:
This avoids having to restart the operator after CRD setup.

The broker is left alone for the time being since its behaviour is
slightly different.


tpantelis (approved):

dfarrell07 (approved):
Nice, yay for fewer restarts!

PR #2186, opened 2022-08-18
Check the running operator version in e2e

skitt:
Once the clusters have been deployed, check that we're actually
running the version of the operator we expect to test.

### DIFF
--- a/scripts/test/e2e.sh
+++ b/scripts/test/e2e.sh
@@ -17,6 +17,18 @@ ${SCRIPTS_DIR}/e2e.sh "$@"
 # Reload KUBECONFIG in case deploy was triggered
 . "${SCRIPTS_DIR}/lib/kubecfg"
 
+# Check that we're testing the version we expect
+function check_operator_version() {
+    local opv
+    opv=$(kubectl --context="$cluster" logs -n submariner-operator -l name=submariner-operator --tail=-1 | awk '/Submariner operator version/ { print $NF }')
+    printf "Cluster %s is running operator version %s\n" "$cluster" "$opv"
+    if [ "$opv" != "$VERSION" ]; then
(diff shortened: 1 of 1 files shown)


scripts/test/e2e.sh
```diff
@@ -17,6 +17,18 @@ ${SCRIPTS_DIR}/e2e.sh "$@"
 # Reload KUBECONFIG in case deploy was triggered
 . "${SCRIPTS_DIR}/lib/kubecfg"
 
+# Check that we're testing the version we expect
+function check_operator_version() {
+    local opv
+    kubectl --context=$cluster logs -n submariner-operator -l name=submariner-operator --tail=-1 | grep "Submariner operator version"
```
tpantelis: This line looks extraneous...

skitt: Yes, it’s to check what happens here on e2e.

tpantelis: Sorry not sure I follow....isn't it basically the same as the command run on the next line? Is this line just to print it out?

skitt: That’s right, this line is just there so the log contents show up in the build log. I found it useful when debugging this on my own setup; if this fails to do the right thing on CI, I’d rather have this info too. I’ll remove the line before this is merged.

tpantelis: Doesn't hurt to always print it - how about something like:

```bash
vs=$(kubectl --context=$cluster logs -n submariner-operator -l name=submariner-operator --tail=-1 | grep "Submariner operator version")
echo $vs
opv=$(awk '/$vs/ { print $NF }')
```

tpantelis (approved):

dfarrell07 (approved):

PR #2191, opened 2022-08-21
Remove `names.ImagePrefix/ImagePostfix`

mkolesnik:
These have been deprecated for a long while now, removing them adds to
the maintainability of our code.


tpantelis (approved):

Jaanki (approved):

skitt (approved):
These are no longer used downstream so we won’t even need to deal with any fallout from this.

PR #2192, opened 2022-08-22
Set DNSPolicy to ClusterFirstWithHostNet

vthapar:
For pods that run with `HostNetworking: true` DNSPolicy should be set to
`ClusterFirstWithHostNet`

Refer https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

Fixes https://github.com/stolostron/backlog/issues/25169

### DIFF
--- a/controllers/submariner/gateway_resources.go
+++ b/controllers/submariner/gateway_resources.go
@@ -215,9 +215,9 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 			},
 			ServiceAccountName:            names.GatewayComponent,
 			HostNetwork:                   true,
+			DNSPolicy:                     corev1.DNSClusterFirstWithHostNet,
 			TerminationGracePeriodSeconds: pointer.Int64(1),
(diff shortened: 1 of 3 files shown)


controllers/submariner/gateway_resources.go
```diff
@@ -217,7 +217,7 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 			HostNetwork:                   true,
 			TerminationGracePeriodSeconds: pointer.Int64(1),
 			RestartPolicy:                 corev1.RestartPolicyAlways,
-			DNSPolicy:                     corev1.DNSClusterFirst,
+			DNSPolicy:                     corev1.DNSClusterFirstWithHostNet,
```
skitt: Could you move this up a couple of lines so it’s immediately after `HostNetwork: true`? That way if we ever change the `HostNetwork` setting we’ll have a better chance of noticing the associated `DNSPolicy`.

skitt (changes requested):

skitt (approved):

tpantelis (approved):

PR #2199, opened 2022-08-23
Use current operator SDK-style metrics setup

skitt:
Instead of starting our own HTTP listener, use the server provided by
the operator library.

The current operator SDK produces service and service monitor
templates for use with Kubebuilder. We still need to support setting
up metrics without Kubebuilder, so we still take care of creating the
metrics service and service monitor; however, instead of using the old
SDK-inspired code, which needs obsolete SDK packages, use the metrics
setup code we have for our own managed applications. The major
difference between the two is that the old SDK-inspired code set the
service's owner reference appropriately, whereas the new code doesn't;
but the Kubebuilder-based setup also does *not* set the owner
reference, so this doesn't seem too bad.

The Kubebuilder-based setup also supports an authenticating proxy,
which this doesn't; using this also requires adding permissions for
Prometheus, and should be handled as a separate change.


tpantelis (approved):

main.go
```diff
@@ -158,6 +150,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	log.Info("Setting up metrics services and monitors")
+
+	// Setup the metrics services and service monitors
+	labels := map[string]string{"name": os.Getenv("OPERATOR_NAME")}
+	if err := metrics.Setup(namespace, nil, labels, metricsPort, mgr.GetClient(), cfg, scheme, log); err != nil {
```
skitt: This is failing with

```
Error setting up metrics services and monitors error="error creating or updating Service submariner-operator/submariner-operator-metrics: the cache is not started, can not read objects" 
```

dfarrell07 (approved):

dfarrell07:
[E2E (globalnet, 1.24, lighthouse)](https://github.com/submariner-io/submariner-operator/runs/8019177944?check_suite_focus=true#logs) timed out at 30m, re-running.

```
   • Checking the firewall configuration to determine if intra-cluster VXLAN traffic is allowed  ...
  
  Error: The operation was canceled.
```

PR #2209, opened 2022-08-26
Ensure we build the operator image for Prometheus

skitt:
When testing metrics, we want to ensure the operator image is
rebuilt.

Fix metric service setup to actually take the updated key into account
when checking for "name" labels.


skitt:
Operator metrics are currently broken because the metrics service is created with the wrong label selector.

tpantelis (approved):

dfarrell07 (approved):

mkolesnik (approved):

PR #2219, opened 2022-08-30
ABC-order variant analysis linting config

dfarrell07:
This job was called vulnerability variant analysis during most of
development. It should have been reordered when the name was shortened.

Relates-to: submariner-io/submariner#1970

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -125,6 +125,24 @@ jobs:
       - name: Run shellcheck
         run: make shellcheck
 
+  variant-analysis:
+    name: Variant Analysis
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@86f3159a697a097a813ad9bfa0002412d97690a4
+        with:
+          languages: go
+      - name: Build code, creating CodeQL database
+        run: make build
+      - name: Run CodeQL variant analysis
+        uses: github/codeql-action/analyze@86f3159a697a097a813ad9bfa0002412d97690a4
+      - name: Show CodeQL scan SARIF report
+        if: always()
+        run: cat ../results/go.sarif
+
   vulnerability-scan:
     name: Vulnerability Scanning
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/linting.yml
```diff
@@ -116,6 +116,24 @@ jobs:
       - name: Run packagedoc-lint
         run: make packagedoc-lint
 
+  variant-analysis:
+    name: Variant Analysis
```
skitt, line 129: Wouldn’t it be better to give this job a self-explanatory name? As it stands, it only indicates that it’s a variant of some other job. Wouldn’t it be better to name it “Vulnerability Scanning (CodeQL)” (and rename the Anchore job accordingly)?

skitt (changes requested):

.github/workflows/linting.yml
dfarrell07: "Variant" in this case isn't meant as "variant of some other job", but instead as the name of a category of static analysis.

> Variant analysis is the process of using a known security vulnerability as a seed to find similar problems in your code. It’s a technique that security engineers use to identify potential vulnerabilities, and ensure these threats are properly fixed across multiple codebases.
Querying code using CodeQL is the most efficient way to perform variant analysis. You can use the standard CodeQL queries to identify seed vulnerabilities, or find new vulnerabilities by writing your own custom CodeQL queries. Then, develop or iterate over the query to automatically find logical variants of the same bug that could be missed using traditional manual techniques.

https://codeql.github.com/docs/codeql-overview/about-codeql/

CodeQL doesn't only do variant analysis for security issues, it also has semantic queries/rules for other types of issues.

https://github.com/github/codeql/tree/main/go/ql/src

I dropped "vulnerability" from the description because I don't think any of the [issues it found in our codebase](https://github.com/submariner-io/submariner/issues/1970#issue-1343160083) were security-related.

I can add some more docs like this to the commit messages?

skitt: Ah right, TIL, thanks! I hadn’t found anything relevant in a web search, so it might be worth at least linking to the “about CodeQL” page in the commit message.

dfarrell07: Yeah there aren't a lot of docs, I'll spread some more info around.

tpantelis:
It looks like the System Tests were broken by https://github.com/submariner-io/shipyard/pull/924 @mkolesnik

dfarrell07:
> It looks like the System Tests were broken by [submariner-io/shipyard#924](https://github.com/submariner-io/shipyard/pull/924) @mkolesnik

Cross-linking [my comment about adding System Tests to the Consuming Projects tests](https://github.com/submariner-io/shipyard/pull/924#issuecomment-1233061460).

tpantelis (approved):

dfarrell07:
The rebase here keeps getting broken by updates to the codeql-action GHA SHA.

.github/workflows/linting.yml
dfarrell07: Added to the commit message :heavy_check_mark:

skitt (approved):

PR #2230, opened 2022-09-06
Add health and readyness probes

Jaanki:
and use the new ctrl.NewManager, which is an alias of manager.Manager to
create the manager.

Various options are defined at
https://godocs.io/sigs.k8s.io/controller-runtime/pkg/manager#Options.

Health and readyness logic is not implemented yet but are added for
future usage.

Part of epic: https://github.com/submariner-io/enhancements/issues/81


skitt (approved):

main.go
```diff
@@ -73,7 +74,15 @@ func init() {
 	flag.BoolVar(&help, "help", help, "Print usage options")
 }
 
+// nolint:gocyclo // No further refactors necessary
```
Jaanki, line 77: @skitt I have added this ignore `gocyclo` lint after your +1. So if you can reapprove.

main.go
```diff
@@ -73,7 +74,15 @@ func init() {
...
 func main() {
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
```
tpantelis, line 84: Why do we need these as CLI params and how would we set them?

main.go
```diff
@@ -139,11 +148,14 @@ func main() {
...
-		Namespace:          namespace,
-		MapperProvider:     apiutil.NewDiscoveryRESTMapper,
-		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
-		Scheme:             scheme,
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "2a1e5b0d.submariner.io",
```
tpantelis: Where did the magic number "2a1e5b0d" come from?

main.go
```diff
@@ -203,6 +215,15 @@ func main() {
 
 	// +kubebuilder:scaffold:builder
 
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		log.Error(err, "unable to set up health check")
+		os.Exit(1)
+	}
+
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		log.Error(err, "unable to set up ready check")
```
tpantelis, line 225:
```suggestion
		log.Error(err, "unable to set up ready check")
		os.Exit(1)
```

tpantelis (changes requested):

main.go
Jaanki: These are needed if we want to customise `health-probe-bind-address` or enable leader election for controller manager. These can be set during runtime like this

`go run main.go --probeAddr 8082`

Again this is an unused code added for future use.

main.go
Jaanki: Thats autogenerated. Its explanation from official doc

```
// LeaderElectionID determines the name of the resource that leader election
// will use for holding the leader lock.
LeaderElectionID string
```

tpantelis: OK - can you add a comment explaining this?

Jaanki: Done

tpantelis (approved):

PR #2235, opened 2022-09-07
Add judgment before split the string

pengbinbin1:
Which issue(s) this PR fixes:
Without the judgment before split, the slices's length may less than 2, and the code _strings.Split(override, ":")[1]_ will panic


pkg/images/images.go
```diff
@@ -51,7 +51,7 @@ func GetImagePath(repo, version, image, component string, imageOverrides map[str
 }
 
 func GetPullPolicy(version, override string) v1.PullPolicy {
-	if len(override) > 0 {
+	if len(override) > 0 && strings.Contains(override, ":") {
```
tpantelis: Just curious - did you observe an actual panic in a running system or did you just notice this on code inspection?

```suggestion
        tags := strings.Split(override, ":")
	if len(tags) > 1 {
	    return getPullPolicy(tags[1])
	}
	
	return getPullPolicy(version)
```

tpantelis (changes requested):

pkg/images/images.go
pengbinbin1: Yes,I have observe an actual panic in my running system,like the below image
微信截图_20220908090401

pkg/images/images.go
```diff
@@ -51,9 +51,11 @@ func GetImagePath(repo, version, image, component string, imageOverrides map[str
...
 
 func GetPullPolicy(version, override string) v1.PullPolicy {
-	if len(override) > 0 {
-		tag := strings.Split(override, ":")[1]
-		return getPullPolicy(tag)
+	if len(override) > 0
+		tags := strings.Split(override, ":")
+		if len(tags) > 1{
+			return getPullPolicy(tags[1])
+		}
```
pengbinbin1: fixed

pkg/images/images.go
```diff
@@ -51,9 +51,11 @@ func GetImagePath(repo, version, image, component string, imageOverrides map[str
 }
 
 func GetPullPolicy(version, override string) v1.PullPolicy {
-	if len(override) > 0 {
-		tag := strings.Split(override, ":")[1]
-		return getPullPolicy(tag)
+	if len(override) > 0
```
tpantelis: You don't need this check - `Split` handles empty string.

```suggestion
```

tpantelis (changes requested):

pkg/images/images.go
pengbinbin1: I think we need this check,if the string is empty then we will unnecessary to split it even though empty string can also be splited

tpantelis: As I mentioned, `Split` handles an empty string and returns an empty array, ie there's nothing to split. This is a bit cleaner than having a redundant check.

pengbinbin1: Done

tpantelis (approved):

skitt (approved):

PR #2237, opened 2022-09-08
Add separate RBAC files for OCP

### LINKED ISSUE #117: Subctl join fails (error updating the privileged SCC) on ROSA cluster
**What happened**:

I tried to deploy Submariner on ROSA cluster using subctl : 

subctl join --kubecontext <xx>  broker-info.subm  --clusterid aws-rosa  --load-balancer

Deployment failed with the following error : 

Join error :

✗ Deploying the Submariner operator 
 ✓ Created operator CRDs
 ✓ Created operator namespace: submariner-operator
 ✓ Created operator service account and role
Error deploying the operator: error updating the SCC resource: error updating OpenShift privileged SCC: admission webhook "scc-validation.managed.openshift.io" denied the request: Modifying default SCCs [anyuid hostaccess hostmount-anyuid hostnetwork node-exporter nonroot privileged restricted] is not allowed

subctl version: v0.12.1

Successful deployment 

see above 

As part of Submariner deployment we update [1]  the privileged SCC users with the Submariner Service Accounts , and in ROSA clusters seems that the admission webhook controller is rejecting any changes to the privileged SCC

I was able to  add user to privileged SCC using [2] command

[1] 
https://github.com/submariner-io/subctl/blob/devel/pkg/scc/update.go#L40 

[2]
oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-gateway

**Environment**:
- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`)
- Cloud provider or hardware configuration:
- Install tools:
- Others:

yboaron:
Add separate RBAC files for Openshift
    
Add separate ClusterRole and ClusterRoleBinding yaml files for OCP.
    
A. The OCP ClusterRole and ClusterRoleBinding yaml files should
    be applied by [1] PR in Subctl Repo.
    
B. After [1] is merged the rest of OCP specific rules should be
    moved to OCP ClusterRole and ClusterRoleBinding files in a
    follow up PR.
    
Fixes: https://github.com/submariner-io/subctl/issues/117
    
[1] https://github.com/submariner-io/subctl/pull/258

Signed-off-by: Yossi Boaron <yboaron@redhat.com>


yboaron:
Ooops, just noticed that pkg/embeddedyamls directory is auto-generated , will update the PR

pkg/embeddedyamls/yamls.go
```diff
@@ -3015,5 +3015,132 @@ roleRef:
   kind: Role
   name: submariner-metrics-reader
   apiGroup: rbac.authorization.k8s.io
+`
```
tpantelis: This file is generated - you need to add yaml files under _config/rbac_ and then run `make pkg/embeddedyamls/yamls.go`.

tpantelis (changes requested):

tpantelis:
You need to also run run `make pkg/embeddedyamls/yamls.go` and check in the modified `pkg/embeddedyamls/yamls.go`.

tpantelis (approved):

config/rbac/kustomization.yaml
```diff
@@ -5,35 +5,44 @@ resources:
   - submariner-operator/role_binding.yaml
   - submariner-operator/cluster_role.yaml
   - submariner-operator/cluster_role_binding.yaml
+  - submariner-operator/privileged_cluster_role.yaml
```
sridhargaddam: We are not removing the existing cluster_role/role_binding yamls. Can you please elaborate the plan?

yboaron: Hmm, good point.
A. We need to keep the existing cluster role to allow submariner component watch/create/etc k8s resources , check [0] for LH agent for example.

B. In addition, to allow privileged permission we need to create a clusterrole , like [1] and for each submariner component have clusterrolebinding resource pointing to this resource like [2] .

Initially, I thought that we can only  create a new clusterrole (like [1]) and add another roleref to existing clusterrolebinding (like 3) , but IIUC we can only define single roleref in clusterrolebinding (check 4).

so, IIUC, we have two options here:
1. Don't create any new files, add new rule (allowing access to privileged scc) in existing clusterrolebinding  of each SM componenet.
2. Create seperate files for SCC permission.

I decided to go with option 2, because didn't want to tight privileged SCC permission to other API permissions, but I'm totally open to new ideas.


[0]
https://github.com/submariner-io/submariner-operator/blob/devel/config/rbac/lighthouse-agent/cluster_role.yaml
[1] 
https://github.com/submariner-io/submariner-operator/blob/6b15cfeedec578cbe18ec2f977000db17175bb79/config/rbac/submariner-operator/privileged_cluster_role.yaml
[2] 
https://github.com/submariner-io/submariner-operator/blob/6b15cfeedec578cbe18ec2f977000db17175bb79/config/rbac/submariner-operator/privileged_cluster_role_binding.yaml

[3] 
https://github.com/submariner-io/submariner-operator/blob/devel/config/rbac/lighthouse-agent/cluster_role_binding.yaml

[4]
https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/rbac#ClusterRoleBinding

yboaron: In option '1' above, it should be 'add new rule (allowing access to privileged scc) in existing clusterrole of each SM componenet' and not clusterrolebinding.

So, for each relevant component under https://github.com/submariner-io/submariner-operator/tree/devel/config/rbac , add a new rule in clusterrole allowing privileged access.

sridhargaddam: >In option '1' above, it should be 'add new rule (allowing access to privileged scc) in existing clusterrole of each SM componenet' and not clusterrolebinding.

Correct. So far AFAICT whenever we require any new permissions for a component, we used to update the existing ClusterRole with the "apiGroups/resources/verbs" and as you said, this does not require any changes to the ClusterRoleBinding. This looks clean and simple. However, at the time of upgrades, the operator has to check (i think it might already be doing it?) not only if the clusterRole exists, but it matches with the clusterRole definition of the new version. If there is any mis-match a new clusterRole has to be installed. 

I did not work much in this area so cannot suggest if we should go with 2 (or 1), but I was personally following 1 so far. Also, the clusterRoles are "additive", so it should not matter if multiple ClusterRoles/RoleBindings are used or a single ClusterRole/RoleBinding is used. Anyways, I'd like to hear from @skitt @vthapar @tpantelis for their thoughts.

yboaron: I haven't worked much in this area either, IIUC, changes are expected in the way to acquire privileged permisisons in the next versions of OCP.
So, maybe it makes sense to keep the separate API  (scc.Update API in subctl) which implies for option '2'.

mkolesnik: FWIW I think we should strive to be as vendor agnostic as possible, so it makes more sense to store any "extra" vendor stuff, such as the openshift specific resources, in separate files, and any generic stuff, such as the K8S RBAC, in the general files we already have.

yboaron: Yes, it makes sense not to mix K8S RBAC with SCC and apply the relevant files only when necessary

vthapar: +1 to Mike's comment. Separate files for specific vendors and single file for cmmon generic. Then update existing ClusterRoles with newer permissions. Add new file if need something specific for a new vendor.

vthapar:
@yboaron Does this depend on https://github.com/submariner-io/subctl/pull/258 or 258 depends on this? Or cna they go in any order?

yboaron:
> @yboaron Does this depend on [submariner-io/subctl#258](https://github.com/submariner-io/subctl/pull/258) or 258 depends on this? Or cna they go in any order?

 [submariner-io/subctl#258](https://github.com/submariner-io/subctl/pull/258) depends on this PR, so we should first merge this PR and then 258 , 258 PR should be updated to use yaml files from this PR if needed (for OCP case)

config/rbac/kustomization.yaml
yboaron: I added separate Openshift RBAC files for each component.

A. The OCP ClusterRole and ClusterRoleBinding yaml files should
be applied by [1] PR in Subctl Repo.

B. After [1] is merged the rest of OCP specific rules should be
moved to OCP ClusterRole and ClusterRoleBinding files in a
follow up PR.

cc @mkolesnik @tpantelis @vthapar @sridhargaddam 

[1] https://github.com/submariner-io/subctl/pull/258

config/rbac/lighthouse-agent/ocp_cluster_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: ocp-submariner-lighthouse-agent
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ocp-submariner-lighthouse-agent
+subjects:
+  - kind: ServiceAccount
+    name: submariner-lighthouse-agent
+    namespace: submariner-operator
```
tpantelis: We shouldn't hardcode `submariner-operator`. Other similar files specify `placeholder` which the generator looks for.

Ditto for the other files.

```suggestion
    namespace: placeholder
```

tpantelis (changes requested):

config/rbac/lighthouse-agent/ocp_cluster_role_binding.yaml
yboaron: Ooops, missed it. thanks.

tpantelis (approved):

mkolesnik (approved):

yboaron:
Please don't merge , I want to to check OCP deployment with this fix

yboaron:
> Please don't merge , I want to to check OCP deployment with this fix

OK, I checked OCP deployment using these RBAC files (instead of scc edit), GW connections status = connected.
I think we can merge this PR.

config/rbac/lighthouse-agent/ocp_cluster_role.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
sridhargaddam, line 3: All the clusterRoles are exactly the same except for the name. So, can we create a single `ClusterRole` and use separate `ClusterRoleBindings` to assign the clusterRole to various components?

CC: @skitt

sridhargaddam (changes requested):

config/rbac/lighthouse-agent/ocp_cluster_role.yaml
yboaron: I created a separate clusterrole per component coz the plan is to move all the ocp relevant permissions to the relevant ClusterRole per component, for example [gateway ocp permissions](https://github.com/submariner-io/submariner-operator/blob/devel/config/rbac/submariner-gateway/cluster_role.yaml#L37-#L52) should move to gateway clusterrole file.

We'll have a follow-up PR to address that.

Having a separate file per component allows flexibility in the future.

yboaron: Maybe we should have a single RBAC file per component for both k8s and ocp and generate the OCP part conditionally on subctl side before applying it. 

Not sure if we can deliver it for 0.14 (maybe we can file an enhancement) , we need to unblock submariner deployments on ROSA and ARO as soon as possible to start testing.

skitt: I think it’s fine to have one file per component, matching the other `ClusterRole` definitions. As Yossi says above, they won’t all stay the same anyway.

skitt (approved):

sridhargaddam (approved):

PR #2239, opened 2022-09-09
Fix globalnet diagnose test

tomer-shefler:
Add the following permissions required by the test: list permission for clusterglobalegressips and globalegressips in API group submariner.io list permission for serviceexports in API group multicluster.x-k8s.io

Signed-off-by: Tomer Shefler <sheflertomer@gmail.com>


pkg/embeddedyamls/yamls.go
```diff
@@ -20,12 +20,13 @@ limitations under the License.
...
 const (
-	Deploy_crds_submariner_io_brokers_yaml = `---
+	Deploy_crds_submariner_io_brokers_yaml = `
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.9.2
+    controller-gen.kubebuilder.io/version: v0.4.1
```
tpantelis: The version here should not have been reverted. That's probably why there's a lot of other extraneous changes in this file. How are you generating this file? Something's out-of-date in your env...

tpantelis (changes requested):

pkg/embeddedyamls/yamls.go
tomer-shefler: Even on a devel branch recently pulled and with changes. Whn I run `go generate pkg/embeddedyamls/generate.go` I get all of these changes. Could this be related to something in mt enviroment? What could it be?

tpantelis: That command runs `pkg/embeddedyamls/generators/yamls2go.go` which merely copies the yaml file contents as is to `yamls.go`. So thats not what's making the changes. The yamls that are modified are all CRDs under `deploy` that are generated by `bin/controller-gen` which is also generated by `$(GO) build -o $@ sigs.k8s.io/controller-tools/cmd/controller-gen` in the Makefile. I believe that relies on the `sigs.k8s.io/controller-tools `version in the `go.mod` which was updated to _v0.9.2_ from _v0.4.1_ a few weeks ago.  I don't know what's going on in your env but make sure your branch is up-to-date and try running `make pkg/embeddedyamls/yamls.go`.

tpantelis (changes requested):

tpantelis (approved):

dfarrell07 (approved):
Is this failing in CI somewhere? If not, it might be good to add coverage that would catch it.

skitt (approved):

PR #2293, opened 2022-10-23
Expose `Nettest{Component,Image}`

mkolesnik:
Expose as generic Nettest image (instead of specific to metrics proxy) so that it can be consumed by other projects (such as `subctl`) and properly overriden in downstream forks.

### DIFF
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -32,6 +32,7 @@ const (
 	ServiceDiscoveryCrName       = "service-discovery"
 	SubmarinerCrName             = "submariner"
 	MetricsProxyComponent        = "submariner-metrics-proxy"
+	NettestComponent             = "submariner-nettest"
 )
 
 /* These values are used by downstream distributions to override the component default image name. */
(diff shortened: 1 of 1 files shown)


skitt:
The use of `nettest` is a convenience for now, the metrics proxy might not continue using it. Conceptually it’s a different component which just happens to use the `nettest` image.

mkolesnik:
> The use of `nettest` is a convenience for now, the metrics proxy might not continue using it. Conceptually it’s a different component which just happens to use the `nettest` image.

The question is do we treat nettest as a component or not?
I think it's already is a sort of component, and in that case while I agree conceptually the metrics proxy is a distinct component, I dont see a benefit of making that separation now (we could always separate them if necessary).

Regardless, I don't mind to add a separate `nettest` component and leave the proxy be, even though they both use the same image.

Thoughts?

skitt:
OK, I hadn’t understood that the important part of this is exposing `nettest` to make it overridable ;-).

I agree we need to make `nettest` overridable, and it is sort of a component (for `subctl` at least in `diagnose` etc.); let’s keep the metrics proxy component separate and add a new `nettest` component.

skitt (approved):

maayanf24 (approved):

PR #2315, opened 2022-10-31
Use the nettest image for the pause container

### LINKED ISSUE #2304: subctl uninstall should use the nettest image to pause
The pause image isn’t available for all the architectures we end up caring about, or it’s available under different names; using an external image also means we end up relying on something we don’t build.

Our `nettest` images include `sleep`, so they can be used as pause images — all that’s required is to run a pod with `sleep infinity`.

tpantelis:
...used for uninstall. The K8s pause image isn’t available for all the architectures we support or has a different name so use the `nettest` image with the `sleep` command instead.

Fixes https://github.com/submariner-io/submariner-operator/issues/2304


skitt (approved):

nyechiel (approved):

sridhargaddam:
@tpantelis @nyechiel should we backport this to release-0.13 as well?

nyechiel:
> @tpantelis @nyechiel should we backport this to release-0.13 as well?

Yes.

Issue #1105, opened 2021-02-23
RBAC should be auto-generated


skitt:
We should generate the RBAC declarations from code (see the `TODO` in `controllers/submariner/broker_controller.go`). The resulting YAML files could also be re-used for the operator bundle, and the RBAC-declaration code in `pkg/broker/rbac.go` would no longer be necessary.

mangelajo:
@skitt @SteveMattar shall this be an item for 0.9.0? sounds risky to me.

skitt:
I’ve got a patch in progress for this, but it is indeed somewhat risky (although the code generates files we can compare with the existing setup, so it’s easy enough to verify). I’ve moved it to 0.10.

SteveMattar:
I agree let's wait with this

SteveMattar:
The kustomize part is handled in this patch: https://github.com/submariner-io/submariner-operator/pull/1375

dfarrell07:
@skitt Do you have any pointers about how this should be done?

skitt:
The relevant section in the Operator SDK documentation is https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#restricting-roles-and-permissions

The kubebuilder reference is https://book.kubebuilder.io/reference/markers/rbac.html

See https://github.com/skitt/submariner-operator/pull/16 where I started working on this a while back — my process then was to compare the generated RBAC with the RBAC we had in our YAML files, determine whether a missing permission was significant, and add the corresponding annotation.

dfarrell07:
Relates to https://github.com/submariner-io/submariner-operator/issues/1240 and https://github.com/submariner-io/submariner-operator/issues/1241

skitt:
See https://github.com/operator-framework/operator-sdk/issues/6100 for namespace permissions instead of cluster permissions.

dfarrell07:
I should stop beating my head against this and ask...what about the [non-operator roles](https://github.com/submariner-io/submariner-operator/tree/devel/config/rbac)? I went through some ideas that turned out obviously-in-retrospect bad, like annotating some calls with extra permissions to try to cover them or cloning the other repos during generation.

dfarrell07:
[...]

PR #2336, opened 2022-11-10
Remove duplicate pod RBAC

dfarrell07:
The RBAC for pods duplicates the get permission, as it's granted * elsewhere.

This was recently modified in #2225, #2214, and #2008.

### DIFF
--- a/config/rbac/submariner-gateway/role.yaml
+++ b/config/rbac/submariner-gateway/role.yaml
@@ -41,12 +41,6 @@ rules:
     verbs:
       - update
   - apiGroups:
-      - ""
-    resources:
-      - pods
-    verbs:
-      - get
-  - apiGroups:
       - apps
     resources:
       - replicasets
--- a/config/rbac/submariner-globalnet/role.yaml
+++ b/config/rbac/submariner-globalnet/role.yaml
@@ -43,12 +43,6 @@ rules:
     verbs:
       - update
   - apiGroups:
-      - ""
-    resources:
-      - pods
-    verbs:
-      - get
-  - apiGroups:
       - apps
     resources:
       - replicasets
--- a/config/rbac/submariner-operator/role.yaml
+++ b/config/rbac/submariner-operator/role.yaml
@@ -43,12 +43,6 @@ rules:
     verbs:
       - update
   - apiGroups:
-      - ""
-    resources:
-      - pods
-    verbs:
-      - get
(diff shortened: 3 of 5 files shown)


dfarrell07:
@skitt How did you determine in #2008 that pods didn't need some permissions? Just looking at the code?

@tpantelis How did you spot the error in #2225, as it didn't seem to fail CI?

I ask because I don't know if pods really needs *, or if we can get away with a subset.

tpantelis:
> How did you spot the error in https://github.com/submariner-io/submariner-operator/pull/2225, as it didn't seem to fail CI?

It did fail CI according to the PR summary.

dfarrell07:
> > How did you spot the error in #2225, as it didn't seem to fail CI?
> 
> It did fail CI according to the PR summary.

Ah okay, I see. I actually linked to the wrong PR there, this is the gateway which was PR #2214. But still, the point is that it should have failed CI on the PR that made the change, not have been merged.

tpantelis:
> 
> Ah okay, I see. I actually linked to the wrong PR there, this is the gateway which was PR #2214. But still, the point is that it should have failed CI on the PR that made the change, not have been merged.

It changed the RBAC yaml and the associated Go file but those are used by `subctl`  so they won't take effect until the Go dependency version for **submariner-operator** is updated in the **subctl** repo. This is a side-effect of splitting out the **subctl** repo.

dfarrell07:
> > Ah okay, I see. I actually linked to the wrong PR there, this is the gateway which was PR #2214. But still, the point is that it should have failed CI on the PR that made the change, not have been merged.
> 
> It changed the RBAC yaml and the associated Go file but those are used by `subctl` so they won't take effect until the Go dependency version for **submariner-operator** is updated in the **subctl** repo. This is a side-effect of splitting out the **subctl** repo.

Thanks for clarifying Tom, that makes sense.

tpantelis (approved):

sridhargaddam (approved):

skitt (changes requested):
I would rather move in the other direction: instead of removing explicit permissions which are subsumed by wildcard permissions, remove the wildcard permissions and explicitly list all the permissions we really need (ideally, auto-generated from code comments).

dfarrell07:
Closing in favor of https://github.com/submariner-io/submariner-operator/issues/1105

PR #2337, opened 2022-11-11
Remove servicediscoveries RBAC, covered by *

dfarrell07:
All resources in the API group submariner.io are already granted all verbs. It shouldn't be necessary to explicitly list servicediscoveries.


dfarrell07:
I couldn't find anything about this, but one risk I can imagine is if CRDs aren't covered by * for some reason.

tpantelis (approved):

skitt (approved):
OK for this one since the unnecessary resource is listed right alongside `*`.

PR #2401, opened 2022-12-12
Stop pretending to support Kubernetes 1.17

skitt:
At some point we lost the ability to test against Kubernetes 1.17; let's stop pretending to support it (it reached EOL on 2021-01-13).

### DIFF
--- a/.github/workflows/e2e-all-k8s.yml
+++ b/.github/workflows/e2e-all-k8s.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
-        k8s_version: ['1.17', '1.22', '1.23', '1.24', '1.25']
+        k8s_version: ['1.19', '1.22', '1.23', '1.24', '1.25']
         lighthouse: ['', 'lighthouse']
         ovn: ['', 'ovn']
         exclude:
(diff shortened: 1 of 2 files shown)


Jaanki (approved):

.github/workflows/e2e-all-k8s.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
-        k8s_version: ['1.17', '1.22', '1.23', '1.24', '1.25']
```
sridhargaddam, line 22: Rather than removing the 1.17 version, can we change it to 1.18?

sridhargaddam (approved):

sridhargaddam:
@skitt should we also update this? - https://github.com/submariner-io/subctl/blob/91a438f8e461da2536e966771fffb4319cc7d08c/pkg/version/version.go#L35

nyechiel (approved):

skitt:
> @skitt should we also update this? - https://github.com/submariner-io/subctl/blob/91a438f8e461da2536e966771fffb4319cc7d08c/pkg/version/version.go#L35

Yes, see https://github.com/submariner-io/subctl/pull/446

tpantelis (approved):

PR #2412, opened 2022-12-16
Remove replace for sigs.k8s.io/controller-runtime

tpantelis:
It's been bumped to 0.14.1 which provides K8s v0.26 support so we can remove the replace in the go.mod.


skitt (approved):

dfarrell07 (approved):

go.mod
```diff
@@ -29,9 +29,6 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
-// Until controller-runtime releases with v0.26 support
```
skitt, line 32: Please update the commit message to match what this commit is doing now (dropping the replace, not upgrading to controller-runtime 0.14.0).

skitt (changes requested):

skitt (approved):

PR #2425, opened 2022-12-19
Run the upgrade job only on devel

skitt:
This job doesn't work on any branch other than devel: it always downloads the latest release, and tries to "upgrade" from there, whereas on non-devel branches it would have to download the last release from the branch preceding the target branch (e.g. the last 0.13.x on the release-0.14 branch).


mkolesnik:
What if we always test upgrade from latest built to current PR, shouldn't that be good enough?

We could use eg https://github.com/submariner-io/shipyard/pull/1088 and basically still have upgrade tests which should also work properly for stable branches.

skitt:
> What if we always test upgrade from latest built to current PR, shouldn't that be good enough?
> 
> We could use eg [submariner-io/shipyard#1088](https://github.com/submariner-io/shipyard/pull/1088) and basically still have upgrade tests which should also work properly for stable branches.

Not really, using the base branch release would only test across one PR (from the latest commit on the target branch to the PR). Ideally we want this:

* on `devel`, test from the latest release to the current PR
* on release branches, test from the latest release on the branch to the current PR

Perhaps our definition of `subctl-release-foo` is wrong, and we shouldn’t bump it on every successful merge but only on actual releases.

mkolesnik:
> Not really, using the base branch release would only test across one PR (from the latest commit on the target branch to the PR).

So my line of thought is that for the regular PR CI this is quite fine, as cumulatively we would've tested that we get from point A (last known release) to point B (this PR) by going over all the points along the way.

As the saying goes "don't let perfect be the enemy of good enough" :)

I think for the latest released version to the latest branch-tip could be a timed job, and/or a job on releases (same as we test combined E2E on each release).
This should also allow us to test stable to stable also.

skitt:
> > Not really, using the base branch release would only test across one PR (from the latest commit on the target branch to the PR).
> 
> So my line of thought is that for the regular PR CI this is quite fine, as cumulatively we would've tested that we get from point A (last known release) to point B (this PR) by going over all the points along the way.

I disagree, we really do want to test upgrades from the last release to the PR being checked in one go, not the result of applying all PRs in the interim.

> As the saying goes "don't let perfect be the enemy of good enough" :)

That’s exactly the thinking behind this PR: we have a good test on `devel`, it’s broken and misleading on other branches, so let’s keep the `devel` test and disable upgrade tests on the other branches. We can set up proper upgrade tests on release branches as another PR.

> I think for the latest released version to the latest branch-tip could be a timed job, and/or a job on releases (same as we test combined E2E on each release). This should also allow us to test stable to stable also.

I much prefer PR checks when possible, the errors they catch have a much better chance of being addressed.

tpantelis (approved):

dfarrell07 (approved):

PR #2468, opened 2023-01-23
Remove unused BrokerCR name

Jaanki:
Epic: https://github.com/submariner-io/enhancements/issues/143


pkg/names/names.go
```diff
@@ -33,7 +33,7 @@ const (
 	SubmarinerCrName             = "submariner"
 	MetricsProxyComponent        = "submariner-metrics-proxy"
 	NettestComponent             = "submariner-nettest"
-	BrokerCrName                 = "broker"
+	BrokerCrName                 = "submariner-broker"
```
tpantelis: This isn't used in this project and neither currently in **subctl**.  I assume you intend to use it in **subctl** but **subctl** already defines a [constant](https://github.com/submariner-io/subctl/blob/d798bb839f3d49f522f83dd5830d4743becde286/pkg/brokercr/ensure.go#L33) for this. So I suggest to use that one and remove this one.

tpantelis (changes requested):

pkg/names/names.go
Jaanki: yupe. Thats what I was looking for. Then I guess I will remove the whole constant (L36) from here.

skitt (approved):

tpantelis (approved):

PR #2480, opened 2023-01-30
Reduce dependabot frequency, target k8s bumps

tpantelis:
- Check monthly for gh-actions
- Ignore k8s dependencies included with apiextensions-apiserver


dfarrell07 (approved):

skitt (approved):

skitt:
The new commit title isn’t great :-/. I’m fine with merging this as-is, but “Reduce dependabot frequency, target k8s bumps” would be more informative for the short log.

PR #2503, opened 2023-02-15
Stop filling in token/CA in the Submariner CR

### LINKED ISSUE #1869: Stop publishing sensitive information directly in CRs
Now that the PSK and broker token are stored in secrets, we should stop publishing them directly in CRs.

This won’t really improve security, but it will make CR dumps easier to read.

This follows on from https://github.com/submariner-io/submariner/issues/326.

Depends on https://github.com/submariner-io/subctl/issues/631

skitt:
This ensures we rely only on the secret containing the broker access token.

Fixes: #1869

Depends on https://github.com/submariner-io/releases/issues/644
Depends on https://github.com/submariner-io/subctl/issues/631


tpantelis (approved):

mkolesnik (approved):

dfarrell07:
@skitt Why does this depend on RC0? Should it be the other way around, or am I missing something about why an API change should be after the RC?

skitt:
> @skitt Why does this depend on RC0? Should it be the other way around, or am I missing something about why an API change should be after the RC?

This depends on rc0 so that it isn’t included in 0.15. It will be merged to `devel` after we cut the 0.15 branch, so it will target 0.16 — I want to make sure it gets as much testing as possible before it gets released.

nyechiel:
@skitt is it the right time to merge this and move to the 0.16 board?

skitt:
> @skitt is it the right time to merge this and move to the 0.16 board?

This will break `recover-broker-info`, so we need to address that first (https://github.com/submariner-io/subctl/issues/631).

nyechiel:
> > @skitt is it the right time to merge this and move to the 0.16 board?
> 
> This will break `recover-broker-info`, so we need to address that first ([submariner-io/subctl#631](https://github.com/submariner-io/subctl/issues/631)).

Is this planned for 0.16? It seems to be it in the backlog board at the moment

skitt:
> > > @skitt is it the right time to merge this and move to the 0.16 board?
> > 
> > 
> > This will break `recover-broker-info`, so we need to address that first ([submariner-io/subctl#631](https://github.com/submariner-io/subctl/issues/631)).
> 
> Is this planned for 0.16? It seems to be it in the backlog board at the moment

We didn’t finish planning 0.16 yesterday...

tpantelis:
Is this still relevant?

PR #2529, opened 2023-03-06
Fix bundle generation

skitt:
See individual commits for details.


tpantelis (approved):

vthapar (approved):

config/manifests/bases/submariner.clusterserviceversion.yaml
```diff
@@ -17,7 +17,304 @@ metadata:
...
+    - description: Submariner is the Schema for the submariners API.
+      displayName: Submariner
+      kind: Submariner
+      name: submariners.submariner.io
+      resources:
+      - kind: Deployment
+        name: submariner-operator
+        version: v1
+      specDescriptors:
+      - description: Type of broker (must be "k8s").
```
nyechiel, line 90: @skitt didn't we drop this one?

skitt: Not yet, I’ve started removing uses of the field, we need to remove all of those before we can drop it here.

PR #2530, opened 2023-03-08
Add config field for nodePlacement

vthapar:
Epic https://github.com/submariner-io/enhancements/issues/149

Refer https://github.com/submariner-io/enhancements/pull/156


api/v1alpha1/servicediscovery_types.go
```diff
@@ -47,6 +48,10 @@ type ServiceDiscoverySpec struct {
 	// +listType=set
 	CustomDomains  []string          `json:"customDomains,omitempty"`
 	ImageOverrides map[string]string `json:"imageOverrides,omitempty"`
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
skitt, line 52: Shouldn’t this be

```suggestion
	NodeSelector v1.NodeSelector `json:"nodeSelector,omitempty"`
```

?

skitt (changes requested):

api/v1alpha1/servicediscovery_types.go
vthapar: I don't think there is a type `NodeSelector`. Even `PodSpec` uses `map[string]string`

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3273

```
	// NodeSelector is a selector which must be true for the pod to fit on a node.
	// Selector which must match a node's labels for the pod to be scheduled on that node.
	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
	// +optional
	// +mapType=atomic
	NodeSelector map[string]string `json:"nodeSelector,omitempty" protobuf:"bytes,7,rep,name=nodeSelector"`
```

skitt: There is a `NodeSelector` type, hence my question: https://pkg.go.dev/k8s.io/api/core/v1#NodeSelector. I’ve seen it used in other specs, but it doesn’t really seem to be appropriate for that; let’s follow `PodSpec`.

skitt (approved):

dfarrell07 (approved):

PR #2552, opened 2023-03-29
Sycn Broker client and Admin RBACs with subctl

Jaanki:
subctl uses its own Broker RBAC permissions. New permissions are recently added in subctl. Sync those permissions in operator repo as well because these are used by ACM or other third parties.

Depends on https://github.com/submariner-io/subctl/pull/621

### DIFF
--- a/config/broker/broker-admin/role.yaml
+++ b/config/broker/broker-admin/role.yaml
@@ -18,6 +18,13 @@ rules:
       - update
       - delete
   - apiGroups:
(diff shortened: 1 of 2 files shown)


Jaanki:
Raised this issue https://github.com/submariner-io/submariner-operator/issues/2553 to avoid such problems on the future.

tpantelis (approved):

skitt (approved):

PR #2555, opened 2023-03-31
Embed Broker admin, client RBACs

### LINKED ISSUE #2553: Embedded Broker client and admin RBAC yamls
Currently the Broker client and admin RBAC yamls (under `config/broker/broker-admin` and `config/broker/broker-client`) are not part of the logic that generates the embedded yamls and hence not used  by `subctl`. Instead `subctl` has its own version of these RBAC permissions.

Include these RBAC yamls as embedded yamls so that `subctl` can use them.

This also means that there won't be duplicacy of these permissions and singular maintenance point.

Jaanki:
Currently the Broker client and admin RBAC yamls (under config/broker/broker-admin and config/broker/broker-client) are not part of the logic that generates the embedded yamls and hence not used by subctl. Instead subctl has its own version of these RBAC permissions.

Include these RBAC yamls as embedded yamls so that subctl can use them.

This also means that there won't be duplicacy of these permissions and singular maintenance point.

Closes: https://github.com/submariner-io/submariner-operator/issues/2553


tpantelis (changes requested):
Based on the current structure, it seems _config/broker/kustomization.yaml_  should be merged into _config/rbac/kustomization.yaml_.  Also it seems _config/broker/kustomizeconfig.yaml_ should be in _config/rbac_ but there already is such a file that looks identical.  

Similarly _config/broker/namespace.yaml_  should be in _config/rbac_ as it's referenced by the _kustomization.yaml_.

Jaanki:
> Based on the current structure, it seems _config/broker/kustomization.yaml_ should be merged into _config/rbac/kustomization.yaml_. Also it seems _config/broker/kustomizeconfig.yaml_ should be in _config/rbac_ but there already is such a file that looks identical.
> 
> Similarly _config/broker/namespace.yaml_ should be in _config/rbac_ as it's referenced by the _kustomization.yaml_.

To get the ball rolling, I am not moving the `config/broker/*` under `config/rbac/*`

tpantelis (dismissed):

config/rbac/broker/kustomization.yaml
Jaanki, line 1: This file creates all Broker RBACs in `submariner-k8s-broker` namespace as defined by `namespace` and `resources` fields. `config/rbac/kustomization.yaml` will create all RBACs mentioned under `resources` in [`submariner-operator` namespace](https://github.com/submariner-io/submariner-operator/blob/devel/config/default/kustomization.yaml#L3) as specified by [default kustomization.yaml](https://github.com/submariner-io/submariner-operator/blob/devel/config/default/kustomization.yaml#L19). Since `kustomize` does not allow specifying 2 `resources` entries in a single `kustomization.yaml`, 2 `kustomization.yaml` files are written.

tpantelis (approved):

sridhargaddam (approved):

PR #2557, opened 2023-03-31
Add permissions on Globalnet status subresources

skitt:
Now that Admiral handles these separately, we need to give permissions on them explicitly.

(cherry picked from commit d60c75da75dbb4468ca31209ca9ab5458349feef)

### DIFF
--- a/config/rbac/submariner-globalnet/cluster_role.yaml
+++ b/config/rbac/submariner-globalnet/cluster_role.yaml
@@ -41,8 +41,11 @@ rules:
       - submariner.io
     resources:
       - clusterglobalegressips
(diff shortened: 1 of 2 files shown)


dfarrell07 (approved):

tpantelis (changes requested):
We may not need this if we can instruct the resource syncer whether or not the resource type has the status subresource.

tpantelis (approved):

tpantelis:
> We may not need this if we can instruct the resource syncer whether or not the resource type has the status subresource.

With https://github.com/submariner-io/submariner-operator/pull/2561, we will need this.

PR #2602, opened 2023-04-19
Verify broker-admin secret in system tests

Jaanki:
...instead of broker-client secret as is currently done because with https://github.com/submariner-io/subctl/pull/653, subctl no longer creates a broker client SA and its associated secret.

### DIFF
--- a/scripts/test/system.sh
+++ b/scripts/test/system.sh
@@ -13,7 +13,7 @@ function verify_subm_gateway_label() {
 
 function broker_vars() {
     SUBMARINER_BROKER_URL=$(kubectl -n default get endpoints kubernetes -o jsonpath="{.subsets[0].addresses[0].ip}:{.subsets[0].ports[?(@.name=='https')].port}")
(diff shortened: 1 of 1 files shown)


skitt (approved):

skitt (dismissed):
You also need to change line 16.

I’m wondering whether we shouldn’t just drop this test — it retrieves the same `ca.crt` twice and compares the result, which isn’t useful. Alternatively, we can get the admin token in `SUBMARINER_BROKER_CA`, then compare each `cluster1` client token.

tpantelis (approved):

tpantelis:
Since it's `subctl` that creates these secrets, perhaps these tests should be in the **subctl** project.

sridhargaddam (approved):

nyechiel:
Do we need this in 0.15? I am not sure why it's part of the 0.16 board

tpantelis:
> Do we need this in 0.15?

No this is related to https://github.com/submariner-io/subctl/pull/653.

PR #2639, opened 2023-05-15
Disable SC2317 on an indirect function

skitt:
Shellcheck trips over verify_subm_cr_status_ which is defined inside verify_subm_cr_status_with_retries and is only called indirectly. Disable this detection for that function.


dfarrell07 (approved):

tpantelis (approved):

skitt:
There’s a catch-22 here — we need both the updated docker dependency (to pass the vulnerability scanning) and the linting fix. So I’ve updated #2638 with the latter.

PR #2726, opened 2023-07-12
OVNKubernetes CNI IC deployment support

### LINKED ISSUE #191: Update the Submariner-operator to remove network-pulgin syncer deployment
Update the Submariner-operator to remove network-pulgin syncer deployment

aswinsuryan:
With OVN IC CNI 3 kind of deployment are possible

1)SIngle Zone (Or in non-ic deployment)

This is similar to the current deployment. Here a OVN Kubernetes db service will be running and it will be passed by submariner-operator to the routeagent/network plugin syncer code.

2)MultiZone and more than one node per zone.

In this case there will be no OVN db service will be present. But there will be an endpoint per zone. Submariner operator will pass the comma separated list of IP Port and protocol and the endpoint name to the route agent and networkpluginsyncer. The pod will chose one of them depending on the zone it is running

3)Multizone , one node per zone.

For this deployment there will be no service or endpoints. Submariner-Operator will pass the value local. The pods needs to use socket connection to program ovn.

Fixes: https://github.com/submariner-io/enhancements/issues/191

This PR also passed the NBDB and SBDB values to the submariner-route agent daemon set as route agent will be now responsible for programming the OVN routes and policies.

### DIFF
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -135,20 +135,18 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, clusterNetwork *network.Clu
 		},
 	}
 
-	if clusterNetwork.PluginSettings != nil {
-		if ovndb, ok := clusterNetwork.PluginSettings[network.OvnNBDB]; ok {
-			ds.Spec.Template.Spec.Containers[0].Env = append(
-				ds.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
-					Name: network.OvnNBDB, Value: ovndb,
-				})
-		}
+	if ovndb, ok := clusterNetwork.PluginSettings[network.OvnNBDB]; ok {
+		ds.Spec.Template.Spec.Containers[0].Env = append(
+			ds.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+				Name: network.OvnNBDB, Value: ovndb,
+			})
+	}
 
-		if ovnsb, ok := clusterNetwork.PluginSettings[network.OvnSBDB]; ok {
-			ds.Spec.Template.Spec.Containers[0].Env = append(
-				ds.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
-					Name: network.OvnSBDB, Value: ovnsb,
-				})
-		}
+	if ovnsb, ok := clusterNetwork.PluginSettings[network.OvnSBDB]; ok {
+		ds.Spec.Template.Spec.Containers[0].Env = append(
+			ds.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+				Name: network.OvnSBDB, Value: ovnsb,
+			})
 	}
 
 	return ds
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
 	"github.com/submariner-io/submariner-operator/controllers/apply"
+	"github.com/submariner-io/submariner-operator/pkg/discovery/network"
 	"github.com/submariner-io/submariner-operator/pkg/images"
 	"github.com/submariner-io/submariner-operator/pkg/names"
 	appsv1 "k8s.io/api/apps/v1"
@@ -35,13 +36,14 @@ import (
 )
 
 //nolint:wrapcheck // No need to wrap errors here.
-func (r *Reconciler) reconcileRouteagentDaemonSet(ctx context.Context, instance *v1alpha1.Submariner, reqLogger logr.Logger,
+func (r *Reconciler) reconcileRouteagentDaemonSet(ctx context.Context, instance *v1alpha1.Submariner,
+	clusterNetwork *network.ClusterNetwork, reqLogger logr.Logger,
 ) (*appsv1.DaemonSet, error) {
-	return apply.DaemonSet(ctx, instance, newRouteAgentDaemonSet(instance, names.RouteAgentComponent), reqLogger, r.config.ScopedClient,
-		r.config.Scheme)
+	return apply.DaemonSet(ctx, instance, newRouteAgentDaemonSet(instance, clusterNetwork, names.RouteAgentComponent),
+		reqLogger, r.config.ScopedClient, r.config.Scheme)
 }
 
-func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.DaemonSet {
+func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, clusterNetwork *network.ClusterNetwork, name string) *appsv1.DaemonSet {
 	labels := map[string]string{
 		"app":       name,
 		"component": "routeagent",
@@ -49,7 +51,7 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 
 	maxUnavailable := intstr.FromString("100%")
 
-	return &appsv1.DaemonSet{
+	ds := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cr.Namespace,
 			Name:      name,
@@ -83,6 +85,9 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 						{Name: "host-sys", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
 							Path: "/sys",
 						}}},
+						{Name: "host-var-run-openvswitch-nbdb-sock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
+							Path: "/var/run/openvswitch/ovnnb_db.sock",
+						}}},
 					},
 					Containers: []corev1.Container{
 						{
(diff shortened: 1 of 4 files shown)


pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,117 @@ const (
...
+		PluginSettings: map[string]string{
+			OvnNBDB: "local",
+			OvnSBDB: "local",
+		},
+	}
+}
+
+func createClusterNetworkWithEndpoints(endPoints []corev1.Endpoints) *ClusterNetwork {
+	pluginSettings := map[string]string{}
+	var OvnNBDBIPs, OVNSBDBIps string
```
skitt: The names should be consistently cased and made more distinguishable, *e.g.*

```suggestion
	var northboundDBIPs, southboundDBIPs string
```

(this is all about OVN so the prefix isn’t all that useful).

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,117 @@ const (
...
+
+	for index := 0; index < len(endPoints); index++ {
+		for _, subset := range endPoints[index].Subsets {
+			for _, port := range subset.Ports {
+				if strings.Contains(port.Name, "north") {
+					OvnNBDBIPs += fmt.Sprintf("%s:%s:%s:%s:%d,",
```
skitt: The constant part should be included in the format string:

```suggestion
					OvnNBDBIPs += fmt.Sprintf("IC::%s:%s:%s:%d,",
```

This ends up with two colons after “IC”, is that intended?

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,117 @@ const (
...
+						"IC:", endPoints[index].Name, port.Protocol, subset.Addresses[0].IP, OvnNBDBDefaultPort)
+				} else if strings.Contains(port.Name, "south") {
+					OVNSBDBIps += fmt.Sprintf("%s:%s:%s:%s:%d,",
```
skitt: Same as above.

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -82,6 +160,31 @@ func discoverOvnKubernetesNetwork(ctx context.Context, client controllerClient.C
 			clusterNetwork.ServiceCIDRs = []string{svcCidr}
 		}
 	}
+}
 
-	return clusterNetwork, nil
+func FindEndpoint(ctx context.Context, client controllerClient.Client, endpointNameSpace string) (*corev1.EndpointsList, error) {
```
skitt: Does this need to be public?

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Done,

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Yes, I thought of splitting it with ":" when we later consume it in submariner.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: No, changed it to private.

pkg/discovery/network/ovnkubernetes.go
skitt: But do we need two colons? Wouldn’t `IC:%s:%s:%s:%d,` be sufficient?

aswinsuryan: Oh, that was an oversight, fixed it. Thanks

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+func createClusterNetworkWithEndpoints(endPoints []corev1.Endpoints) *ClusterNetwork {
+	pluginSettings := map[string]string{}
+	var northboundDBIPs string
+
+	for index := 0; index < len(endPoints); index++ {
+		for _, subset := range endPoints[index].Subsets {
+			for _, port := range subset.Ports {
+				if strings.Contains(port.Name, "north") {
+					northboundDBIPs += fmt.Sprintf("%s:%s:%s:%s:%d,",
+						"IC", endPoints[index].Name, port.Protocol, subset.Addresses[0].IP, OvnNBDBDefaultPort)
```
skitt:
```suggestion
					northboundDBIPs += fmt.Sprintf("IC:%s:%s:%s:%d,",
						endPoints[index].Name, port.Protocol, subset.Addresses[0].IP, OvnNBDBDefaultPort)
```

skitt (approved):
Minor `Printf` tweak.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Done.

controllers/submariner/route_agent_resources.go
```diff
@@ -128,4 +134,22 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 			},
 		},
 	}
+
+	if clusterNetwork.PluginSettings != nil {
```
tpantelis: Minor thing but the nil check is unnecessary as Go handles nil map when accessing an element via index.

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+func discoverOvnNodeClusterNetwork(ctx context.Context, client controllerClient.Client) (*ClusterNetwork, error) {
+	// In OVN IC deployments, the ovn DB will be a part of ovnkube-node
+	ovnPod, err := FindPod(ctx, client, "name=ovnkube-node")
+	if err != nil || ovnPod == nil {
+		return nil, err
+	}
+
+	endpointList, err := findEndpoint(ctx, client, ovnPod.Namespace)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error retrieving the endpoints from namespace %q", ovnPod.Namespace)
```
tpantelis:
```suggestion
		return nil, errors.Wrapf(err, "error retrieving the endpoints from namespace %q", ovnPod.Namespace)
```

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+
+	if endpointList == nil || len(endpointList.Items) == 0 {
+		clusterNetwork, err = createLocalClusterNetwork(), nil
+	} else {
+		clusterNetwork, err = createClusterNetworkWithEndpoints(endpointList.Items), nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
tpantelis: `err` is always nil here.

```suggestion
	if endpointList == nil || len(endpointList.Items) == 0 {
		clusterNetwork = createLocalClusterNetwork()
	} else {
		clusterNetwork = createClusterNetworkWithEndpoints(endpointList.Items)
	}
```

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+			OvnNBDB: "unix:/var/run/openvswitch/ovnnb_db.sock",
+		},
+	}
+}
+
+func createClusterNetworkWithEndpoints(endPoints []corev1.Endpoints) *ClusterNetwork {
+	pluginSettings := map[string]string{}
+	var northboundDBIPs string
+
+	for index := 0; index < len(endPoints); index++ {
```
tpantelis:
```suggestion
	for index := range endPoints {
```

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+				}
+			}
+		}
+	}
+
+	pluginSettings[OvnNBDB] = northboundDBIPs
+
+	return &ClusterNetwork{
+		PluginSettings: pluginSettings,
+	}
```
tpantelis: nit - you can set `PluginSettings` inline.

```suggestion
	return &ClusterNetwork{
		PluginSettings: map[string]string{OvnNBDB, northboundDBIPs},
	}
```

tpantelis: What if `northboundDBIPs` is empty.... Should we still set `OvnNBDB` in the map? Is this an error condition?

pkg/discovery/network/ovnkubernetes.go
```diff
@@ -39,40 +40,112 @@ const (
...
+	return &ClusterNetwork{
+		PluginSettings: map[string]string{
+			OvnNBDB: "unix:/var/run/openvswitch/ovnnb_db.sock",
+		},
+	}
+}
+
+func createClusterNetworkWithEndpoints(endPoints []corev1.Endpoints) *ClusterNetwork {
+	pluginSettings := map[string]string{}
+	var northboundDBIPs string
```
tpantelis: I think we should make this an array and use `strings.Join(northboundDBIPs, ",")` afterwards. This avoids repeated concatenation and the trailing comma.

tpantelis (changes requested):

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: done.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Done.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Done.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: Done.

controllers/submariner/route_agent_resources.go
aswinsuryan: Done.

pkg/discovery/network/ovnkubernetes.go
aswinsuryan: if endpoint is present there should be a NorthBound and SouthBound IP , else it is an invalid ovn configurartion.

tpantelis (approved):
Need to fix compile errors but otherwise looks good.

PR #2727, opened 2023-07-13
Provision GW LB service in accordance with OCP platform

### LINKED ISSUE #2603: LoadBalancer: 'service.beta.kubernetes.io/aws-load-balancer-type: nlb' annotation exists regardless of platform type
When Submariner gateway is deployed using LoadBalancer service, 'service.beta.kubernetes.io/aws-load-balancer-type: nlb' annotation exists regardless of platform type.

For example [1]  is the submariner-gateway manifest for when Submariner deployed on ARO.

This should be treated as low severity, as it should not affect the LoadBalancer service, but only may cause confusion.

[1] 

kind: Service
apiVersion: v1
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
  resourceVersion: '622004'
  name: submariner-gateway
  uid: c3dca72e-bf5b-46ed-bed5-14cfb02b90d8
  creationTimestamp: '2023-04-19T13:49:20Z'
  managedFields:
    - manager: kube-controller-manager
      operation: Update
      apiVersion: v1
      time: '2023-04-19T13:49:20Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:finalizers':
            .: {}
            'v:"service.kubernetes.io/load-balancer-cleanup"': {}
      subresource: status
    - manager: submariner-operator
      operation: Update
      apiVersion: v1
      time: '2023-04-19T13:49:20Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:annotations':
            .: {}
            'f:service.beta.kubernetes.io/aws-load-balancer-type': {}
          'f:ownerReferences':
            .: {}
            'k:{"uid":"34d24fcc-6ec8-4bb9-b146-e6c531b46e27"}': {}
        'f:spec':
          'f:allocateLoadBalancerNodePorts': {}
          'f:externalTrafficPolicy': {}
          'f:internalTrafficPolicy': {}
          'f:ports':
            .: {}
            'k:{"port":4490,"protocol":"UDP"}':
              .: {}
              'f:name': {}
              'f:port': {}
              'f:protocol': {}
              'f:targetPort': {}
            'k:{"port":4500,"protocol":"UDP"}':
              .: {}
              'f:name': {}
              'f:port': {}
              'f:protocol': {}
              'f:targetPort': {}
          'f:selector': {}
          'f:sessionAffinity': {}
          'f:type': {}
  namespace: submariner-operator
  ownerReferences:
    - apiVersion: submariner.io/v1alpha1
      kind: Submariner
      name: submariner
      uid: 34d24fcc-6ec8-4bb9-b146-e6c531b46e27
      controller: true
      blockOwnerDeletion: true
  finalizers:
    - service.kubernetes.io/load-balancer-cleanup
spec:
  clusterIP: 172.31.238.232
  externalTrafficPolicy: Local
  ipFamilies:
    - IPv4
  healthCheckNodePort: 32488
  ports:
    - name: cable-encaps
      protocol: UDP
      port: 4500
      targetPort: 4500
      nodePort: 31778
    - name: natt-discovery
      protocol: UDP
      port: 4490
      targetPort: 4490
      nodePort: 30513
  internalTrafficPolicy: Cluster
  clusterIPs:
    - 172.31.238.232
  allocateLoadBalancerNodePorts: true
  type: LoadBalancer
  ipFamilyPolicy: SingleStack
  sessionAffinity: None
  selector:
    app: submariner-gateway
    gateway.submariner.io/status: active
status:
  loadBalancer: {}

yboaron:
A LoadBalancer type Service relies on the cloud provider to create an external load balancer with an IP address in the relevant network space.

Different cloud providers support different Service annotations, this PR updates LB service annotations in accordance with OCP platform.

Fixes: https://github.com/submariner-io/submariner-operator/issues/2603

### DIFF
--- a/config/rbac/submariner-operator/ocp_cluster_role.yaml
+++ b/config/rbac/submariner-operator/ocp_cluster_role.yaml
@@ -12,3 +12,9 @@ rules:
       - securitycontextconstraints
     verbs:
       - use
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - infrastructures
+    verbs:
+      - get
--- a/controllers/submariner/loadbalancer_resources.go
+++ b/controllers/submariner/loadbalancer_resources.go
@@ -20,15 +20,20 @@ package submariner
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/pkg/errors"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
 	"github.com/submariner-io/submariner-operator/controllers/apply"
 	"github.com/submariner-io/submariner-operator/pkg/names"
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/port"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
@@ -43,18 +48,74 @@ const (
 func (r *Reconciler) reconcileLoadBalancer(
 	ctx context.Context, instance *v1alpha1.Submariner, reqLogger logr.Logger,
 ) (*corev1.Service, error) {
-	return apply.Service(ctx, instance, newLoadBalancerService(instance), reqLogger, r.config.ScopedClient, r.config.Scheme)
+	var err error
+
+	platformTypeOCP, err := r.getOCPPlatformType(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	svc := newLoadBalancerService(instance, platformTypeOCP)
+	svc, err = apply.Service(ctx, instance, svc, reqLogger, r.config.ScopedClient, r.config.Scheme)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// For IBM cloud also needs to annotate the allocated health check node port
+	if platformTypeOCP == string(configv1.IBMCloudPlatformType) {
+		healthPortStr := fmt.Sprintf("%d", svc.Spec.HealthCheckNodePort)
+		svc.ObjectMeta.Annotations = map[string]string{
+			"service.kubernetes.io/ibm-load-balancer-cloud-provider-vpc-health-check-port": healthPortStr,
+		}
+		svc, err = apply.Service(ctx, instance, svc, reqLogger, r.config.ScopedClient, r.config.Scheme)
+	}
+
+	return svc, err
 }
 
-func newLoadBalancerService(instance *v1alpha1.Submariner) *corev1.Service {
+func (r *Reconciler) getOCPPlatformType(ctx context.Context) (string, error) {
+	clusterInfra := &configv1.Infrastructure{}
+	err := r.config.GeneralClient.Get(ctx, types.NamespacedName{Name: "cluster"}, clusterInfra)
+
+	if apierrors.IsNotFound(err) {
+		return "", nil
+	}
+
+	if err != nil {
+		return "", errors.Wrap(err, "error retrieving cluster Infrastructure resource")
+	}
+
+	if clusterInfra.Status.PlatformStatus == nil {
+		return string(clusterInfra.Status.Platform), nil //nolint:staticcheck //Purposely using deprecated field for backwards compatibility
+	}
+
+	return string(clusterInfra.Status.PlatformStatus.Type), nil
+}
+
+func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP string) *corev1.Service {
+	var svcAnnotations map[string]string
+
+	switch platformTypeOCP {
+	case string(configv1.AWSPlatformType):
+		svcAnnotations = map[string]string{
+			"service.beta.kubernetes.io/aws-load-balancer-type": "nlb",
+		}
+	case string(configv1.IBMCloudPlatformType):
+		svcAnnotations = map[string]string{
+			"service.kubernetes.io/ibm-load-balancer-cloud-provider-enable-features":           "nlb",
+			"service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type":                   "public",
+			"service.kubernetes.io/ibm-load-balancer-cloud-provider-vpc-health-check-protocol": "http",
+		}
+	default:
+		svcAnnotations = map[string]string{}
+	}
+
 	return &corev1.Service{
 		ObjectMeta: v1meta.ObjectMeta{
-			Name:      loadBalancerName,
-			Namespace: instance.Spec.Namespace,
-			Annotations: map[string]string{
-				// AWS requires nlb Load Balancer for UDP
-				"service.beta.kubernetes.io/aws-load-balancer-type": "nlb",
-			},
+			Name:        loadBalancerName,
+			Namespace:   instance.Spec.Namespace,
+			Annotations: svcAnnotations,
 		},
 		Spec: corev1.ServiceSpec{
 			ExternalTrafficPolicy: corev1.ServiceExternalTrafficPolicyTypeLocal,
--- a/controllers/submariner/submariner_controller_test.go
+++ b/controllers/submariner/submariner_controller_test.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	v1config "github.com/openshift/api/config/v1"
 	"github.com/submariner-io/admiral/pkg/fake"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
 	"github.com/submariner-io/submariner-operator/controllers/test"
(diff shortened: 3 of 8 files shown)


controllers/submariner/submariner_controller_test.go
```diff
@@ -233,6 +234,7 @@ func testReconciliation() {
 
 	When("load balancer is enabled", func() {
 		BeforeEach(func() {
+			t.InitScopedClientObjs = append(t.InitScopedClientObjs, client.Object(newInfrastructureCluster()))
```
yboaron: @aswinsuryan @tpantelis , any idea how to address the unit testing failure?

I changed reconcileLoadBalancer function  to read Openshift ConfigV1().Infrastructures CR , unit test fails though I added infrastructure CR to t.InitScopedClientObjs

tpantelis: It panics due to

`failed to add object &{{ } {cluster     999 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []} {{ } { <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil>}} { AWS <nil>      }} to fake client: no kind is registered for the type v1.Infrastructure in scheme "pkg/runtime/scheme.go:100"`

You need to register the type with the scheme. Add 

`Expect(v1config.Install(scheme.Scheme)).To(Succeed())`

after [this line](https://github.com/submariner-io/submariner-operator/blob/737cc2ccfd2cef23a9e87aa4bfaaabfc4c635c0c/controllers/submariner/submariner_suite_test.go#L48).

However, it then panics here:

```
runtime error: invalid memory address or nil pointer dereference

  Full Stack Trace
    github.com/openshift/client-go/config/clientset/versioned.NewForConfig(0x1e3c4b8?)
    	/home/tpanteli/go/src/github.com/submariner-io/submariner-operator/vendor/github.com/openshift/client-go/config/clientset/versioned/clientset.go:54 +0x29
    github.com/submariner-io/submariner-operator/controllers/submariner.(*Reconciler).getOCPPlatformType(0x18?, {0x1e39bb8, 0xc000192018})

```

```
func (r *Reconciler) getOCPPlatformType(ctx context.Context) string {
    configClient, err := configv1client.NewForConfig(r.config.RestConfig)
    ...
}

```
`r.config.RestConfig` is not  set by the unit tests since we can't create fake clients from a `rest.Config`. You need to use the `r.config.GeneralClient` to query the OpenShift resource. Also in the test setup you need to add it to `t.InitGeneralClientObjs`.

controllers/submariner/submariner_controller_test.go
```diff
@@ -511,3 +513,16 @@ func testDeletion() {
...
 	})
 }
+
+func newInfrastructureCluster() *v1config.Infrastructure {
+	infraCluster := &v1config.Infrastructure{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+		Status: v1config.InfrastructureStatus{
+			Platform: v1config.AWSPlatformType,
```
tpantelis: This field is deprecated. Need to set `PlatformStatus.Type`.

tpantelis (changes requested):

controllers/submariner/submariner_controller_test.go
yboaron: Well, after changing the code to use r.config.GeneralClient 'getOCPPlatformType' failed in both  'real' cluster env and test env.

In 'real' cluster, I'm getting [1] error although I added it to scheme [2], 

also I noticed that we use ( check [3]) unstructured.Unstructured to retrieve 'Network' CR from **config.openshift.io** (for LB we need the Infrastructure CR from this group) group, should we use the same approach also here? 

[1] 
`2023-07-18T10:44:08.853Z ERR ..alancer_resources.go:86 ..troller_submariner Error r.config.GeneralClient.Get error="no kind is registered for the type v1.Infrastructure in scheme \"pkg/runtime/scheme.go:100\""
`

[2] 
https://github.com/submariner-io/submariner-operator/pull/2727/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R149


[3] 
https://github.com/submariner-io/submariner-operator/blob/devel/pkg/discovery/network/openshift4.go#L36

tpantelis: The unit test panicked b/c `clusterInfra.Status.PlatformStatus` was nil - the test set the deprecated `Platform` instead. I pushed an [update](https://github.com/submariner-io/submariner-operator/compare/97b6d30d4d88293229ebf38709af07eb42a95d33..152a2257e6a56546b7959bd208d44c72c8ca590b) to fix this. In addition, I added unit test cases for _AWS_ and _IBMCloud_ and modified `getOCPPlatformType` to propagate the error.

tpantelis: For the E2E failure, while you registered  the type with the scheme in _main.go_, the general client didn't use that scheme. I modified it to do so in [this update](https://github.com/submariner-io/submariner-operator/compare/ce29883fc683c2dfc1fb301b1e4e36f2f13bc2e0..5d51deafa593547d3c386803da17405548631de6).

tpantelis (approved):

controllers/submariner/submariner_controller_test.go
yboaron: Thanks a lot, @tpantelis .

I checked it on ROKS platform - looks fine!

controllers/submariner/loadbalancer_resources.go
```diff
@@ -43,18 +48,74 @@ const (
...
+		return "", nil
+	}
+
+	if err != nil {
+		return "", errors.Wrap(err, "error retrieving cluster Infrastructure resource")
+	}
+
+	if clusterInfra.Status.PlatformStatus == nil {
+		return "", nil
+	}
```
tpantelis: For backwards compatibility, perhaps we should return the deprecated `Platform` field if the newer field isn't set:

```suggestion
	if clusterInfra.Status.PlatformStatus == nil {
		return string(clusterInfra.Platform), nil
	}
```

yboaron: > For backwards compatibility, perhaps we should return the deprecated `Platform` field if the newer field isn't set:

Could be (it should be `return string(clusterInfra.Status.Platform), nil` ) , but then we'll get below linter error, so not sure about this change.

```
controllers/submariner/loadbalancer_resources.go:90:17: SA1019: clusterInfra.Status.Platform is deprecated: Use platformStatus.type instead. (staticcheck)
                return string(clusterInfra.Status.Platform), nil
```

skitt: Since we’d be using a deprecated field on purpose, it would be find to disable the check (`//nolint:staticcheck`).

controllers/submariner/loadbalancer_resources.go
```diff
@@ -43,18 +48,74 @@ const (
...
+	if apierrors.IsNotFound(err) {
+		return "", nil
+	}
+
+	if err != nil {
+		return "", errors.Wrap(err, "error retrieving cluster Infrastructure resource")
+	}
+
+	if clusterInfra.Status.PlatformStatus == nil {
+		return string(clusterInfra.Status.Platform), nil //nolint:staticcheck //avoid deprecated field error
```
tpantelis: Minor thing but the comment should be the reason for disabling the violation

```suggestion
		return string(clusterInfra.Status.Platform), nil //nolint:staticcheck //Purposely using deprecated field for backwards compatibility
```

config/rbac/submariner-operator/ocp_cluster_role.yaml
```diff
@@ -12,3 +12,9 @@ rules:
       - securitycontextconstraints
     verbs:
       - use
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - infrastructures
```
sridhargaddam, line 18: In the addon code, I see that we are configuring explicit access to "infrastructures/status"
Please check if similar thing needs to be done even here.

https://github.com/stolostron/submariner-addon/blob/61ece25b8f65c6fa522b4f3216871c85f472b27b/deploy/config/rbac/cluster_role.yaml#L50

controllers/submariner/loadbalancer_resources.go
```diff
@@ -43,18 +48,74 @@ const (
...
+		return nil, err
+	}
+
+	// For IBM cloud also needs to annotate the allocated health check node port
+	if platformTypeOCP == string(configv1.IBMCloudPlatformType) {
+		healthPortStr := fmt.Sprintf("%d", svc.Spec.HealthCheckNodePort)
+		svc.ObjectMeta.Annotations = map[string]string{
+			"service.kubernetes.io/ibm-load-balancer-cloud-provider-vpc-health-check-port": healthPortStr,
+		}
+		svc, err = apply.Service(ctx, instance, svc, reqLogger, r.config.ScopedClient, r.config.Scheme)
```
sridhargaddam, line 71: Should we retry here if there is a conflict?

controllers/submariner/loadbalancer_resources.go
```diff
@@ -43,18 +48,74 @@ const (
...
+
+	if apierrors.IsNotFound(err) {
+		return "", nil
+	}
+
+	if err != nil {
+		return "", errors.Wrap(err, "error retrieving cluster Infrastructure resource")
+	}
+
+	if clusterInfra.Status.PlatformStatus == nil {
```
sridhargaddam, line 89: What is the difference between `clusterInfra.Status.PlatformStatus` and `clusterInfra.Status.Platform` ?

tpantelis: The latter is deprecated, replaced by the former.

controllers/submariner/loadbalancer_resources.go
tpantelis: [apply.Service](https://github.com/submariner-io/submariner-operator/blob/edf6e41f3f7ae01fd51ab4ab9ba10cb90d67a5fb/controllers/apply/apply.go#L201) does that.

config/rbac/submariner-operator/ocp_cluster_role.yaml
yboaron: Well, I checked it on OCP (ROKS) cluster and it works without explicit access to "infrastructures/status", 
also I can see that for example SRIOV operator [doesn't specify infrastructure/status](https://github.com/openshift/sriov-network-operator/blob/24856584100497eb402de673ab8fd37f25add796/deploy/clusterrole.yaml#L33) and [use fields from infra/status](https://github.com/openshift/sriov-network-operator/blob/24856584100497eb402de673ab8fd37f25add796/pkg/utils/cluster.go#L112)

skitt: Right, explicit permissions on `/status` sub-structs is only required for code which updates status information.

skitt (approved):

dfarrell07:
@yboaron Do you think this needs to be backported? And reminder about release notes, if they are needed.

yboaron:
> @yboaron Do you think this needs to be backported? And reminder about release notes, if they are needed.

@dfarrell07  , this PR fixed a minor bug in ARO (unneeded annotation relevant for ROSA appears also in ARO) and adds ROKS support.

Since ROKS support should start from 0.16 I don't think backport is needed here.

skitt:
@yboaron do you think this is worth mentioning in the release notes? It seems like an implementation detail.

yboaron:
> @yboaron do you think this is worth mentioning in the release notes? It seems like an implementation detail.

Yep,  also think we shouldn't mention it in release notes.

PR #2770, opened 2023-08-29
Improve podCIDR detection

sridhargaddam:
In K8s, each node is typically assigned a unique PodCIDR range for the pods that run on that node. Each node's PodCIDR is used to allocate IP addresses to the pods scheduled on that node. Only if the cluster is a single node deployment, we should rely on the node.Spec.PodCIDR as podCIDR of the cluster.

Related to: https://github.com/submariner-io/submariner/issues/2128
Related to: https://github.com/submariner-io/submariner/issues/2489


skitt (approved):

sridhargaddam:
This fix is not only for generic clusters but also for calico clusters.

yboaron (approved):

dfarrell07:
@sridhargaddam Do you think this should be backported?

sridhargaddam:
> @sridhargaddam Do you think this should be backported?

Yes @dfarrell07, this will be useful for environments like K3s. I'll handle the backports.

sridhargaddam:
Release note PR: https://github.com/submariner-io/submariner-website/pull/1037

PR #2772, opened 2023-08-31
Mount both Openshift and non-openshift  nbdb path to support upgrades

### LINKED ISSUE #193: Support upgrades to the new OVN IC implementation
Update should be supported 
* when submariner moves from old implementation to new
*  when Openshift moves from an non-ic to ic version

aswinsuryan:
Mount both Openshift and non-openshift  nbdb path to support upgrades

Fixes: https://github.com/submariner-io/enhancements/issues/193

During upgrade to IC route agent will try to connect using one of these socket path. So mounting both.


controllers/submariner/route_agent_resources.go
```diff
@@ -20,9 +20,6 @@ package submariner
 
 import (
 	"context"
```
skitt, line 22:
```suggestion
	"context"
	"strconv"
```

controllers/submariner/route_agent_resources.go
```diff
@@ -35,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
+	"strconv"
```
skitt:
```suggestion
```

controllers/submariner/route_agent_resources.go
```diff
@@ -97,7 +85,11 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, clusterNetwork *network.Clu
 							Path: "/sys",
 						}}},
 						{Name: "host-var-run-openvswitch-nbdb-sock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
-							Path: ovnNBDBPath,
+							Path: "/var/run/openvswitch/ovnnb_db.sock",
+						}}},
+						//Path used by Openshift
```
skitt:
```suggestion
						// Path used by OpenShift
```

Is this really OpenShift-specific, or is it tied to OVN-IC only?

controllers/submariner/route_agent_resources.go
```diff
@@ -118,7 +110,9 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, clusterNetwork *network.Clu
 								{Name: "host-sys", MountPath: "/sys", ReadOnly: true},
 								{Name: "host-run-xtables-lock", MountPath: "/run/xtables.lock"},
 								{Name: "host-run-openvswitch-db-sock", MountPath: "/run/openvswitch/db.sock"},
-								{Name: "host-var-run-openvswitch-nbdb-sock", MountPath: ovnNBDBPath},
+								{Name: "host-var-run-openvswitch-nbdb-sock", MountPath: "/var/run/openvswitch/ovnnb_db.sock"},
+								//Path used by Openshift
```
skitt:
```suggestion
								// Path used by Openshift
```

Same question as above.

skitt (changes requested):

controllers/submariner/route_agent_resources.go
aswinsuryan: it is specific to Openshift IC, since OVN want support certain scenarios when cluster  upgraded, they are using path with ovn-ic instead of OVN , for non-openshift it is just ovn

controllers/submariner/route_agent_resources.go
aswinsuryan: Done

controllers/submariner/route_agent_resources.go
aswinsuryan: As explained above it is specific to Openshift

controllers/submariner/route_agent_resources.go
aswinsuryan: Done.

skitt (approved):

vthapar (approved):

PR #2822, opened 2023-09-20
Create xtables.lock as a file if it doesn't already exist

skitt:
Default host path mounts create a directory if the mounted path doesn't correspond to anything, which ends up breaking iptables. This fixes that by ensuring that xtables.lock is created as a file if it doesn't already exist; subsequent invocations of iptables on the host (if any) will then be able to use the new file for locking.


tpantelis:
This seems to cause a vulnerability error.

skitt:
> This seems to cause a vulnerability error.

The vulnerability is a false positive, see https://github.com/github/advisory-database/pull/2742

tpantelis:
> The vulnerability is a false positive, see [github/advisory-database#2742](https://github.com/github/advisory-database/pull/2742)

OK. I changed the job to required - did you make it non-required due to this?

skitt:
> OK. I changed the job to required - did you make it non-required due to this?

Yes — we can wait for the advisory fix to be merged, or “unrequire” the job until it’s fixed...

dfarrell07 (approved):

tpantelis (approved):

sridhargaddam (approved):

PR #2824, opened 2023-09-21
Wait for node readiness before starting route-agent

### LINKED ISSUE #2802: HostPath mounts in gateway and/or route agent prevent OVN from starting
**What happened**:

In some scenarios, deploying Submariner with OVN causes OVN to fail to start on the gateway node, because by the time `nbdb` starts, `/var/run/ovn/ovnnb_db.sock` has been created as a directory as a result of a host-path mount. This happens because our mounts are unqualified.

If ` /var/run/ovn/ovnnb_db.sock` (or any other file we want to mount) doesn’t exist, it shouldn’t be mounted as a directory.

Deploy using OVN, and examine the contents of `/var/run/ovn` in the `ovnkube-node` container on one of the worker nodes.

skitt:
Route-agent startup is subject to races with OVN components: it mounts the sockets from the host, and since the default host path mount behaviour is to create a directory if the path is missing, if route-agent pod initialisation happens before OVN has opened the sockets, they get created as directories. This blocks socket creation and OVN fails to start.

This wouldn't be a problem for most pods because they have tolerations ensuring the node is ready (including CNI readiness) before they start. The route-agent however has tolerates all taints, to ensure it runs everywhere; this means it starts as soon as a node is ready. There is no way to set up tolerations "except" a specific taint, so the route-agent can't be specified in such a way that it will start with any taint except node readiness or network availability.

It also isn't possible to handle this by specifying the socket host path type; this causes the scheduler to wait for the socket to be available before starting the pod. The route-agent needs to be able to mount a number of different socket paths, to handle different setups, and there is never a configuration where all socket paths are available; so enforcing a socket type prevents the route-agent from starting at all.

To handle this, an init container is set up for the route-agent, and waits until the node is ready before allowing the route-agent setup to continue. This init container does not specify the host path volumes used by the main container, so the corresponding paths aren't touched on the host. As a result, the route-agent is only set up once the node is fully ready, including OVN sockets, so the appropriate sockets are mounted correctly. (Directories are still created for missing socket mounts, but that doesn't matter, because once this stage is reached the missing socket mounts correspond to paths which aren't used by OVS or OVN.)

Fixes: #2802
Depends on https://github.com/submariner-io/submariner/pull/2719

### DIFF
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -92,6 +92,24 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 							Path: "/var/run/ovn-ic/ovnnb_db.sock",
 						}}},
 					},
+					// The route agent needs to wait for the node to be ready before starting,
+					// to avoid racing with the CNI for socket setup; this init container takes care of that
+					InitContainers: []corev1.Container{
+						{
+							Name:            name + "-init",
+							Image:           getImagePath(cr, opnames.RouteAgentImage, names.RouteAgentComponent),
+							ImagePullPolicy: images.GetPullPolicy(cr.Spec.Version, cr.Spec.ImageOverrides[names.RouteAgentComponent]),
+							Command:         []string{"submariner-route-agent.sh"},
+							Env: []corev1.EnvVar{
+								{Name: "SUBMARINER_WAITFORNODE", Value: "true"},
+								{Name: "NODE_NAME", ValueFrom: &corev1.EnvVarSource{
+									FieldRef: &corev1.ObjectFieldSelector{
+										FieldPath: "spec.nodeName",
+									},
+								}},
+							},
+						},
+					},
 					Containers: []corev1.Container{
 						{
 							Name:            name,


controllers/submariner/route_agent_resources.go
```diff
@@ -92,6 +92,21 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 							Path: "/var/run/ovn-ic/ovnnb_db.sock",
 						}}},
 					},
+					InitContainers: []corev1.Container{
+						{
+							Name:            name + "-init",
+							Image:           getImagePath(cr, opnames.RouteAgentImage, names.RouteAgentComponent),
+							ImagePullPolicy: images.GetPullPolicy(cr.Spec.Version, cr.Spec.ImageOverrides[names.RouteAgentComponent]),
+							Command:         []string{"kubectl", "wait", "node/$(NODE_NAME)", "--for", "condition=ready"},
```
sridhargaddam: Instead of using a new kubectl binary, can we not use the same route-agent pod as an initContainer where we set some env' variable (similar to how we are [doing in Uninstall](https://github.com/submariner-io/submariner/blob/b9f63cda37c4810dae9d54fd76cb809727119dcb/pkg/routeagent_driver/main.go#L136)) and when the variable is set, it can simply read the [node status](https://github.com/submariner-io/submariner/blob/devel/pkg/node/node.go) and check if its Ready?

skitt: I thought of doing that in a second phase but it does make the fix simpler to ship, so I’ll do that now.

sridhargaddam:
@skitt In the first sentence, the following is mentioned.

> if route-agent pod initialisation happens before OVN has opened the sockets, they get created as directories. This blocks socket creation and OVN fails to start.

Whereas the last sentence says...
>(Directories are still created for missing socket mounts, **but that doesn't matter**.)

It appears like contradicting the first sentence. Please clarify.

skitt:
> It appears like contradicting the first sentence. Please clarify.

I’ve updated the commit message and PR description.

sridhargaddam (approved):

aswinsuryan (approved):

PR #2841, opened 2023-10-11
Ensure that ovn-ic/ovnnb_db.sock is created as a file

yboaron:
Apply fix similar to [1] for ovn-ic/ovnnb_db.sock

[1]
https://github.com/submariner-io/submariner-operator/commit/93f8d3aa66f1c2f898ef4c6a7a6644fa7a6bf956

### DIFF
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -89,7 +89,7 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 						}}},
 						// Path used by Openshift
 						{Name: "host-var-run-ovn-ic-nbdb-sock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
-							Path: "/var/run/ovn-ic/ovnnb_db.sock",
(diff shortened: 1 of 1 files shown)


controllers/submariner/route_agent_resources.go
```diff
@@ -89,7 +89,7 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 						}}},
 						// Path used by Openshift
 						{Name: "host-var-run-ovn-ic-nbdb-sock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
-							Path: "/var/run/ovn-ic/ovnnb_db.sock",
+							Path: "/var/run/ovn-ic/ovnnb_db.sock", Type: ptr.To(corev1.HostPathFileOrCreate),
```
sridhargaddam, line 92: I think we can apply this change to the other sockets too? WDYS @skitt @aswinsuryan

aswinsuryan: yes I think , we would need that for removing existing bridges in the case of an upgrade from old implementation to new implementation.

skitt (changes requested):
This can’t work :frowning_face:. See #2824 for details. (Note too that a socket isn’t a file.)

PR #2846, opened 2023-10-13
Mount host path directories instead of sockets

skitt:
So far we’ve tried to mount individual host files which are needed in the containers, to limit the exposure. However this doesn’t work well for sockets: because the consuming end can’t create them, they need to be mounted as files, and because Kubernetes (or CRI) doesn’t support “opportunistic” mounts, their type can’t be specified; as a result, if the container starts before the process which opens the socket, the mount ends up being created as a directory, and the process trying to open the socket then fails.

There are two potential fixes for this:

* if the operator could determine conclusively which socket mounts are required for a container, it could define only the necessary mounts, and have the container block waiting for them to be available;
* as done in ovnkubernetes, the containers can mount the containing directories.

The first approach seems preferable but creates a strong dependency on specific behaviours from the cluster environment (specific paths used by specific versions of OpenShift components in specific scenarios). The second isn’t great from a security perspective but seems like the best compromise.

/var/run is a symlink to /run; this also changes the various paths to use canonical paths and merges the duplicates.

### DIFF
--- a/controllers/submariner/route_agent_resources.go
+++ b/controllers/submariner/route_agent_resources.go
@@ -73,23 +73,21 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 				Spec: corev1.PodSpec{
 					TerminationGracePeriodSeconds: ptr.To(int64(1)),
 					Volumes: []corev1.Volume{
-						// We need to share /run/xtables.lock with the host for iptables
+						// Share /run/xtables.lock with the host for iptables
 						{Name: "host-run-xtables-lock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
 							Path: "/run/xtables.lock", Type: ptr.To(corev1.HostPathFileOrCreate),
 						}}},
-						// We need to share /run/openvswitch/db.sock with the host for OVS
-						{Name: "host-run-openvswitch-db-sock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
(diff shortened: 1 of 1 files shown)


controllers/submariner/route_agent_resources.go
```diff
@@ -127,10 +125,8 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 							VolumeMounts: []corev1.VolumeMount{
 								{Name: "host-sys", MountPath: "/sys", ReadOnly: true},
 								{Name: "host-run-xtables-lock", MountPath: "/run/xtables.lock"},
-								{Name: "host-run-openvswitch-db-sock", MountPath: "/run/openvswitch/db.sock"},
-								{Name: "host-var-run-openvswitch-nbdb-sock", MountPath: "/var/run/openvswitch/ovnnb_db.sock"},
```
sridhargaddam, line 131: Why are we removing "/var/run/openvswitch/ovnnb_db.sock" and mounting only "/run/openvswitch"?

sridhargaddam (changes requested):

controllers/submariner/route_agent_resources.go
skitt: `/var/run` is a symlink to `/run`, so `/run/openvswitch` and `/var/run/openvswitch` are the same; this only mounts one of them, the other appears through the symlink. I’ve added this to the commit message.

sridhargaddam (approved):

vthapar (approved):

aswinsuryan (approved):

PR #2854, opened 2023-10-24
Allow halting (and restarting) pods on certificate errors

skitt:
General practice in Kubernetes is to restart pods relying on certificates when those certificates are changed. Because Kubernetes isn't aware of cross-cluster certificate changes, this doesn't happen automatically when the broker certificate (or trust chain) changes; this produces certificate errors and ultimately results in a broken setup.

To avoid this, provide a setting to set up gateway and Lighthouse agent pods to halt on certificate errors.

See https://github.com/submariner-io/submariner/pull/2761 and https://github.com/submariner-io/lighthouse/pull/1408.

### DIFF
--- a/controllers/servicediscovery/servicediscovery_controller.go
+++ b/controllers/servicediscovery/servicediscovery_controller.go
@@ -246,6 +246,7 @@ func newLighthouseAgent(cr *submarinerv1alpha1.ServiceDiscovery, name string) *a
 								{Name: "SUBMARINER_CLUSTERID", Value: cr.Spec.ClusterID},
 								{Name: "SUBMARINER_DEBUG", Value: strconv.FormatBool(cr.Spec.Debug)},
 								{Name: "SUBMARINER_GLOBALNET_ENABLED", Value: strconv.FormatBool(cr.Spec.GlobalnetEnabled)},
+								{Name: "SUBMARINER_HALT_ON_CERT_ERROR", Value: strconv.FormatBool(cr.Spec.HaltOnCertificateError)},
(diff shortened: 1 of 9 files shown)


tpantelis (approved):

controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -246,6 +246,7 @@ func newLighthouseAgent(cr *submarinerv1alpha1.ServiceDiscovery, name string) *a
 								{Name: "SUBMARINER_CLUSTERID", Value: cr.Spec.ClusterID},
 								{Name: "SUBMARINER_DEBUG", Value: strconv.FormatBool(cr.Spec.Debug)},
 								{Name: "SUBMARINER_GLOBALNET_ENABLED", Value: strconv.FormatBool(cr.Spec.GlobalnetEnabled)},
+								{Name: "SUBMARINER_HALTONCERTIFICATEERROR", Value: strconv.FormatBool(cr.Spec.HaltOnCertificateError)},
```
sridhargaddam: This looks mouthful, what about the following?

```suggestion
								{Name: "SUBMARINER_HALT_ON_CERT_ERROR", Value: strconv.FormatBool(cr.Spec.HaltOnCertificateError)},
```

skitt: Fixed in all three PRs, thanks for the suggestion!

sridhargaddam (approved):

dfarrell07 (approved):

PR #2883, opened 2023-11-08
Bump opentelemetry dependencies

skitt:
This is required to avoid dependency issues on internal opentelemetry modules when upgrading other dependencies.


tpantelis (approved):

sridhargaddam (approved):

dfarrell07:
We need a grpc bump, I can send on top of this PR.

PR #2888, opened 2023-11-13
Revert to default mount behaviour for xtables.lock

skitt:
The xtables.lock mount was fixed to specify its type: it must exist as a file, or be created as a file.

xtables.lock is only used with legacy iptables. On platforms using iptables-nft, the file isn't used and doesn't exist. As a result, previous versions of Submariner created it as a directory (this is the default behaviour for volume mounts in Kubernetes: if the mount doesn't exist, it is created as a directory). When the volume mount type is specified as a file, the existence of a directory causes the mount to fail and the corresponding pod is never scheduled.

To avoid this, revert to the default behaviour. On systems where the lock is important, it already exists so the directory isn't created and the correct behaviour is guaranteed. On systems where the lock isn't needed, it is created as a directory but that doesn't matter.

Future releases of Submariner will have to deal with this correctly, and handle upgrades, ideally without mounting all of /run permanently.

### DIFF
--- a/controllers/submariner/globalnet_resources.go
+++ b/controllers/submariner/globalnet_resources.go
@@ -72,7 +72,7 @@ func newGlobalnetDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.DaemonS
 				Spec: corev1.PodSpec{
 					Volumes: []corev1.Volume{
 						{Name: "host-run-xtables-lock", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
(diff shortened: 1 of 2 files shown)


sridhargaddam (approved):

sridhargaddam:
I'm fine with merging this, but maybe we should have reverted the following PR adding additional details - https://github.com/submariner-io/submariner-operator/pull/2822

skitt:
> I'm fine with merging this, but maybe we should have reverted the following PR adding additional details - #2822

You’re right, I’ll revert that in the morning.

PR #2971, opened 2024-02-16
Bump helm from 3.11.1 to 3.14.1 for CVE-2024-25620

dfarrell07:
Update generated by:

```
go get -u helm.sh/helm/v3@v3.14.1
go mod tidy
<manually undo go min-version/toolchain updates>
```

Issue identified by a Dependabot alert (#2968) on devel.


skitt (changes requested):
Please redo this without `-u` to reduce the amount of churn. See https://github.com/submariner-io/shipyard/pull/1555#issuecomment-1948029416 for details. Note too that you can’t undo _all_ the Go language level bumps: Helm 3.14.1 needs (or says it needs) Go 1.21.

dfarrell07:
Redid without `-u`, but I don't think that changed anything. Also didn't remove the go update this time, as discussed it seems that's caused by the Helm update.

I think this update is going to be a problem on this branch, as we don't have a `tools` directory here to keep the "required"-by-Helm K8s API update separate. We may not be able to do this update on this branch in this repo.

tpantelis:
We decided to ignore the CVE: https://github.com/submariner-io/submariner-operator/pull/2980.

PR #2979, opened 2024-02-27
Ignore helm CVE-2024-25620

tpantelis:
Fixing it requires updating K8s et al to incompatible versions.

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -10,3 +10,10 @@ ignore:
   - vulnerability: CVE-2021-22570
     package:
       name: google.golang.org/protobuf
+  # Fixing this CVE requires updating K8s et al to incompatible versions.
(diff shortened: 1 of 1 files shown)


.grype.yaml
```diff
@@ -10,3 +10,7 @@ ignore:
   - vulnerability: CVE-2021-22570
     package:
       name: google.golang.org/protobuf
+  # Fixing this CVE requires updating K8s et al to incompatible versions.
+  - vulnerability: GHSA-r53h-jv2g-vpx6
```
dfarrell07, line 14: I guess the ID Grype is using is `GHSA-v53g-5gjp-272r`, that's what I see in the SARIF output.

vthapar (approved):

dfarrell07 (approved):

PR #2989, opened 2024-03-05
Fix K8s versions tested in all-K8s GHA

dfarrell07:
Now that we test the top and bottom of the non-EOL K8s version range in the E2E-full job, there's no need to test them in the all-K8s job. The test matrix coverage for other features is also the same for both GHAs.


.github/workflows/e2e-all-k8s.yml
```diff
@@ -19,9 +19,9 @@ jobs:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
-        # Top/bottom of supported version range. For details:
+        # All non-EOL K8s versions not tested in E2E-full
         # https://submariner.io/development/building-testing/ci-maintenance/
-        k8s_version: ['1.26', '1.27', '1.28', '1.29']
+        k8s_version: ['1.27', '1.28']
```
tpantelis, line 24: Isn't the purpose of this job to test all K8s versions, specifically the intermediate ones? Otherwise it looks like this job is essentially the same as E2E full.

dfarrell07: The top/bottom of the range are already tested with this same test matrix in E2E-full. There's no point in testing them twice. This job fills the gaps not covered by E2E-full to provide full K8s version coverage. See the docs update: https://github.com/submariner-io/submariner-website/pull/1114

tpantelis: ahh ok  - I missed that.

tpantelis (approved):

yboaron (approved):

PR #3013, opened 2024-04-02
Populate proxy settings in operator-managed pods

skitt:
The operator can be given proxy settings e.g. by OLM, but it's up to the operator to populate those settings in any pods it manages.


adibrastegarnia (approved):

tpantelis (approved):

vthapar (approved):

adibrastegarnia:
Is the devel image gets published for this PR after merge? I would like to give it a try.

skitt:
@adibrastegarnia yes, the `devel` images have been updated with this change.

adibrastegarnia:
I redeployed broker and join the cluster using --version=devel but I still don't see the enviroment variables in each pod.

tpantelis:
> I redeployed broker and join the cluster using --version=devel but I still don't see the enviroment variables in each pod.

This PR  changed the operator to extract the HTTP proxy settings from its container env and pass to the managed gateway et al `DaemonSets`.  However the env settings still need to be populated in the operator pod. Try modifying the operator `Deployment` and set the HTTP proxy variables in the container env.

@skitt We could also modify `subctl join` to pass the HTTP proxy env settings to the operator `Deployment`. Of course, this assumes the user has the settings in their env when running `subctl join`.

adibrastegarnia:
> > I redeployed broker and join the cluster using --version=devel but I still don't see the enviroment variables in each pod.
> 
> This PR changed the operator to extract the HTTP proxy settings from its container env and pass to the managed gateway et al `DaemonSets`. However the env settings still need to be populated in the operator pod. Try modifying the operator `Deployment` and set the HTTP proxy variables in the container env.

Not sure this is gonna help because I assume the submariner components will be deployed after join so even If I change the enviroment variable then the operator gets installed again with no enviroment variable in it. Did I miss anything @tpantelis ?

tpantelis:
> > > I redeployed broker and join the cluster using --version=devel but I still don't see the enviroment variables in each pod.
> > 
> > 
> > This PR changed the operator to extract the HTTP proxy settings from its container env and pass to the managed gateway et al `DaemonSets`. However the env settings still need to be populated in the operator pod. Try modifying the operator `Deployment` and set the HTTP proxy variables in the container env.
> 
> Not sure this is gonna help because I assume the submariner components will be deployed after join so even If I change the enviroment variable then the operator gets installed again with no enviroment variable in it. Did I miss anything @tpantelis ?

Yes that would occur if you run `subclt join` again.  But it's a workaround for now.

skitt:
> @skitt We could also modify `subctl join` to pass the HTTP proxy env settings to the operator `Deployment`. Of course, this assumes the user has the settings in their env when running `subctl join`.

I thought about that but I think that may be more of a trap than anything — in enterprise environments it’s common for users to have proxy settings on their desktops that don’t work for servers. I agree a complete solution would involve adding arguments to `subctl join` (and perhaps in the `Submariner` CR). A comprehensive solution would also support certificates for proxy authentication...

skitt:
> Not sure this is gonna help because I assume the submariner components will be deployed after join so even If I change the enviroment variable then the operator gets installed again with no enviroment variable in it. Did I miss anything @tpantelis ?

You need to run `join` first, then modify the operator `Deployment` object to include the proxy environment variables. That will cause any existing deployed components to be re-deployed with the environment variables set. Any subsequent `join` will remove the variables but that shouldn’t be necessary after the initial installation.

tpantelis:
> > @skitt We could also modify `subctl join` to pass the HTTP proxy env settings to the operator `Deployment`. Of course, this assumes the user has the settings in their env when running `subctl join`.
> 
> I thought about that but I think that may be more of a trap than anything — in enterprise environments it’s common for users to have proxy settings on their desktops that don’t work for servers. I agree a complete solution would involve adding arguments to `subctl join` (and perhaps in the `Submariner` CR). A comprehensive solution would also support certificates for proxy authentication...

OK. I pushed https://github.com/submariner-io/submariner-operator/pull/3016 but we won't need it then.

adibrastegarnia:
> > Not sure this is gonna help because I assume the submariner components will be deployed after join so even If I change the enviroment variable then the operator gets installed again with no enviroment variable in it. Did I miss anything @tpantelis ?
> 
> You need to run `join` first, then modify the operator `Deployment` object to include the proxy environment variables. That will cause any existing deployed components to be re-deployed with the environment variables set. Any subsequent `join` will remove the variables but that shouldn’t be necessary after the initial installation.

What I did for now is to build a docker image on top of submariner-operator:devel to set those env variables in container level and deployed it and it seems it is setting proxy env variables in gateway, route-agent, etc

skitt:
> > You need to run `join` first, then modify the operator `Deployment` object to include the proxy environment variables. That will cause any existing deployed components to be re-deployed with the environment variables set. Any subsequent `join` will remove the variables but that shouldn’t be necessary after the initial installation.
> 
> What I did for now is to build a docker image on top of submariner-operator:devel to set those env variables in container level and deployed it and it seems it is setting proxy env variables in gateway, route-agent, etc

Apologies for not being clear enough. If you run `subctl join` with the current images, you can then edit the operator’s deployment object:

```
kubectl edit deployment -n submariner-operator submariner-operator
```

Look for an `env:` section with `WATCH_NAMESPACE` etc., and add your proxy settings there:

```
        env:
        - name: WATCH_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: OPERATOR_NAME
          value: submariner-operator
        - name: HTTPS_PROXY
          value: https://proxy.example.com:3128/
        - name: HTTP_PROXY
          value: https://proxy.example.com:3128/
```

This will update the operator to use the proxy itself, then the operator will update the other pods. You don’t need to create a separate container image.

vthapar:
> > > @skitt We could also modify `subctl join` to pass the HTTP proxy env settings to the operator `Deployment`. Of course, this assumes the user has the settings in their env when running `subctl join`.
> > 
> > 
> > I thought about that but I think that may be more of a trap than anything — in enterprise environments it’s common for users to have proxy settings on their desktops that don’t work for servers. I agree a complete solution would involve adding arguments to `subctl join` (and perhaps in the `Submariner` CR). A comprehensive solution would also support certificates for proxy authentication...
> 
> OK. I pushed #3016 but we won't need it then.

I second the idea to have this in subctl as user can specify proxy during join and don't have to manually edit the deployment. Manually editing deployment should alwyas be a workaround and not standard procedure. To clarify, I mean allowing users to set any proxy, not picking proxy settings from user's env where subctl is run from.

PR #3016, opened 2024-04-03
Move HTTP proxy code to a public package

tpantelis:
...so it can be used by **subctl**.


vthapar (approved):

vthapar:
@tpantelis Do we still need this?

tpantelis:
> @tpantelis Do we still need this?

I don't think so based on @skitt [comment here](https://github.com/submariner-io/submariner-operator/pull/3013#issuecomment-2035366304). The settings should be passed on the command line in which case we wouldn't use this code.

PR #3042, opened 2024-05-03
Restrict route agent secret access

skitt:
The route agent needs to access OVN secrets; currently that's done by granting it access to all secrets. Access to secrets can be used for privilege escalation, so it's best to restrict this to only required secrets; for OVN this is the openshift-ovn-kubernetes namespace.

### DIFF
--- a/config/rbac/submariner-route-agent/ovn_cluster_role.yaml
+++ b/config/rbac/submariner-route-agent/ovn_cluster_role.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: submariner-routeagent-ovn
+rules:
+  - apiGroups:  # the route agent needs access to ovn secrets
+      - ""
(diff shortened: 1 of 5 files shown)


aswinsuryan (approved):

dfarrell07 (approved):

tpantelis (approved):

config/rbac/submariner-route-agent/ovn_cluster_role.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: submariner-routeagent-ovn
+rules:
+  - apiGroups:  # the route agent needs access to ovn secrets
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
```
tpantelis: We don't need `watch` access. I removed that in my PR.

tpantelis (changes requested):

tpantelis (approved):

PR #3051, opened 2024-05-07
Remove node update permission for the route-agent

tpantelis:
It only needs read/watch access.

### DIFF
--- a/config/rbac/submariner-route-agent/cluster_role.yaml
+++ b/config/rbac/submariner-route-agent/cluster_role.yaml
@@ -30,7 +30,6 @@ rules:
       - get
       - list
       - watch
-      - update
   - apiGroups:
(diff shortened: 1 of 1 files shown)


skitt (approved):

tpantelis:
It turns out RA does update nodes - adds an annotation for GN:

`2024-05-07T03:19:56.903Z ERR ..gent_driver/main.go:155 main                 Error while annotating the node error=“error annotating node with CNI interface IP: error updatating node \“cluster1-control-plane\“: nodes \“cluster1-control-plane\” is forbidden: User \“system:serviceaccount:submariner-operator:submariner-routeagent\” cannot update resource \“nodes\” in API group \“\” at the cluster scope”`

Perhaps we should revisit this later.

PR #3057, opened 2024-05-07
Remove packagemanifests

tpantelis:
These are for old releases and we haven’t had a working bundle upstream in quite a while.

The K8s yaml files cause a lot of Sonar violations.


skitt (approved):

dfarrell07:
There is `using=bundle` deployment automation in Shipyard to test this, but I think we're not actually running it. I can remove it if you want?

dfarrell07 (approved):

PR #3059, opened 2024-05-08
Expand wildcards in RBAC yaml files

tpantelis:
It's recommended to not use wildcards for resources and verbs as it can lead to security issues by granting overly broad permissions.

Even though the wildcard uses pertained to `multicluster.x-k8s.io `and `submariner.io` resources which may not carry security risk, we should be consistent and follow the recommendation. Also doing so avoids Sonar violations.

### DIFF
--- a/config/rbac/submariner-diagnose/role.yaml
+++ b/config/rbac/submariner-diagnose/role.yaml
@@ -33,14 +33,19 @@ rules:
   - apiGroups:
       - submariner.io
     resources:
-      - '*'
+      - clusters
+      - endpoints
+      - gateways
+      - gatewayroutes
+      - nongatewayroutes
     verbs:
       - get
       - list
   - apiGroups:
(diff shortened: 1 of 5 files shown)


dfarrell07 (approved):

config/rbac/submariner-diagnose/role.yaml
```diff
@@ -33,14 +33,19 @@ rules:
   - apiGroups:
       - submariner.io
     resources:
-      - '*'
+      - clusters
```
aswinsuryan, line 36: Does it need globalnet CR( like ClusterGlobalEgressIP) permission too ?

tpantelis: Those are cluster-scoped so specified in the ClusterRole.

aswinsuryan (approved):

PR #3064, opened 2024-05-14
Configure SAs to enforce mountable secrets

skitt:
This prevents accessing arbitrary secrets in pods running with these SAs. See
https://kubernetes.io/docs/reference/labels-annotations-taints/#enforce-mountable-secrets for details.

### DIFF
--- a/config/rbac/submariner-gateway/service_account.yaml
+++ b/config/rbac/submariner-gateway/service_account.yaml
@@ -3,3 +3,8 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: submariner-gateway
+  annotations:
+    kubernetes.io/enforce-mountable-secrets: "true"
+secrets:
+  - name: submariner-broker-secret
+  - name: submariner-ipsec-psk
--- a/config/rbac/lighthouse-agent/service_account.yaml
+++ b/config/rbac/lighthouse-agent/service_account.yaml
@@ -3,3 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: submariner-lighthouse-agent
+  annotations:
+    kubernetes.io/enforce-mountable-secrets: "true"
+secrets:
+  - name: submariner-broker-secret
--- a/config/rbac/lighthouse-coredns/service_account.yaml
+++ b/config/rbac/lighthouse-coredns/service_account.yaml
@@ -3,3 +3,5 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: submariner-lighthouse-coredns
+  annotations:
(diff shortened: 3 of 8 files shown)


skitt:
Verified by https://github.com/submariner-io/subctl/pull/1128

config/rbac/submariner-gateway/service_account.yaml
```diff
@@ -3,3 +3,8 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: submariner-gateway
+  annotations:
+    kubernetes.io/enforce-mountable-secrets: "true"
+secrets:
+  - name: submariner-broker-secret
+  - name: submariner-ipsec-psk
```
tpantelis, line 10: I see you're listing mounted secrets here but not for the others. How come?

skitt: I think this is the only SA that mounts secrets other than the Kubernetes-provisioned secret (which is allowed by default).

skitt: Ah, no, the Lighthouse agent does this too...

dfarrell07 (approved):

config/rbac/submariner-gateway/service_account.yaml
tpantelis: So if specific `secrets` are not specified, then this will only allow the default K8s-provisioned secret to be mounted?

I wonder how it works with `automountServiceAccountToken: false` ...

skitt: I think that’s orthogonal: this setting determines which secrets are allowed to be mounted, the automount setting determines whether the default provisioned secret is mounted automatically.

tpantelis (approved):

dfarrell07:
Should this be backported?

skitt:
> Should this be backported?

I think it should, but only after thorough testing — in particular, it reverts https://github.com/submariner-io/subctl/pull/884 so I want to make sure that’s not a problem.

dfarrell07:
> > Should this be backported?
> 
> I think it should, but only after thorough testing — in particular, it reverts [submariner-io/subctl#884](https://github.com/submariner-io/subctl/pull/884) so I want to make sure that’s not a problem.

What's the plan for getting this testing for devel? Do we want to bump subctl and other repos to this operator version to consume the change?

skitt:
> What's the plan for getting this testing for devel? Do we want to bump subctl and other repos to this operator version to consume the change?

That’s already planned, https://github.com/submariner-io/subctl/pull/1128

dfarrell07:
We were hoping to see this work in 0.18, including d/s, before backporting it.

PR #3067, opened 2024-05-15
Add more missing resources for diagnose and broker-client

skitt:
The broker-client role needs to be able to update
serviceimports/status. The broker-admin role needs to have at least the same privileges as broker-client to be able to grant those permissions.


tpantelis (approved):

skitt:
Check https://github.com/submariner-io/subctl/pull/1130 for verification before merging this.

dfarrell07 (approved):

PR #3088, opened 2024-05-28
Use the GeneralClient when accessing operatorv1.DNS resources

tpantelis:
...instead of the `ScopedClient`. The `ScopedClient` should only be used for accessing resources in the operator namespace. Also, since it caches internally, it requires list and watch permissions.

### DIFF
--- a/controllers/servicediscovery/servicediscovery_controller.go
+++ b/controllers/servicediscovery/servicediscovery_controller.go
@@ -561,10 +561,9 @@ func (r *Reconciler) configureOpenshiftClusterDNSOperator(ctx context.Context, i
 func (r *Reconciler) updateLighthouseConfigInOpenshiftDNSOperator(ctx context.Context, instance *submarinerv1alpha1.ServiceDiscovery,
 	clusterIP string,
 ) error {
-	//nolint:wrapcheck // No need to wrap errors here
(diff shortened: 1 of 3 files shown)


skitt (approved):

controllers/servicediscovery/servicediscovery_controller.go
```diff
@@ -582,21 +582,17 @@ func (r *Reconciler) updateLighthouseConfigInOpenshiftDNSOperator(ctx context.Co
...
-			Name:   dnsOperator.Name,
-			Labels: dnsOperator.Labels,
-		}}
-
-		result, err := controllerutil.CreateOrUpdate(ctx, r.ScopedClient, toUpdate, func() error {
-			toUpdate.Spec = dnsOperator.Spec
-			for k, v := range dnsOperator.Labels {
-				toUpdate.Labels[k] = v
-			}
+		err := util.MustUpdate[*operatorv1.DNS](ctx, resource.ForControllerClient(r.GeneralClient, "", dnsOperator), dnsOperator,
```
vthapar, line 584: Tested and verified that this fixes the issue. Change from CreateOrUpdate to MustUpdate is fine. It can fail if object is deleted after Get but before MustUpdate finishes, though in that case  we shouldn't be trying to Create coz we don't own the object.

vthapar (approved):
Tested

PR #3091, opened 2024-05-30
Adjust globalnet RBAC permissions

tpantelis:
Globalnet now annotates `Gateways` instead of nodes.

Related to https://github.com/submariner-io/submariner/pull/3019

### DIFF
--- a/config/rbac/submariner-globalnet/role.yaml
+++ b/config/rbac/submariner-globalnet/role.yaml
@@ -6,6 +6,15 @@ metadata:
   name: submariner-globalnet
 rules:
   - apiGroups:
+      - submariner.io
+    resources:
+      - gateways
+    verbs:
+      - get
+      - list
+      - watch
+      - update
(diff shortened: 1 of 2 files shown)


vthapar (approved):

config/rbac/submariner-globalnet/cluster_role.yaml
```diff
@@ -5,15 +5,6 @@ metadata:
   creationTimestamp: null
   name: submariner-globalnet
 rules:
-  - apiGroups:
```
vthapar: To resovle the chicken and egg scenario, how about we don't delete this RBAC yet and just add it for gateways. This will allow submariner PR to pass as well as find other issues, if any. Then we can create another PR to delete nodes RABC.

vthapar (changes requested):

vthapar (approved):

PR #3162, opened 2024-07-24
Address TODOs regarding cluster network discovery

tpantelis:
See individual commits.

### DIFF
--- a/pkg/discovery/network/network.go
+++ b/pkg/discovery/network/network.go
@@ -65,27 +65,13 @@ func (cn *ClusterNetwork) IsComplete() bool {
 
 func Discover(ctx context.Context, client controllerClient.Client, operatorNamespace string) (*ClusterNetwork, error) {
 	discovery, err := networkPluginsDiscovery(ctx, client)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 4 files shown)


pkg/discovery/network/network.go
```diff
@@ -65,27 +65,13 @@ func (cn *ClusterNetwork) IsComplete() bool {
...
-		globalCIDR, _ := getGlobalCIDRs(ctx, client, operatorNamespace)
-		discovery.GlobalCIDR = globalCIDR
-
-		if discovery.IsComplete() {
-			return discovery, nil
-		}
-
 		// If the info we got from the non-generic plugins is incomplete
 		// try to complete with the generic discovery mechanisms
-		if len(discovery.ServiceCIDRs) == 0 || len(discovery.PodCIDRs) == 0 {
```
aswinsuryan, line 83: we don't need these checks? just checking for 'discovery.IsComplete()' enough?

tpantelis: These are redundant with `discovery.IsComplete()` AFAICT

aswinsuryan (approved):

skitt (approved):

PR #3201, opened 2024-08-21
Use NewClientset instead of NewSimpleClientset

skitt:
The latter is deprecated in Kubernetes 1.31.

### DIFF
--- a/pkg/crd/updater_test.go
+++ b/pkg/crd/updater_test.go
@@ -49,7 +49,7 @@ var _ = Describe("Updater", func() {
 	)
 
 	BeforeEach(func() {
-		client = extendedfakeclientset.NewSimpleClientset()
+		client = extendedfakeclientset.NewClientset()
 		updater = crd.UpdaterFromClientSet(client)
 	})
 


tpantelis:
Unit tests are failing due to:

`failed to convert new object (/submariners.submariner.io; apiextensions.k8s.io/v1, Kind=CustomResourceDefinition) to smd typed: schema error: no type found matching: io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinition`

Long story short, `NewClientset` handles field management via schema. It ends up trying to validate a resource type using the global `schemaYAML` var defined in the generated file _pkg/client/applyconfiguration/internal/internal.go_. For core K8s types, `schemaYAML` is populated appropriately (https://github.com/kubernetes/client-go/blob/master/applyconfigurations/internal/internal.go). But for `apiextensions`, it is not: https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/client/applyconfiguration/internal/internal.go. Thus the error. 

`NewClientset` is new in K8s 0.31. It seems support for it isn't complete - `apiextensions` client code needs to be updated. So we can't use `NewClientset` except for K8s core types. Perhaps that's why DEPRECATED is capitalized in `NewSimpleClientset` so **golangci-lint** doesn't flag it 😄

PR #3212, opened 2024-09-06
Refactor CIDR manipulation code from globalnet to cidr module

tpantelis:
...for reuse with the clusterset IP functionality.


vthapar (approved):

pkg/discovery/globalnet/globalnet.go
```diff
@@ -65,152 +44,8 @@ type Config struct {
...
-
-func uintToIP(ip uint32) net.IP {
-	netIP := make(net.IP, 4)
-	binary.BigEndian.PutUint32(netIP, ip)
-
-	return netIP
+	return cidr.Allocate(&globalnetInfo.Info) //nolint:wrapcheck // No need to wrap
 }
 
 func GetValidClusterSize(cidrRange string, clusterSize uint) (uint, error) {
```
vthapar, line 51: This can also move to cidr/cidr.go as this will also get reused in clusersetip. But I can do that as part of my changes to add clustersetip code. I'd rather have this merged first.

tpantelis: I didn't move this b/c I didn't think we were going to make cluster/allocation size configurable for clustersetip.

aswinsuryan (approved):

PR #3273, opened 2024-11-20
Add init container to the gateway to wait for node readiness

tpantelis:
See https://github.com/submariner-io/submariner/pull/3222

### DIFF
--- a/controllers/submariner/gateway_resources.go
+++ b/controllers/submariner/gateway_resources.go
@@ -133,6 +133,19 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 		})
 	}
 
+	securityContext := &corev1.SecurityContext{
+		Capabilities: &corev1.Capabilities{
+			Add:  []corev1.Capability{"net_admin"},
+			Drop: []corev1.Capability{"all"},
+		},
+		// The gateway needs to be privileged so it can write to /proc/sys
+		AllowPrivilegeEscalation: ptr.To(true),
+		Privileged:               ptr.To(true),
+		RunAsNonRoot:             ptr.To(false),
+		// We need to be able to update /var/lib/alternatives (for iptables)
+		ReadOnlyRootFilesystem: ptr.To(false),
+	}
+
 	podTemplate := corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: podSelectorLabels,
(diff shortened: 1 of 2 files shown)


controllers/submariner/route_agent_resources.go
```diff
@@ -110,7 +110,6 @@ func newRouteAgentDaemonSet(cr *v1alpha1.Submariner, name string) *appsv1.Daemon
 							Name:            name + "-init",
 							Image:           getImagePath(cr, opnames.RouteAgentImage, names.RouteAgentComponent),
 							ImagePullPolicy: images.GetPullPolicy(cr.Spec.Version, cr.Spec.ImageOverrides[names.RouteAgentComponent]),
-							Command:         []string{"submariner-route-agent.sh"},
```
skitt, line 113: This removal is harmless, but was it intended as part of this PR?

controllers/submariner/gateway_resources.go
```diff
@@ -133,6 +133,19 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
...
+		Capabilities: &corev1.Capabilities{
+			Add:  []corev1.Capability{"net_admin"},
+			Drop: []corev1.Capability{"all"},
+		},
+		// The gateway needs to be privileged so it can write to /proc/sys
+		AllowPrivilegeEscalation: ptr.To(true),
+		Privileged:               ptr.To(true),
+		RunAsNonRoot:             ptr.To(false),
+		// We need to be able to update /var/lib/alternatives (for iptables)
+		ReadOnlyRootFilesystem: ptr.To(false),
```
skitt, line 146: This shouldn’t be necessary in the init container.

controllers/submariner/gateway_resources.go
```diff
@@ -133,6 +133,19 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
...
 	}
 
+	securityContext := &corev1.SecurityContext{
+		Capabilities: &corev1.Capabilities{
+			Add:  []corev1.Capability{"net_admin"},
+			Drop: []corev1.Capability{"all"},
+		},
+		// The gateway needs to be privileged so it can write to /proc/sys
+		AllowPrivilegeEscalation: ptr.To(true),
+		Privileged:               ptr.To(true),
```
skitt, line 143: Given how the gateway container image is set up currently, this is necessary in all cases (since the entrypoint script writes to `/proc/sys`), but ideally the init container shouldn’t need these.

skitt (commented):
I won’t object to this to let the bug fix in quickly, but I think the security context should be reworked, and perhaps we could even have a specific “wait for node readiness” binary.

controllers/submariner/route_agent_resources.go
tpantelis: It's cleanup that was missed before but technically should be in another PR.  I'll remove it.

tpantelis:
> I won’t object to this to let the bug fix in quickly, but I think the security context should be reworked, and perhaps we could even have a specific “wait for node readiness” binary.

I agree - it's not ideal for the init container. As you surmised, I added it b/c the entry script failed [here](https://github.com/submariner-io/submariner/blob/44272f008bee836d154ad35adb0ad57eb39466a0/package/submariner.sh#L17):

`/usr/local/bin/submariner.sh: line 17: /proc/sys/net/ipv4/conf/all/send_redirects: Read-only file system`

Perhaps all we need is `ReadOnlyRootFilesystem: ptr.To(false)`. 

Longer term, we could add a “wait for node readiness” binary to the image as you suggest and specify that for the init container command.

skitt (approved):

yboaron (approved):

PR #3285, opened 2024-11-28
Optimize the logic of finding cluster service/pod ip range

unilinu:
#### What type of PR is this?

Fix or Optimize

#### What this PR does / why we need it:

According to the original logic, do some optimizations:

1. Correct the label selector of kube-proxy from `component=kube-proxy` to `k8s-app=kube-proxy`

2. Similar to `findPodIPRange` function, add the redundant finding from kube-controller-manager.

### DIFF
--- a/pkg/discovery/network/generic.go
+++ b/pkg/discovery/network/generic.go
@@ -82,6 +82,11 @@ func findClusterIPRange(ctx context.Context, client controllerClient.Client) (st
 		return clusterIPRange, err
 	}
 
+	clusterIPRange, err = findClusterIPRangeFromKubeController(ctx, client)
+	if err != nil || clusterIPRange != "" {
+		return clusterIPRange, err
+	}
+
 	clusterIPRange, err = findClusterIPRangeFromServiceCreation(ctx, client)
 	if err != nil || clusterIPRange != "" {
 		return clusterIPRange, err
@@ -92,6 +97,10 @@ func findClusterIPRange(ctx context.Context, client controllerClient.Client) (st
 
 func findClusterIPRangeFromApiserver(ctx context.Context, client controllerClient.Client) (string, error) {
 	return FindPodCommandParameter(ctx, client, "component=kube-apiserver", "--service-cluster-ip-range")
+}
+
+func findClusterIPRangeFromKubeController(ctx context.Context, client controllerClient.Client) (string, error) {
+	return FindPodCommandParameter(ctx, client, "component=kube-controller-manager", "--service-cluster-ip-range")
 }
 
 func findClusterIPRangeFromServiceCreation(ctx context.Context, client controllerClient.Client) (string, error) {
@@ -153,12 +162,12 @@ func parseServiceCIDRFrom(msg string) (string, error) {
 }
 
 func findPodIPRange(ctx context.Context, client controllerClient.Client) (string, error) {
-	podIPRange, err := findPodIPRangeKubeController(ctx, client)
+	podIPRange, err := findPodIPRangeFromKubeController(ctx, client)
 	if err != nil || podIPRange != "" {
 		return podIPRange, err
 	}
 
-	podIPRange, err = findPodIPRangeKubeProxy(ctx, client)
+	podIPRange, err = findPodIPRangeFromKubeProxy(ctx, client)
 	if err != nil || podIPRange != "" {
 		return podIPRange, err
 	}
@@ -171,12 +180,12 @@ func findPodIPRange(ctx context.Context, client controllerClient.Client) (string
 	return "", nil
 }
 
-func findPodIPRangeKubeController(ctx context.Context, client controllerClient.Client) (string, error) {
+func findPodIPRangeFromKubeController(ctx context.Context, client controllerClient.Client) (string, error) {
 	return FindPodCommandParameter(ctx, client, "component=kube-controller-manager", "--cluster-cidr")
 }
 
-func findPodIPRangeKubeProxy(ctx context.Context, client controllerClient.Client) (string, error) {
-	return FindPodCommandParameter(ctx, client, "component=kube-proxy", "--cluster-cidr")
+func findPodIPRangeFromKubeProxy(ctx context.Context, client controllerClient.Client) (string, error) {
+	return FindPodCommandParameter(ctx, client, "k8s-app=kube-proxy", "--cluster-cidr")
 }
 
 func findPodIPRangeFromNodeSpec(ctx context.Context, client controllerClient.Client) (string, error) {
--- a/pkg/discovery/network/generic_test.go
+++ b/pkg/discovery/network/generic_test.go
@@ -120,16 +120,13 @@ var _ = Describe("Generic Network", func() {
 			Expect(clusterNet).NotTo(BeNil())
 		})
 
-		It("Should return the ClusterNetwork structure with pod CIDR", func() {
+		It("Should return the ClusterNetwork structure with both CIDR", func() {
 			Expect(clusterNet.PodCIDRs).To(Equal([]string{testPodCIDR}))
+			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDR}))
 		})
 
 		It("Should identify the networkplugin as generic", func() {
 			Expect(clusterNet.NetworkPlugin).To(BeIdenticalTo(cni.Generic))
-		})
-
-		It("Should return the ClusterNetwork structure with the service CIDR", func() {
-			Expect(clusterNet.ServiceCIDRs).To(Equal([]string{testServiceCIDRFromService}))
 		})
 	})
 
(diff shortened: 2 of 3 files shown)


pkg/discovery/network/generic.go
```diff
@@ -78,22 +77,31 @@ func discoverNetwork(ctx context.Context, client controllerClient.Client) (*Clus
 
 func findClusterIPRange(ctx context.Context, client controllerClient.Client) (string, error) {
 	clusterIPRange, err := findClusterIPRangeFromApiserver(ctx, client)
-	if err != nil || clusterIPRange != "" {
-		return clusterIPRange, err
+	if clusterIPRange != "" {
```
tpantelis: Why did you remove the error checking here and other places below? This hides the error which isn't best practice. An error here means it could not determine if the IP range could be obtained, most likely because the API server wasn't available, in which case we want to abort and report the error.

tpantelis (changes requested):

pkg/discovery/network/generic.go
unilinu: Well, It seems I understand wrong the purpose of these error checking. I thought this function is to detect the available ClusterIP range, but it seems to detect the availability of component/service related to ClusterIP. I will cancel the related modifications later. 

In addition, I think the point 1&2 should be retained, but for point 3, I wonder if add the redundant finding from kube-controller-manager.

unilinu: @tpantelis Another question, the original if conditions `err != nil || clusterIPRange != ""` is almost true, which will cover/skip the following judging!!
In my opinion, this function is used to find the service cluster ip range. Even if occurs error, which should be detected and handled somewhere, adding a waring/error log but continuing finding is a good idea here.

If you agree, I will modify it as above. If not, looking forward to your reply.

tpantelis: > In my opinion, this function is used to find the service cluster ip range. Even if occurs error, which should be detected and handled somewhere, adding a waring/error log but continuing finding is a good idea here.
>
> If you agree, I will modify it as above. If not, looking forward to your reply.

I don't agree. Consider this scenario:

Let's say in a particular environment the first call, `findClusterIPRangeFromApiserver`, under  normal circumstances with no error finds the cluster IP. But on a particular run a transient error occurs, eg the API server is busy and times out, in which case we would just log a warning and continue. For the rest of the code below, the API server responds and no error occurs but the cluster IP isn't found by any of the following methods. This function would return and report to the caller that no cluster IP was found and no error occurred. But this is incorrect as an error did occur but it was essentially ignored by just logging it. We must propagate errors to the caller to indicate that the result was indeterminate and to let the caller handle the error as it sees fit.

unilinu: I can understand it. So, each error should be checked and not be covered by assignments. However, there is no doubt that `err != nil || clusterIPRange != ""` condition will cover/skip the following statements. This point must be corrected. I'll modify the PR correspondingly.

tpantelis: No - the error checking was fine as is. Given your recent changes and the scenario I outlined above, it would return a valid cluster IP **and** an error which is non-sensical and non-idiomatic.

Please just focus on the purpose of this PR as outlined in your summary and not modify the existing error handling.

unilinu:
Thanks for your patience firstly. In fact, I wonder why the error checking was fine. Is an error or a valid cluster ip always return true? Thus, what effect the following methods make?在 2024年12月3日，20:48，Thomas Pantelis ***@***.***> 写道：﻿
@tpantelis commented on this pull request.



In pkg/discovery/network/generic.go:
> @@ -78,22 +77,31 @@ func discoverNetwork(ctx context.Context, client controllerClient.Client) (*Clus
 
 func findClusterIPRange(ctx context.Context, client controllerClient.Client) (string, error) {
 	clusterIPRange, err := findClusterIPRangeFromApiserver(ctx, client)
-	if err != nil || clusterIPRange != "" {
-		return clusterIPRange, err
+	if clusterIPRange != "" {

No - the error checking was fine as is. Given your recent changes and the scenario I outlined above, it would return a valid cluster IP and an error which is non-sensical and non-idiomatic.

pkg/discovery/network/generic.go
unilinu: Thanks for your patience. I know what I misunderstood. `err != nil || clusterIPRange != ""` is not always true. There is a situation not finding cluster ip range from the args of apiserver/controller-manager. 

I have modified my patch, removing the changes of error checking.

pkg/discovery/network/generic.go
```diff
@@ -153,50 +162,23 @@ func parseServiceCIDRFrom(msg string) (string, error) {
...
-	return parseToPodCidr(nodes.Items)
-}
-
-func parseToPodCidr(nodes []corev1.Node) (string, error) {
-	// In K8s, each node is typically assigned a unique PodCIDR range for the pods that run on that node.
-	// Each node's PodCIDR is used to allocate IP addresses to the pods scheduled on that node. Only if
-	// the cluster is a single node deployment, we should rely on the node.Spec.PodCIDR as podCIDR of the cluster.
-	if len(nodes) == 1 {
-		return nodes[0].Spec.PodCIDR, nil
-	}
```
tpantelis: From the PR summary:

> The node.spec.podCIDR field is not the cluster pod ip range, but the node-assigned.

But this is only considered if there's only one node as explained in the comments above and if the pod CIDR wasn't found from any prior methods. It seems valid to keep this.

tpantelis (changes requested):

tpantelis:
@unilinu The DCO job failed - you need to add Signed-off-by your commit.

pkg/discovery/network/generic.go
unilinu: You are right. I have fixed it.

tpantelis (approved):

tpantelis (changes requested):
Please add unit tests to cover the new code you added.

tpantelis:
[Linting/Commit Message](https://github.com/submariner-io/submariner-operator/actions/runs/12154031393/job/33911594332?pr=3285) job failed.

Also unit tests failed.

unilinu:
> [Linting/Commit Message](https://github.com/submariner-io/submariner-operator/actions/runs/12154031393/job/33911594332?pr=3285) job failed.
> 
> Also unit tests failed.

Added and fixed.

pkg/discovery/network/generic_test.go
```diff
@@ -120,30 +120,33 @@ var _ = Describe("Generic Network", func() {
...
-		})
 	})
 
 	When("There is a kube-controller pod with the right parameter passed as an Arg", func() {
 		BeforeEach(func(ctx SpecContext) {
 			clusterNet = testDiscoverGenericWith(
 				ctx,
-				fakePodWithArg("kube-controller-manager", []string{"kube-controller-manager"}, "--cluster-cidr="+testPodCIDR),
+				fakePodWithArg("kube-controller-manager", []string{"kube-controller-manager"},
+					[]string{"--cluster-cidr=" + testPodCIDR, "--service-cluster-ip-range=" + testServiceCIDR}),
```
tpantelis, line 138: Should be able to use:

```suggestion
				fakeKubeControllerManagerPod(),
```

pkg/discovery/network/generic_test.go
```diff
@@ -120,30 +120,33 @@ var _ = Describe("Generic Network", func() {
 			Expect(clusterNet).NotTo(BeNil())
 		})
 
-		It("Should return the ClusterNetwork structure with pod CIDR", func() {
+		It("Should return the ClusterNetwork structure with all CIDR", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with both CIDRs", func() {
```

pkg/discovery/network/generic_test.go
```diff
@@ -120,30 +120,33 @@ var _ = Describe("Generic Network", func() {
...
 				ctx,
-				fakePodWithArg("kube-controller-manager", []string{"kube-controller-manager"}, "--cluster-cidr="+testPodCIDR),
+				fakePodWithArg("kube-controller-manager", []string{"kube-controller-manager"},
+					[]string{"--cluster-cidr=" + testPodCIDR, "--service-cluster-ip-range=" + testServiceCIDR}),
 			)
 			Expect(clusterNet).NotTo(BeNil())
 		})
 
-		It("Should return the ClusterNetwork structure with pod CIDR", func() {
+		It("Should return the ClusterNetwork structure with all CIDR", func() {
```
tpantelis:
```suggestion
		It("Should return the ClusterNetwork structure with both CIDRs", func() {
```

tpantelis (changes requested):

tpantelis:
[Linting/Commit Message](https://github.com/submariner-io/submariner-operator/actions/runs/12174365280/job/33973867105?pr=3285) job is still failing. The commit title is too long.

There's also [Go linting errors](https://github.com/submariner-io/submariner-operator/actions/runs/12174365280/job/33973867474?pr=3285).

dfarrell07:
Thanks for the PR! Please check the list of CI jobs and see the details of the ones that failed.

(posted at the same time as Tom above, sorry for the repeat)

pkg/discovery/network/generic_test.go
unilinu: The test case is "There is a kube-controller pod with the right parameter **passed as an Arg**". The `fakeKubeControllerManagerPod` function should be and have been used in the last one naming "There is a kube-controller pod with the right parameter".

unilinu:
> [Linting/Commit Message](https://github.com/submariner-io/submariner-operator/actions/runs/12174365280/job/33973867105?pr=3285) job is still failing. The commit title is too long.
> 
> There's also [Go linting errors](https://github.com/submariner-io/submariner-operator/actions/runs/12174365280/job/33973867474?pr=3285).

Thans for your patience. The errors you mentioned have been fixed.

tpantelis (approved):

skitt (approved):

PR #3292, opened 2024-12-02
Set up a base distribution in the images

skitt:
This adds the minimum subset of packages required for Fedora, allowing scanners to understand the image and process it correctly (in exchange for a small size increase).

### DIFF
--- a/package/dnf_install
+++ b/package/dnf_install
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Installs packages using dnf to a named root:
+# -a arch - use arch instead of the native arch
+# -k      - keep the package cache
+# -r root - install to the named root instead of /output/base
+# -v ver  - use the given Fedora version (required)
+#
+# %arch in the package references will be replaced with the chosen arch
+
+INSTALL_ROOT=/output/base
+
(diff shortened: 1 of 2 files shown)


package/dnf_install
```diff
@@ -0,0 +1,62 @@
...
+# -k      - keep the package cache
+# -r root - install to the named root instead of /output/base
+# -v ver  - use the given Fedora version (required)
+#
+# %arch in the package references will be replaced with the chosen arch
+
+INSTALL_ROOT=/output/base
+
+# Limit the number of files so that dnf doesn't spend ages processing fds
+ulimit -n 1048576
```
dfarrell07, line 14: Should this have the conditional to prevent trying to raise the limit?

https://github.com/submariner-io/submariner/pull/3241/files#diff-97c33edf909bead8e6d1991b438dcae867427a84e03274fc08cd5edb45de9362R16

skitt: Yup, what I really need to do is move `dnf_install` to Shipyard and unify it.

dfarrell07 (approved):

tpantelis (approved):

PR #3307, opened 2024-12-19
Change loadbalancer type for HCP deployments

aswinsuryan:
The Kubevirt(HCP) clusters do not have their loadbalancer installedand rely on the host loadblancer( like netallb) when a loadbalancer is created.So when the Submariner exposes the submariner-gateway loadbalancer service with ServiceExternalTrafficPolicy locally does not work, as the traffic from the remotecluster first lands on the host cluster. Hence the traffic policy is updated to clusterin these deployments.

### DIFF
--- a/api/v1alpha1/submariner_types.go
+++ b/api/v1alpha1/submariner_types.go
@@ -160,9 +160,13 @@ type SubmarinerSpec struct {
 	// Enable NAT between clusters.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enable NAT"
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
-	NatEnabled bool `json:"natEnabled"`
-
+	NatEnabled          bool `json:"natEnabled"`
 	AirGappedDeployment bool `json:"airGappedDeployment,omitempty"`
+
+	// Is the cluster a hosted cluster.
+	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Hosted Cluster"
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	HostedCluster bool `json:"hostedCluster,omitempty"`
 
 	// Enable automatic Load Balancer in front of the gateways.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enable Load Balancer"
@@ -227,6 +231,11 @@ type SubmarinerStatus struct {
 	NatEnabled bool `json:"natEnabled"`
 
 	AirGappedDeployment bool `json:"airGappedDeployment,omitempty"`
+
+	// Is the cluster a hosted cluster.
+	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Hosted Cluster"
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	HostedCluster bool `json:"hostedCluster,omitempty"`
 
 	ColorCodes string `json:"colorCodes,omitempty"`
 
--- a/internal/controllers/submariner/loadbalancer_resources.go
+++ b/internal/controllers/submariner/loadbalancer_resources.go
@@ -91,6 +91,8 @@ func (r *Reconciler) getOCPPlatformType(ctx context.Context) (string, error) {
 }
 
 func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP string) *corev1.Service {
+	externalTrafficPolicy := corev1.ServiceExternalTrafficPolicyTypeLocal
+
 	var svcAnnotations map[string]string
 
 	switch platformTypeOCP {
@@ -104,6 +106,10 @@ func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP strin
 			"service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type":                   "public",
 			"service.kubernetes.io/ibm-load-balancer-cloud-provider-vpc-health-check-protocol": "http",
 		}
+	case string(configv1.KubevirtPlatformType):
+		if instance.Spec.HostedCluster {
+			externalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeCluster
+		}
 	default:
 		svcAnnotations = map[string]string{}
 	}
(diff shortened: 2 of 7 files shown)


internal/controllers/submariner/loadbalancer_resources.go
```diff
@@ -104,6 +106,8 @@ func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP strin
 			"service.kubernetes.io/ibm-load-balancer-cloud-provider-ip-type":                   "public",
 			"service.kubernetes.io/ibm-load-balancer-cloud-provider-vpc-health-check-protocol": "http",
 		}
+	case string(configv1.KubevirtPlatformType):
```
yboaron, line 109: Does setting externalTrafficPolicy to corev1.ServiceExternalTrafficPolicyTypeCluster will work for both kubevirt and HCP Kubevirt clusters ?  or it is needed only for HCP

yboaron: IIUC [1] , setting externalTrafficPolicy to corev1.ServiceExternalTrafficPolicyTypeCluster may cause the client's IP address to change. Could this break Submariner inter-cluster traffic?

[1] 
https://access.redhat.com/solutions/7028639

aswinsuryan: In this particular case IP address was preserved, maybe we should check the setup again. There were two level of load balancer service one created in the HCP cluster and in response to that one will be created in the hosted cluster.

The platform type was kubevirt , do we have another way to identify kubevirt vs HCP kubevirt?

yboaron: AFAIK, several fields in managedcluster resource are checked to decide if cluster is hosted, check [1] .

[1] 
https://github.com/stolostron/console/blob/eba0d8336bb3b62313fa841dae3749b4df6e31d8/frontend/src/resources/utils/get-cluster.ts#L575

aswinsuryan: @yboaron  So how about non-ACM clusters ?  This was the clusterVersion in this particular cluster


 ```
apiVersion: config.openshift.io/v1
kind: Infrastructure
metadata:
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    release.openshift.io/create-only: "true"

  generation: 2
  labels:
    hypershift.openshift.io/managed: "true"
  name: cluster
  ownerReferences:
  - apiVersion: config.openshift.io/v1
    kind: ClusterVersion
spec:
  cloudConfig:
    name: ""
  platformSpec:
    type: KubeVirt
status:
  apiServerInternalURI: <>
  apiServerURL:<>
  controlPlaneTopology: External
  cpuPartitioning: None
  etcdDiscoveryDomain: <>
  infrastructureName:<>
  infrastructureTopology: HighlyAvailable
  platform: KubeVirt
  platformStatus:
    type: KubeVirt
```

Comment (yboaron) [reply internal/controllers/submariner/loadbalancer_resources.go:109]: Don't think we can conclude whether cluster is hosted or not from ^^ cluster infra resource.

So, trying to summarize:
 1.  Current code supports loadbalancer for non-hosted platforms, including  Kubevit non-hosted case 
 2. This PR will address loadbalancer issue for hosted Kubevirt and may break loadbalancer for non-hosted kubevirt.

I think we can continue with this PR as a short-term solution and document that this change may break load balancer for non-hosted Kubevirt.

For a long-term solution, if this issue is relevant to all hosted platforms, we might consider adding a new field to submarinerconfig and submariner and configuring the loadbalancer accordingly.

WDYT?

Comment (aswinsuryan) [reply internal/controllers/submariner/loadbalancer_resources.go:109]: Added the field to Submariner. will do the required changes in subctl and addon once this goes in.

File: internal/controllers/submariner/loadbalancer_resources.go [hunk 2]
```diff
@@ -108,14 +110,18 @@ func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP strin
 		svcAnnotations = map[string]string{}
 	}
 
+	if instance.Spec.HostedCluster {
```
yboaron: Thanks for adding the flag to Submariner CRD.
Overall code LGTM , 
Only wondering 
A. If we should apply this change only for hosted KubeVirt at first until we verify it also for other hosted platforms (e.g hosted ROSA) 
B. if this new field might affect upgrade ?

internal/controllers/submariner/loadbalancer_resources.go
yboaron: Do you think we can wait with adding the required changes in subctl and addon until we'll better understand the requirements related to hosted clusters? 

Menawhile, users that deploy Submariner using Subctl should edit submariner/isHosted  fields

api/v1alpha1/submariner_types.go
```diff
@@ -228,6 +232,11 @@ type SubmarinerStatus struct {
 
 	AirGappedDeployment bool `json:"airGappedDeployment,omitempty"`
 
+	// Is the cluster a hosted cluster.
+	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Hosted Cluster"
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	HostedCluster bool `json:"hostedCluster,omitempty"`
```
tpantelis, line 238: This should be set by the controller along with the other status fields.

tpantelis (changes requested):

api/v1alpha1/submariner_types.go
aswinsuryan: Done.

internal/controllers/submariner/loadbalancer_resources.go
aswinsuryan: A) Chances of working with external traffic policy cluster is higher in that case right ? As they might have similar implementations. We can add a separate check if needed for any platforms.
B) It is an optional field and it should take the default value false in an upgraded cluster.

tpantelis (approved):

internal/controllers/submariner/loadbalancer_resources.go
yboaron: A. I successfully deployed Submariner on ROSA HCP a few months ago (with current LB code), so not sure how setting external traffic policy to cluster will affect ROSA HCP.  I think we should consider making this change only for hosted Kubevirt at this point

B. I meant upgrading HCP cluster use case, anyway I think we can address that once we know what logic we are adding in the hosted case

aswinsuryan: Done.

yboaron (approved):

PR #3369, opened 2025-02-11
Remove remnants of the networkplugin syncer

skitt:
The syncer was removed in 0.16, it’s safe to stop handling migrations now.


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
Do we think we need release notes for this?

skitt:
> Do we think we need release notes for this?

For users it means that it won’t be possible to upgrade from 0.15 to 0.21, but by the time 0.21 is released, 0.15 will have been out of support for a while. But it’s easier to add a release note than it is to debate this point :wink:.

dfarrell07:
This is for 0.21, was cut after 0.20.

PR #3376, opened 2025-02-17
Switch to Go 1.24 tool handling

skitt:
This returns tool dependencies to the main module, as indirect dependencies, but this shouldn't have an impact on downstreams thanks to module pruning. The big advantage of this change is that the tools used during the build always match the versions declared in go.mod; there is no longer the possibility that outdated binaries in bin are used instead.

I’m still debating the merits of this. The main benefit is described above, but it has the disadvantage of adding all the tools’ dependencies to the main dependency tree, which affects the module trees of all the project’s descendants. In any case it requires Go 1.24 everywhere which we don’t have yet.


dfarrell07:
@skitt talked us through this change, the move to `go tool`, on the scrub today. It looks very cool, but there are still some trade-offs he's considering. He can add details.

skitt:
I prefer the approach taken in #3576.

PR #3388, opened 2025-03-05
Add nftables enabled flag to Submariner

yboaron:
We want to begin testing Submariner with nftables, starting with Kind upstream.
To do this, we need to modify RouteAgent and GlobalNet to dynamically set the required packet filter driver (iptables or nftables), as it is currently hardcoded to iptables.

### DIFF
--- a/api/v1alpha1/submariner_types.go
+++ b/api/v1alpha1/submariner_types.go
@@ -168,6 +168,12 @@ type SubmarinerSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	HostedCluster bool `json:"hostedCluster,omitempty"`
 
+	// Use nftables for programming IP filter rules.
+	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Use Nftables"
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	// +optional
+	UseNftables bool `json:"useNftables,omitempty"`
+
 	// Enable automatic Load Balancer in front of the gateways.
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Enable Load Balancer"
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
--- a/config/crd/bases/submariner.io_submariners.yaml
+++ b/config/crd/bases/submariner.io_submariners.yaml
@@ -214,6 +214,9 @@ spec:
                       type: string
                   type: object
                 type: array
+              useNftables:
+                description: Use nftables for programming IP filter rules.
+                type: boolean
               version:
                 description: The image tag.
                 type: string
--- a/internal/controllers/submariner/globalnet_resources.go
+++ b/internal/controllers/submariner/globalnet_resources.go
@@ -20,6 +20,7 @@ package submariner
 
 import (
 	"context"
+	"strconv"
 
 	"github.com/go-logr/logr"
 	"github.com/submariner-io/admiral/pkg/names"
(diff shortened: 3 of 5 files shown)


api/v1alpha1/submariner_types.go
```diff
@@ -218,6 +218,11 @@ type SubmarinerSpec struct {
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Is Nftables enabled.
+	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Nftables Enabled"
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	NftablesEnabled bool `json:"nftablesEnabled,omitempty"`
```
tpantelis, line 225:
```suggestion
	UseNftables bool `json:"nftablesEnabled,omitempty"`
```

tpantelis:
```suggestion
	// Use nftables for programming IP rules.
	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Use Nftables"
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
	// +optional
	UseNftables bool `json:"useNftables,omitempty"`
```

tpantelis (changes requested):

skitt:
Is there any way to avoid having a flag for this? I’m wondering about three things:

* Having a flag means potentially having to manually handle it during upgrades. If `nftables` isn’t set, does it mean the user explicitly wants `iptables`, or just the default, whatever that is? What should happen when the user upgrades from a version of OpenShift with `iptables` to a version with `nftables`?
* Having a flag that has to match system configuration means that it could become incorrect.
* What happens if the cluster has nodes with `iptables`, and others with `nftables`?

I would much prefer a system where Submariner auto-detects which tool to use, in each component (not in the operator).

skitt:
Ah, I see this was discussed on https://kubernetes.slack.com/archives/C010RJV694M/p1741176995986509 — could we ask OVNK if there’s a “kubeproxy mode” equivalent there that could be used?

skitt:
Looking at https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_shared_intf.go and https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_nftables.go#L22 suggests that it isn’t all that important, unless rule ordering is significant. The general approach seems to be to use `iptables` if there are `iptables` rules, `nftables` if there are `nftables` rules, except for UDN which seems to only use NFT.

yboaron:
> Looking at https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_shared_intf.go and https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_nftables.go#L22 suggests that it isn’t all that important, unless rule ordering is significant. The general approach seems to be to use `iptables` if there are `iptables` rules, `nftables` if there are `nftables` rules, except for UDN which seems to only use NFT.

I think we should move to nftables ASAP, to allow us to deal with issues like  [1].

But this should be done carefully,  I was thinking to continue with iptables for now, but at least run some CI gates with nftables before we move to nftables.

We also need to make sure that Submariner with nftables works with OCP 4.17. OCP 4.16 (OVNK use iptables), I think adding the field will allow us flexibility during the transition period.

[1] 
https://github.com/submariner-io/submariner/issues/3307

tpantelis:
> We also need to make sure that Submariner with nftables works with OCP 4.17. OCP 4.16 (OVNK use iptables), I think adding the field will allow us flexibility during the transition period.

Once the field is added, it's public-facing and documented and problematic to remove.  If we want to be able to enable nftables during a transition period, then I suggest we add a `ConfigMap` (undocumented) to configure it. We can also use the `ConfigMap` for other config params in the future,  eg work queue settings. We don't have to create the `ConfigMap` out-of-box, just add support for it in the code.

yboaron:
> > We also need to make sure that Submariner with nftables works with OCP 4.17. OCP 4.16 (OVNK use iptables), I think adding the field will allow us flexibility during the transition period.
> 
> Once the field is added, it's public-facing and documented and problematic to remove. If we want to be able to enable nftables during a transition period, then I suggest we add a `ConfigMap` (undocumented) to configure it. We can also use the `ConfigMap` for other config params in the future, eg work queue settings. We don't have to create the `ConfigMap` out-of-box, just add support for it in the code.

Yes, I guess we will have to depercate/delete the new field after the migration to nftables is completed, it makes sense to use configmap. I will close this PR

yboaron:
we'll use configmap instead of new field in submariner spec

PR #3487, opened 2025-06-11
Add node update to routeAgent RBAC

yboaron:
RouteAgent should add remote subnets to node annotation to prevent SNAT by OVNK.

### DIFF
--- a/config/rbac/submariner-route-agent/cluster_role.yaml
+++ b/config/rbac/submariner-route-agent/cluster_role.yaml
@@ -39,6 +39,8 @@ rules:
       - get
       - list
       - watch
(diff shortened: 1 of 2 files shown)


yboaron:
@tpantelis  QQ: should I change the verb to  **patch**  to allow updating node annotations or keep the **update** verb ?

skitt (approved):

skitt:
> @tpantelis QQ: should I change the verb to **patch** to allow updating node annotations or keep the **update** verb ?

That depends on how the annotations are updated. Patch would be better but that’s a different implementation.

tpantelis (approved):

PR #3505, opened 2025-06-25
Add Konflux Tekton task to label bundle

dfarrell07:
Adds labels to the Konflux-built bundle paralleling the current bundle.

All of these labels may not be needed. After testing the full system we an iterate on removing/modernizing them.


dfarrell07:
Weird, it seems like if there's an YAML syntax error at any point then Konflux gives up on the change and doesn't run against updated commits?

dfarrell07:
I guess I'll just close and make another PR, it doesn't seem like Konflux is going to run here.

PR #3518, opened 2025-07-01
Combine OCP network discovery CIDRs into comma-separated string

yboaron:
Currently, during OCP network discovery, multiple Pod CIDRs are handled as separate strings.

This fix updates OCP network discovery logic to join all discovered CIDRs into a single comma-separated string to align with the expected format.

### DIFF
--- a/pkg/discovery/network/openshift4.go
+++ b/pkg/discovery/network/openshift4.go
@@ -21,6 +21,7 @@ package network
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


tpantelis (changes requested):
> This fix updates OCP network discovery logic to join all discovered CIDRs into a single comma-separated string to align with the expected format.

But this is not really the expected format - it just happened that way with the Kind CIDR discovery. 

The network discovery code defines `PodCIDRs` as an array so I think we should leave the OCP code as is and fix the operator controller code to properly handle multiple entries.

PR #3562, opened 2025-07-22
Allow "make bundle" to run without Shipyard

skitt:
Bundle construction is supposed to work without Shipyard, but a couple of fixes are required:

* don't fail the build if Shipyard's Makefile.inc isn't available;
* provide the version information without Shipyard.

The bundle GHA is changed to skip Shipyard, to ensure this isn't broken again in future. Since this involves shell constructs that rely on bash, the Makefile is configured to use bash as its shell (GHAs use Ubuntu which defaults to dash). The bundle build also needs at least the last two tags tied to the base branch in the repository, so the corresponding checkout operation is changed to fetch the repository in its entirety.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -39,8 +39,10 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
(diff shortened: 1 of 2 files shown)


dfarrell07 (approved):
Nice, thanks Stephen!

tpantelis (approved):

PR #3564, opened 2025-07-23
Use Go embedding for YAML files instead of yamls2go

skitt:
See individual commits.

Fixes #1646.

PR #3566, opened 2025-07-24
Commit operator bundle to the repository

dfarrell07:
Transition from dynamically generating the operator bundle during the build to storing the pre-generated bundle directly in the repository.

This is a standard practice and offers several advantages:

* Makes updates to the bundle more visible.
* Reduces CI resource usage by avoiding downloading tools and rebuilding the bundle. A dedicated job will verify the bundle is up-to-date.

It's particularly important in the context of hermetic builds. The dynamic approach with `make bundle` is difficult in hermetic builds:

* The `make bundle` command requires `kustomize` and `operator-sdk`. Building `operator-sdk` from source requires the `gpgme-devel` RPM.
* The generic run-script task used for the bundle generation step does not automatically configure the build environment to use prefetched dependencies. This affects both Go modules and RPMs. Because the RPM cache isn't configured, `dnf install gpgme-devel` fails.
* Attempts to manually create a local DNF repository from the cached files also fail due to a catch-22 problem with needing `createrepo_c` to create a repository index before DNF could be used to install it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 .dapper
 bin/
 build/_output/
-bundle/
-bundle.Dockerfile
 dist/
 deploy/crds
 deploy/submariner/crds
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -1,0 +1,21 @@
+FROM scratch
+
+# Core bundle labels.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=submariner
+LABEL operators.operatorframework.io.bundle.channels.v1=alpha-0.21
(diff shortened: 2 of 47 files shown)


vthapar (approved):

tpantelis:
Should we have a CI job that runs `make bundle` and checks for diffs similar to https://github.com/submariner-io/submariner/blob/406e43fd6374b435f8f621401744b3e89476c47b/.github/workflows/linting.yml#L46?

skitt:
> Should we have a CI job that runs `make bundle` and checks for diffs similar to https://github.com/submariner-io/submariner/blob/406e43fd6374b435f8f621401744b3e89476c47b/.github/workflows/linting.yml#L46?

Yes, I was waiting for this to add a check in https://github.com/submariner-io/submariner-operator/pull/3564. But it would make sense here as an addition to the bundle validation job.

skitt:
> Yes, I was waiting for this to add a check in #3564. But it would make sense here as an addition to the bundle validation job.

Let’s do that as a separate PR though so it can easily be backported to 0.21 again.

skitt (approved):

PR #3568, opened 2025-07-25
Adjust the bundle for the devel branch

skitt:

PR #3569, opened 2025-07-25
Verify the bundle in CI

skitt:
This checks that nothing unexpected has changed when building the bundle.


skitt:
This is supposed to fail initially, it will pass once #3568 is merged.

tpantelis (approved):

yboaron (approved):

PR #3599, opened 2025-08-11
Align ENV vars set by CSV to Admiral image naming, simplify bundle customizations

dfarrell07:

### DIFF
--- a/config/manager/patches/related-images.deployment.config.yaml
+++ b/config/manager/patches/related-images.deployment.config.yaml
@@ -39,3 +39,6 @@
   value:
     name: RELATED_IMAGE_submariner-nettest
     value: registry.redhat.io/rhacm2/nettest-rhel9@sha256:2287b528b2222f3722c8cdfca5830d77ee8af2b918e3941887c86767512c0000
+- op: replace
+  path: /spec/template/spec/containers/0/image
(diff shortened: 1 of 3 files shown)


config/manager/patches/related-images.deployment.config.yaml
```diff
@@ -3,29 +3,42 @@
   path: /spec/template/spec/containers/0/env/-
   value:
     name: RELATED_IMAGE_submariner-operator
-    value: quay.io/submariner/submariner-operator:$(VERSION)
+    value: registry.redhat.io/rhacm2/submariner-rhel9-operator@sha256:cb49afacf892743e5e7552502903a507651bf7c3cd68661d927132b12e3c926b
```
vthapar, line 6: Do we need this change? This doesn't look like get applied for downstream, CSV likely takes precedence.

This looks like used for bundle based upstream deployment, which we don't fully support yet. Most likely this ends up ignored for downstream.

vthapar (approved):

config/manager/patches/related-images.deployment.config.yaml
dfarrell07: This is what causes the CSV to be generated with that config, this is the source file.

Jaanki (approved):

PR #3609, opened 2025-08-13
Revert change to version handling in Makefile

dfarrell07:
Also remove the Dockerfile refs to rc0, although that's unrelated to the build failure.


dfarrell07:
Basically I "fixed" it to allow vN.N.N-rcN to be a semantic version and it turns out it needed to fail and find a SHA for the release job.

tpantelis (approved):

dfarrell07:
I think this version var removal should be forward-ported.

PR #3616, opened 2025-08-21
Adjust the manifests target to generate yamls in deploy/crds

tpantelis:
...and copy them to _config/crd/bases_ to ensure the two copies are in sync so the `CRDs up-to-date` GHA catches changes to both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,8 +178,8 @@ deploy/crds/submariner.io_brokers.yaml deploy/crds/submariner.io_submariners.yam
 	test -f $@
 
 # Generate manifests e.g. CRD etc.
-manifests: $(CONTROLLER_DEEPCOPY) $(CONTROLLER_GEN)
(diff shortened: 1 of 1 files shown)


yboaron (approved):

Makefile
```diff
@@ -178,8 +178,8 @@ deploy/crds/submariner.io_brokers.yaml deploy/crds/submariner.io_submariners.yam
 	test -f $@
 
 # Generate manifests e.g. CRD etc.
-manifests: $(CONTROLLER_DEEPCOPY) $(CONTROLLER_GEN)
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) paths="./..." output:crd:artifacts:config=config/crd/bases
+manifests: $(CONTROLLER_DEEPCOPY) generatedyamls
```
skitt:
```suggestion
manifests: $(CONTROLLER_DEEPCOPY) $(GENERATED_YAMLS)
```

so that Make can check concrete files instead of a phony target.

skitt (changes requested):

skitt (approved):

PR #3621, opened 2025-08-26
Add changes  to act as a CA

aswinsuryan:
To support ovn-ipsec, the broker need to act as the CA. It
creates a CA certificate on startup and signs all CSR on
request.

### DIFF
--- a/bundle/manifests/submariner.clusterserviceversion.yaml
+++ b/bundle/manifests/submariner.clusterserviceversion.yaml
@@ -35,7 +35,7 @@ metadata:
             "globalnetEnabled": false,
             "namespace": "submariner-operator",
             "repository": "quay.io/submariner",
-            "version": "devel-a207299c4c57"
+            "version": "devel-790a1bb1e52a"
           }
         },
         {
@@ -69,7 +69,7 @@ metadata:
             "repository": "quay.io/submariner",
             "serviceCIDR": "192.168.66.0/24",
             "serviceDiscoveryEnabled": true,
-            "version": "devel-a207299c4c57"
+            "version": "devel-790a1bb1e52a"
           }
         }
       ]
@@ -77,7 +77,7 @@ metadata:
     categories: Networking
     certified: "false"
     containerImage: $(SUBMARINER_OPERATOR_IMAGE)
-    createdAt: "2025-10-01 18:13:16"
+    createdAt: "2025-10-06 10:40:39"
     description: Creates and manages Submariner deployments.
     operatorframework.io/suggested-namespace: submariner-operator
     operators.operatorframework.io/builder: operator-sdk-v1.41.1
@@ -690,18 +690,18 @@ spec:
                 - name: OPERATOR_NAME
                   value: submariner-operator
                 - name: RELATED_IMAGE_submariner-operator
-                  value: quay.io/submariner/submariner-operator:devel-a207299c4c57
+                  value: quay.io/submariner/submariner-operator:devel-790a1bb1e52a
                 - name: RELATED_IMAGE_submariner-gateway
-                  value: quay.io/submariner/submariner-gateway:devel-a207299c4c57
+                  value: quay.io/submariner/submariner-gateway:devel-790a1bb1e52a
                 - name: RELATED_IMAGE_submariner-route-agent
-                  value: quay.io/submariner/submariner-route-agent:devel-a207299c4c57
+                  value: quay.io/submariner/submariner-route-agent:devel-790a1bb1e52a
                 - name: RELATED_IMAGE_submariner-globalnet
-                  value: quay.io/submariner/submariner-globalnet:devel-a207299c4c57
+                  value: quay.io/submariner/submariner-globalnet:devel-790a1bb1e52a
                 - name: RELATED_IMAGE_lighthouse-agent
-                  value: quay.io/submariner/lighthouse-agent:devel-a207299c4c57
+                  value: quay.io/submariner/lighthouse-agent:devel-790a1bb1e52a
                 - name: RELATED_IMAGE_lighthouse-coredns
-                  value: quay.io/submariner/lighthouse-coredns:devel-a207299c4c57
-                image: quay.io/submariner/submariner-operator:devel-a207299c4c57
+                  value: quay.io/submariner/lighthouse-coredns:devel-790a1bb1e52a
+                image: quay.io/submariner/submariner-operator:devel-790a1bb1e52a
                 imagePullPolicy: Always
                 name: submariner-operator
                 resources:
@@ -738,6 +738,8 @@ spec:
           - secrets
           verbs:
           - get
+          - list
+          - watch
           - create
           - update
           - delete
@@ -852,17 +854,17 @@ spec:
   provider:
     name: submariner.io
   relatedImages:
-  - image: quay.io/submariner/submariner-operator:devel-a207299c4c57
+  - image: quay.io/submariner/submariner-operator:devel-790a1bb1e52a
     name: submariner-operator
-  - image: quay.io/submariner/submariner-gateway:devel-a207299c4c57
+  - image: quay.io/submariner/submariner-gateway:devel-790a1bb1e52a
     name: submariner-gateway
-  - image: quay.io/submariner/submariner-route-agent:devel-a207299c4c57
+  - image: quay.io/submariner/submariner-route-agent:devel-790a1bb1e52a
     name: submariner-route-agent
-  - image: quay.io/submariner/submariner-globalnet:devel-a207299c4c57
+  - image: quay.io/submariner/submariner-globalnet:devel-790a1bb1e52a
     name: submariner-globalnet
-  - image: quay.io/submariner/lighthouse-agent:devel-a207299c4c57
+  - image: quay.io/submariner/lighthouse-agent:devel-790a1bb1e52a
     name: lighthouse-agent
-  - image: quay.io/submariner/lighthouse-coredns:devel-a207299c4c57
+  - image: quay.io/submariner/lighthouse-coredns:devel-790a1bb1e52a
     name: lighthouse-coredns
   selector:
     matchLabels:
--- a/config/rbac/submariner-operator/certs-role.yaml
+++ b/config/rbac/submariner-operator/certs-role.yaml
@@ -1,0 +1,18 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: submariner-certs-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - delete
--- a/config/rbac/submariner-operator/role.yaml
+++ b/config/rbac/submariner-operator/role.yaml
@@ -21,6 +21,8 @@ rules:
       - secrets
     verbs:
       - get
+      - list
+      - watch
       - create
       - update
       - delete
--- a/internal/controllers/submariner/broker_controller.go
+++ b/internal/controllers/submariner/broker_controller.go
@@ -22,6 +22,7 @@ import (
 	"context"
 
 	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/certificate"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
 	"github.com/submariner-io/submariner-operator/pkg/crd"
 	"github.com/submariner-io/submariner-operator/pkg/discovery/clustersetip"
@@ -40,6 +41,7 @@ type BrokerReconciler struct {
 	ScopedClient  client.Client
 	GeneralClient client.Client
 	Config        *rest.Config
+	signer        certificate.Signer
 }
 
 //+kubebuilder:rbac:groups=submariner.io,resources=brokers,verbs=get;list;watch;create;update;patch;delete
@@ -58,6 +60,10 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
+			if r.signer != nil {
+				r.signer.Stop(request.Namespace)
+			}
+
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
@@ -65,7 +71,6 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 	}
 
 	if instance.ObjectMeta.DeletionTimestamp != nil {
-		// Graceful deletion has been requested, ignore the object
 		return reconcile.Result{}, nil
 	}
 
@@ -107,7 +112,7 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		return ctrl.Result{}, err //nolint:wrapcheck // Errors are already wrapped
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, r.setupCertificateManagement(ctx, request.Namespace)
 }
 
 //nolint:wrapcheck // No need to wrap here.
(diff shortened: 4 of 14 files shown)


bundle/manifests/submariner.clusterserviceversion.yaml
```diff
@@ -724,6 +724,33 @@ spec:
...
+          - update
+          - delete
+        - apiGroups:
+          - submariner.io
+          resources:
+          - brokers
+          verbs:
+          - get
+          - list
+          - watch
```
aswinsuryan: @tpantelis The bundle job asks me to do the changes in this way , but the duplicated broker roles needs to be in different namespaces ( broker and operator ) any idea how we could resolve it?

tpantelis: Sorry not sure I follow

tpantelis: I don't think we need to add the certs-role to the bundle as it pertains to the broker and the new cert signer will be run as part of the hub add-on agent.

config/rbac/submariner-operator/role.yaml
```diff
@@ -5,6 +5,14 @@ metadata:
...
   name: submariner-operator
 rules:
+  - apiGroups:
+      - submariner.io
+    resources:
+      - brokers
+    verbs:
+      - get
+      - list
+      - watch
```
tpantelis: I don't think we need this since the `Broker` instance isn't created in the submariner-operator namespace.

config/rbac/submariner-operator/certs-role.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: submariner-certs-role
+rules:
+  - apiGroups:
+      - submariner.io
+    resources:
+      - brokers
+    verbs:
+      - get
+      - list
+      - watch
```
tpantelis: We shouldn't need this since the [ClusterRole](https://github.com/submariner-io/submariner-operator/blob/ace5cd388efc4ff9503a94f7e19300df9c8f46f2/config/rbac/submariner-operator/cluster_role.yaml#L10) grants access to Brokers. Or we could keep this and remove the rule from the ClusterRole.

go.work
```diff
@@ -0,0 +1,3 @@
+go 1.24.7
```
tpantelis: Why was this file added?

go.work.sum
```diff
@@ -0,0 +1,86 @@
+cel.dev/expr v0.19.1/go.mod h1:MrpN08Q+lEBs+bGYdLxxHkZoUSsCp0nSKTs0nTymJgw=
```
tpantelis: Same here.

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,76 @@ var _ = Describe("Broker controller tests", func() {
...
+
+		caSecretClient := fakeDynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(submarinerNamespace)
+
+		Eventually(func(g Gomega) {
+			caSecretObj, err := caSecretClient.Get(ctx, "submariner-ca", metav1.GetOptions{})
+			g.Expect(err).NotTo(HaveOccurred())
+
+			caSecret := &corev1.Secret{}
+			err = runtime.DefaultUnstructuredConverter.FromUnstructured(caSecretObj.UnstructuredContent(), caSecret)
+			g.Expect(err).NotTo(HaveOccurred())
```
tpantelis:
```suggestion
			caSecret := resource.MustFromUnstructured(caSecretObj, &corev1.Secret{})
```

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,76 @@ var _ = Describe("Broker controller tests", func() {
 		Expect(t.GeneralClient.Get(ctx, client.ObjectKey{Name: "serviceimports.multicluster.x-k8s.io"}, crd)).To(Succeed())
 	})
 
+	It("should create the CA certificate secret", func(ctx SpecContext) {
+		freshBroker := broker.DeepCopy()
+		freshBroker.Spec.ClustersetIPCIDRRange = "10.254.0.0/16"
+		err := t.ScopedClient.Update(ctx, freshBroker)
+		Expect(err).NotTo(HaveOccurred())
```
tpantelis: This looks unnecessary.

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,76 @@ var _ = Describe("Broker controller tests", func() {
...
+
+		t.AssertReconcileSuccess(ctx)
+
+		caSecretClient := fakeDynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(submarinerNamespace)
+
+		Consistently(func(g Gomega) {
+			_, err := caSecretClient.Get(ctx, "submariner-ca", metav1.GetOptions{})
+			g.Expect(err).NotTo(HaveOccurred())
+		}).Should(Succeed())
+	})
```
tpantelis: It looks like this test was generated by Cursor or some AI tool 😄 But it doesn't really make sense, ie it doesn't actually verify the signer was only initialized once. We could probably do it by checking the actions on the fake client but I don't think this test is really necessary.

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -94,4 +193,27 @@ var _ = Describe("Broker controller tests", func() {
...
+
+		It("should handle certificate signer creation errors gracefully", func(ctx SpecContext) {
+			resource.NewDynamicClient = func(_ *rest.Config) (dynamic.Interface, error) {
+				return nil, errors.New("mock client creation error")
+			}
+
+			result, err := t.DoReconcile(ctx)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("error creating certificate signer"))
+			Expect(result.RequeueAfter).To(BeZero())
```
tpantelis: This also looks AI-generated but a better/easier way to do this is to use `FailOnAction`. Also just check that an error was returned.

```suggestion
			fake.FailOnAction(&fakeDynClient.Fake, "secrets", "create", nil, false)

			_, err := t.DoReconcile(ctx)
			Expect(err).To(HaveOccurred())
```

tpantelis (changes requested):

go.work
aswinsuryan: sorry  it was oversight in between on of the commit  , removed

internal/controllers/submariner/broker_controller_test.go
aswinsuryan: ya I used help from Cursor to generate these . :) , removed it

bundle/manifests/submariner.clusterserviceversion.yaml
aswinsuryan: Removed the broker one , but addon-agent needs permission right ? does that come from this file ?

config/rbac/submariner-operator/role.yaml
aswinsuryan: Removed

bundle/manifests/submariner.clusterserviceversion.yaml
tpantelis: I don't think so. The u/s broker controller doesn't run in the managed cluster (since in ACM we don't create a Broker instance).

vthapar: > Removed the broker one , but addon-agent needs permission right ? does that come from this file ?

Addon agent permissions are part of addon bundle code in `submariner-addon` repo.

config/rbac/kustomization.yaml
```diff
@@ -7,6 +7,8 @@ resources:
   - submariner-operator/cluster_role_binding.yaml
   - submariner-operator/ocp_cluster_role.yaml
   - submariner-operator/ocp_cluster_role_binding.yaml
+  - submariner-operator/certs-role.yaml
+  - submariner-operator/certs-role_binding.yaml
```
tpantelis: Isn't this bundle-related? If so, do we need/want this?

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -94,4 +164,23 @@ var _ = Describe("Broker controller tests", func() {
...
+	Context("Certificate management error handling", func() {
+		var originalNewDynamicClient func(*rest.Config) (dynamic.Interface, error)
+
+		BeforeEach(func() {
+			originalNewDynamicClient = resource.NewDynamicClient
+		})
+
+		AfterEach(func() {
+			resource.NewDynamicClient = originalNewDynamicClient
+		})
```
tpantelis: This isn't needed.

bundle/manifests/submariner.clusterserviceversion.yaml
aswinsuryan: The bundle validation fails without these changes, should I revert the whole file?

internal/controllers/submariner/broker_controller.go
```diff
@@ -65,7 +67,10 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	}
 
 	if instance.ObjectMeta.DeletionTimestamp != nil {
-		// Graceful deletion has been requested, ignore the object
+		if r.signer != nil {
+			r.signer.Stop(request.Namespace)
+		}
+
```
tpantelis: In order for `DeletionTimestamp` to be non-zero, there has to be a finalizer on the `Broker` object, which we don't currently add. But we don't really need that in this case. We can just trigger off of deletion, ie if the instance isn't found above. So we can stop the signer on L62 above.

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +112,49 @@ var _ = Describe("Broker controller tests", func() {
...
+					ClustersetIPCIDRRange:       "10.254.0.0/16",
+				},
+			}
+			t.InitScopedClientObjs = []client.Object{brokerWithDeletion}
+		})
+
+		It("should stop the certificate signer", func(ctx SpecContext) {
+			t.AssertReconcileSuccess(ctx)
+		})
+	})
```
tpantelis: Ideally we should verify the signer was stopped. There's no way to do that directly but we could start a SigningRequestor and verify an issued secret isn't signed:

```suggestion
	When("the Broker resource is deleted", func() {
        var signingRequestor certificate.SigningRequestor

		BeforeEach(func() {
			syncerConfig := brokersyncer.SyncerConfig{
				LocalNamespace:  "local-ns",
				LocalClusterID:  "east",
				LocalClient:     dynamicfake.NewSimpleDynamicClient(scheme.Scheme),
				BrokerNamespace: broker.Namespace,
				BrokerClient:    fakeDynClient,
				RestMapper:      test.GetRESTMapperFor(&corev1.Secret{}),
			}

			var err error

			stopCh := make(chan struct{})

			signingRequestor, err = certificate.StartSigningRequestor(syncerConfig, stopCh)
			Expect(err).NotTo(HaveOccurred())

			DeferCleanup(func() {
				close(stopCh)
			})

		})

		It("should stop the certificate signer", func(ctx SpecContext) {
			t.AssertReconcileSuccess(ctx)

			Expect(t.ScopedClient.Delete(ctx, broker)).To(Succeed())

			t.AssertReconcileSuccess(ctx)

			signedDataCh := make(chan map[string][]byte, 10)
			onSigned := func(data map[string][]byte) error {
				signedDataCh <- data
				return nil
			}

			Expect(signingRequestor.Issue(ctx, "test-secret", []string{"1.2.3.4"}, onSigned)).To(Succeed())
			Consistently(signedDataCh).ShouldNot(Receive())
		})
	})
```

tpantelis (changes requested):

bundle/manifests/submariner.clusterserviceversion.yaml
tpantelis: I think you need to revert the changes to the kustomization file as I [commented](https://github.com/submariner-io/submariner-operator/pull/3621/files#r2379981041) on.

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,55 @@ var _ = Describe("Broker controller tests", func() {
...
+
+			caSecret := resource.MustFromUnstructured(caSecretObj, &corev1.Secret{})
+
+			g.Expect(caSecret.Data).To(HaveKey("ca.crt"))
+			g.Expect(caSecret.Data).To(HaveKey("ca.key"))
+		}).Should(Succeed())
+	})
+
+	When("the Broker resource is deleted", func() {
+		var signingRequestor certificate.SigningRequestor
```
tpantelis, line 134: Newlines in between sections improve readability.



```suggestion
		var signingRequestor certificate.SigningRequestor
		
```

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,55 @@ var _ = Describe("Broker controller tests", func() {
...
+				BrokerClient:    fakeDynClient,
+				RestMapper:      test.GetRESTMapperFor(&corev1.Secret{}),
+			}
+			var err error
+			stopCh := make(chan struct{})
+			signingRequestor, err = certificate.StartSigningRequestor(syncerConfig, stopCh)
+			Expect(err).NotTo(HaveOccurred())
+			DeferCleanup(func() {
+				close(stopCh)
+			})
```
tpantelis, line 154:
```suggestion
			
			var err error
			stopCh := make(chan struct{})
			
			signingRequestor, err = certificate.StartSigningRequestor(syncerConfig, stopCh)
			Expect(err).NotTo(HaveOccurred())
			
			DeferCleanup(func() {
				close(stopCh)
			})
```

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,55 @@ var _ = Describe("Broker controller tests", func() {
...
+		})
+		It("should stop the certificate signer", func(ctx SpecContext) {
```
tpantelis, line 157:
```suggestion
		
		It("should stop the certificate signer", func(ctx SpecContext) {
```

internal/controllers/submariner/broker_controller_test.go
```diff
@@ -85,6 +114,55 @@ var _ = Describe("Broker controller tests", func() {
...
+			Expect(t.ScopedClient.Delete(ctx, brokerObject)).To(Succeed())
+			t.AssertReconcileSuccess(ctx)
+			signedDataCh := make(chan map[string][]byte, 10)
+			onSigned := func(data map[string][]byte) error {
+				signedDataCh <- data
+				return nil
+			}
+			Expect(signingRequestor.Issue(ctx, "test-secret", []string{"1.2.3.4"}, onSigned)).To(Succeed())
+			Consistently(signedDataCh).ShouldNot(Receive())
+		})
```
tpantelis, line 172:
```suggestion
			t.AssertReconcileSuccess(ctx)
			
			Expect(t.ScopedClient.Delete(ctx, brokerObject)).To(Succeed())
			
			t.AssertReconcileSuccess(ctx)
			
			signedDataCh := make(chan map[string][]byte, 10)
			onSigned := func(data map[string][]byte) error {
				signedDataCh <- data
				return nil
			}
			
			Expect(signingRequestor.Issue(ctx, "test-secret", []string{"1.2.3.4"}, onSigned)).To(Succeed())
			Consistently(signedDataCh).ShouldNot(Receive())
		})
```

internal/controllers/submariner/broker_controller.go
```diff
@@ -107,6 +112,11 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return ctrl.Result{}, err //nolint:wrapcheck // Errors are already wrapped
 	}
 
+	err = r.setupCertificateManagement(ctx, request.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
tpantelis:
```suggestion
	return ctrl.Result{}, r.setupCertificateManagement(ctx, request.Namespace)

```

tpantelis (changes requested):

tpantelis (approved):

PR #3637, opened 2025-09-08
Revert "Verify the bundle in CI"

skitt:
Reverts submariner-io/submariner-operator#3569

The bundle adjustments and verification need to be kept separate so that the verification can be backported.


tpantelis:
Why don't we just push a  PR directly to the release branch instead of reverting this? The change is simple.

skitt:
> Why don't we just push a PR directly to the release branch instead of reverting this? The change is simple.

TBH I’m mostly hoping that @dfarrell07 will take a look at https://github.com/submariner-io/submariner-operator/pull/3568 — but that could be used as a post-facto review.

tpantelis:
> > Why don't we just push a PR directly to the release branch instead of reverting this? The change is simple.
> 
> TBH I’m mostly hoping that @dfarrell07 will take a look at #3568 — but that could be used as a post-facto review.

I'll try pinging him on slack. I'm not sure he's getting/seeing PR mentions.

skitt:
Closing in favour of #3641.

PR #3641, opened 2025-09-08
Verify the bundle in CI

skitt:
This checks that nothing unexpected has changed when building the bundle.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -39,8 +39,10 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          fetch-depth: 0
       - name: Create the bundle and validate it
-        run: make bundle
+        run: make bundle LOCAL_BUILD=1
 
   check-branch-dependencies:
     name: Check branch dependencies
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -43,6 +43,20 @@ jobs:
           fetch-depth: 0
       - name: Create the bundle and validate it
         run: make bundle LOCAL_BUILD=1
+      - name: Verify generated code matches committed code (ignoring hashes)
+        run: >-
+          baseref=${{ github.base_ref }} &&
+          git add -A &&
+          git diff --staged --exit-code
+          -IcreatedAt:
+          -I"value: quay.io/submariner/"
+          -I"image: quay.io/submariner/"
+          -Icontroller-gen.kubebuilder.io/version:
+          -Ioperators.operatorframework.io.metrics.builder:
+          -I"\"version\": \"${baseref}-"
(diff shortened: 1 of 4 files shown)


skitt:
This fails with the following differences:

```patch
diff --git a/bundle/manifests/submariner.clusterserviceversion.yaml b/bundle/manifests/submariner.clusterserviceversion.yaml
index 196f8f35..afab6d65 100644
--- a/bundle/manifests/submariner.clusterserviceversion.yaml
+++ b/bundle/manifests/submariner.clusterserviceversion.yaml
@@ -35,7 +35,7 @@ metadata:
             "globalnetEnabled": false,
             "namespace": "submariner-operator",
             "repository": "quay.io/submariner",
-            "version": "v0.21.0"
+            "version": "release-0.21-ec6648d5136a"
           }
         },
         {
@@ -69,7 +69,7 @@ metadata:
             "repository": "quay.io/submariner",
             "serviceCIDR": "192.168.66.0/24",
             "serviceDiscoveryEnabled": true,
-            "version": "v0.21.0"
+            "version": "release-0.21-ec6648d5136a"
           }
         }
       ]
diff --git a/config/manifests/kustomization.yaml b/config/manifests/kustomization.yaml
index 6dd1747f..4ca738ad 100644
--- a/config/manifests/kustomization.yaml
+++ b/config/manifests/kustomization.yaml
@@ -9,6 +9,6 @@ resources:
 images:
 - name: controller
   newName: quay.io/submariner/submariner-operator
-  newTag: v0.21.0
+  newTag: release-0.21-ec6648d5136a
 - name: repo
   newName: quay.io/submariner
```

@dfarrell07 are the tags always supposed to be a specific release version?

dfarrell07:
The checked-in version was generated with `make bundle LOCAL_BUILD=1 VERSION=v0.21.0`, so it wasn't using the format from Shipyard's `Makefile.versions`.

skitt:
> The checked-in version was generated with `make bundle LOCAL_BUILD=1 VERSION=v0.21.0`, so it wasn't using the format from Shipyard's `Makefile.versions`.

Yes, that’s why the bundle validation is failing. My question is, should the bundle use specific tags all the time, in which case I’ll update the bundle validation, or will tags be handled specially at release time?

dfarrell07:
I think we'll keep iterating and figuring out the best processes, but currently for me it makes sense to go back to the `<branch>-<sha>` version after a release is finished. I think we'll need to manually generate/push a bundle with updated SHAs and the next version during releases.

skitt:
> I think we'll keep iterating and figuring out the best processes, but currently for me it makes sense to go back to the `<branch>-<sha>` version after a release is finished. I think we'll need to manually generate/push a bundle with updated SHAs and the next version during releases.

Thinking about this further, we need the validation to accept both forms anyway — otherwise releases will require overriding the validation.

dfarrell07 (approved):

tpantelis (approved):

PR #3663, opened 2025-09-15
Fix 0.20.2 bundle versions, update SHAs

dfarrell07:

### DIFF
--- a/bundle.Dockerfile.konflux
+++ b/bundle.Dockerfile.konflux
@@ -23,16 +23,16 @@ COPY bundle/tests/scorecard /tests/scorecard/
 ARG SOURCE_DATE_EPOCH
 
 LABEL cpe="cpe:/a:redhat:acm:2.14::el9"
-LABEL csv-version="0.20.1"
+LABEL csv-version="0.20.2"
 LABEL description="submariner-operator-bundle"
 LABEL distribution-scope="public"
 LABEL maintainer="['multi-cluster-networking@redhat.com']"
 LABEL name="rhacm2/submariner-operator-bundle"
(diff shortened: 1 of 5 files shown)


dfarrell07:
I think we should make the bundle validation test not-required while we're figuring out the workflow.

skitt:
> It turns out we need to create a temporary local tag for all the versions to get set correctly when generating the bundle. See commit messages for details.

`make bundle` expects a version without a leading `v`: `make bundle VERSION=0.20.2`. Then you don’t need a tag.

dfarrell07:
> > It turns out we need to create a temporary local tag for all the versions to get set correctly when generating the bundle. See commit messages for details.
> 
> `make bundle` expects a version without a leading `v`: `make bundle VERSION=0.20.2`. Then you don’t need a tag.

My goal was to match 0.21's result, but we can try this and hope OPM accepts it when rendering the FBC.

skitt:
I’m curious about the explanation in the last commit — what was the `Makefile` bug?

dfarrell07:
> I’m curious about the explanation in the last commit — what was the `Makefile` bug?

I meant this:

```
  -	sed -i -e 's/$$(VERSION)/$(VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml
  +	sed -i -e 's/$$(VERSION)/$(BUNDLE_VERSION)/g' bundle/manifests/submariner.clusterserviceversion.yaml
```

But we don't need that to get past is-semantic-version when we don't use a "v" prefix. I removed it from the commit message.

skitt (approved):

tpantelis (approved):

PR #3673, opened 2025-09-17
Handle release-to-tag changes in bundle validation

skitt:
The ignore rules need to account simultaneously for release lines and tag lines. They also need to allow unquoted versions, and handle the name change with tags.


dfarrell07 (approved):

tpantelis (approved):

dfarrell07:
I backported this to 0.22 last week, might need to go on other branches still.

skitt:
> I backported this to 0.22 last week, might need to go on other branches still.

It was needed on 0.22 because that branch made the switch from devel; it _should_ only be needed on devel itself (#3778), since all the other branches already have release branch references in their bundles, but I backported #3778 everywhere else to be on the safe side.

PR #3693, opened 2025-10-09
Remove unused mounts

aswinsuryan:
Remove unused mounts


tpantelis (approved):

skitt (approved):
Losing `/lib/modules` means we can no longer load kernel modules, but since `_stackmanager` isn’t available anyway that doesn’t change current behaviour.

PR #3744, opened 2025-11-13
Adjust to golangci-lint 2.6.1 changes

tpantelis:
See commits for details.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -239,6 +239,11 @@ linters:
       - linters:
           - goconst
         path: pkg/embeddedyamls/yamls.go
+      # Specifying "omitempty" for CRD structs (eg ObjectMeta, Status) is the norm so allow it.
+      - linter:
+          - modernize
+        path: api/v1alpha1/
+        text: "Omitempty has no effect on nested struct fields"
     paths:
       - third_party$
       - builtin$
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -71,9 +71,11 @@ linters:
     wrapcheck:
       extra-ignore-sigs:
         - .Error(
-    wsl:
-      # Separating explicit var declarations by blank lines seems excessive.
-      allow-cuddle-declarations: true
+    wsl_v5:
```
skitt, line 77: Commit message mismatch again.

internal/controllers/submariner/submariner_controller.go
```diff
@@ -123,7 +123,6 @@ func NewReconciler(config *Config) *Reconciler {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-
```
skitt, line 126: This change doesn’t match the commit message.

skitt (changes requested):

internal/controllers/submariner/submariner_controller.go
tpantelis: It actually fixes the error message indicated by the commit message. With the blank line, it thinks L127 is the start of the method comment.

skitt (approved):

PR #3755, opened 2025-11-25
Fix CVEs in release-0.22

dfarrell07:


vthapar (approved):

maayanf24 (approved):

skitt (approved):
@dfarrell07 assuming this was done with Claude, there’s something slightly amiss: the K8s commit is sufficient, and I would have expected the executable instructions to result in that commit only, not the pair submitted in this PR (which is fine apart from that).

PR #3757, opened 2025-11-25
Avoid building operator when updating bundle

dfarrell07:
Add file change filters to operator Tekton configs to prevent unnecessary builds when only bundle files are modified.


tpantelis (approved):

dfarrell07:
CC @aswinsuryan

PR #3796, opened 2025-12-09
Adjust to golangci-lint v2.7.2

tpantelis:
Revert the prior `nolint:gosec` removals due to v2.7.1.

Depends on https://github.com/submariner-io/shipyard/pull/2231

### DIFF
--- a/internal/controllers/submariner/gateway_resources.go
+++ b/internal/controllers/submariner/gateway_resources.go
@@ -185,6 +185,7 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 						// We need to be able to update /var/lib/alternatives (for iptables)
 						ReadOnlyRootFilesystem: ptr.To(false),
 					},
+					//nolint:gosec // Ignore integer overflow conversion int -> int32
 					Ports: []corev1.ContainerPort{
 						{
 							Name:          encapsPortName,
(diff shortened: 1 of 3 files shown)


internal/controllers/submariner/gateway_resources.go
```diff
@@ -349,3 +349,7 @@ func (r *Reconciler) retrieveGateways(ctx context.Context, owner metav1.Object,
 
 	return foundGateways.Items, nil
 }
+
+func toInt32(from int) int32 {
```
skitt: I think we should still keep this function removal (with the associated changes).

skitt (changes requested):

internal/controllers/submariner/gateway_resources.go
tpantelis: So you mean inline the `nolint`s. It's 3 places but would be consistent with everywhere else.

internal/controllers/submariner/gateway_resources.go
```diff
@@ -185,6 +185,7 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 						// We need to be able to update /var/lib/alternatives (for iptables)
 						ReadOnlyRootFilesystem: ptr.To(false),
 					},
+					//nolint:gosec // Tgnore integer overflow conversion int -> int32
```
skitt:
```suggestion
					//nolint:gosec // Ignore integer overflow conversion int -> int32
```

Typo :wink:

internal/controllers/submariner/loadbalancer_resources.go
```diff
@@ -135,6 +135,7 @@ func newLoadBalancerService(instance *v1alpha1.Submariner, platformTypeOCP strin
 				appLabel:           names.GatewayComponent,
 				gatewayStatusLabel: string(submv1.HAStatusActive),
 			},
+			//nolint:gosec // Tgnore integer overflow conversion int -> int32
```
skitt:
```suggestion
			//nolint:gosec // Ignore integer overflow conversion int -> int32
```

skitt (approved):

PR #3807, opened 2025-12-18
Add CVE fix slash command

dfarrell07:
Provides /cve-fix <branch> command for Claude Code to streamline invoking the CVE fix workflow.

## Summary by CodeRabbit

* **Documentation**
  * Added command documentation for CVE fix workflow guidance, including argument hints and command reference materials.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.claude/commands/cve-fix.md
+++ b/.claude/commands/cve-fix.md
@@ -1,0 +1,10 @@
+---
+description: Fix CVEs for a branch
+argument-hint: <version|devel>
+---
+
+Fix CVEs on the $ARGUMENTS branch (0.22 → release-0.22, devel → devel).
+
+If no argument provided, ask which branch to fix.
+
+@.agents/workflows/cve-fix.md


dfarrell07:
IMHO "fix cves on 0.22" is much better than "/fix-cves 0.22", but everyone seems into slash commands so here's one.

tpantelis:
> IMHO "fix cves on 0.22" is much better than "/fix-cves 0.22", but everyone seems into slash commands so here's one.

Maybe that's b/c you are intimately familiar with the proper wording of the prompts necessary to trigger the desired task since you developed all of it.  "fix cves" is pretty straightforward but perhaps some of the other 20 steps/tasks may not be.  I think the advantage of slash commands is that they can be listed so one can easily see the available ones - so in case one is unsure of how to properly word a prompt for a task, seeing a slash command that looks familiar may help.

tpantelis (approved):

Jaanki (approved):

PR #3839, opened 2026-02-03
Update CVE workflow to use Docker-based grype

dfarrell07:
Use Docker with --pull=always to guarantee the latest grype version
on every scan.

AI agents skip prerequisite steps inconsistently, often missing the DB
update. Embedding db update in each scan command ensures the database is
always current.

Local installation documented as alternative.

## Summary by CodeRabbit

* **Chores**
  * CVE fix workflow updated to use a containerized vulnerability scanner for all scans.
  * Removed local scanner installation and replaced per-package local scans with container-based runs that persist a shared vulnerability database cache.
  * Scanning flow (scan, verify, rescan, commit templates) preserved while improving consistency, performance, and cache reliability.


tpantelis (approved):

skitt:
I won’t get in the way but this feels overkill — the same result could be achieved by dropping the `grype` existence check before installing it, so that it always gets updated at the start of the run. The database was already being updated anyway.

dfarrell07:
> I won’t get in the way but this feels overkill — the same result could be achieved by dropping the `grype` existence check before installing it, so that it always gets updated at the start of the run. The database was already being updated anyway.

The idea is that the agents would sometimes skip the DB update, reading ahead in the doc. Basically they don't handle prereqs consistently.

skitt:
> > I won’t get in the way but this feels overkill — the same result could be achieved by dropping the `grype` existence check before installing it, so that it always gets updated at the start of the run. The database was already being updated anyway.
> 
> The idea is that the agents would sometimes skip the DB update, reading ahead in the doc. Basically they don't handle prereqs consistently.

That seems like useful information to have in the commit message :wink:. The key part then is the repetition of the update/scan instructions every time a scan is needed (which could be done using either a container or the installed command).

dfarrell07:
> That seems like useful information to have in the commit message 😉. 

Yep, reworded it to clarify. Thanks!

> which could be done using either a container or the installed command

Added the local install docs option back for reference.

skitt (approved):

PR #3841, opened 2026-02-05
Substitute containerImage annotation in bundle

dfarrell07:
The $(SUBMARINER_OPERATOR_IMAGE) placeholder in the CSV containerImage annotation appears raw in OperatorHub instead of showing the actual image.

This worked previously when bundles were built via CPaaS, which used a midstream render_templates script to substitute the placeholder. When Konflux replaced CPaaS for bundle builds, no equivalent substitution was added to the upstream Makefile.

Extract the operator image from related-images config and substitute it during bundle generation, matching the midstream approach.

## Summary by CodeRabbit

* **Chores**
  * Bundle generation now replaces operator image placeholders with the actual operator image so generated bundles reference a real operator image for deployment.
  * Development image tags for the operator and related components were updated to a new devel tag to ensure bundle manifests reference the latest build.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -49,6 +49,7 @@ jobs:
           git add -A &&
           git diff --staged --exit-code
           -IcreatedAt:
+          -IcontainerImage:
           -I"value: quay.io/submariner/"
           -I"image: quay.io/submariner/"
           -Icontroller-gen.kubebuilder.io/version:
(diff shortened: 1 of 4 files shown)


tpantelis (approved):

dfarrell07:
Backports: 
https://github.com/submariner-io/submariner-operator/pull/3842
https://github.com/submariner-io/submariner-operator/pull/3843
https://github.com/submariner-io/submariner-operator/pull/3844
https://github.com/submariner-io/submariner-operator/pull/3845

PR #4139, opened 2026-07-09
feat: add NodeSelector support to RouteAgent DaemonSet

### LINKED ISSUE #4138: spec.nodeSelector and spec.tolerations are not applied to managed DaemonSets
**What happened**:

The Submariner CRD defines `spec.nodeSelector` and `spec.tolerations` fields in `SubmarinerSpec` (`api/v1alpha1/submariner_types.go:235-237`), but all four managed DaemonSets never read them:

- **Gateway**: NodeSelector hardcoded to `submariner.io/gateway: true`, Tolerations hardcoded to `[{Op: Exists}]`
- **Globalnet**: same as Gateway
- **MetricsProxy**: same as Gateway
- **RouteAgent**: NodeSelector only set when `DisableIntraClusterConnectivity=true` (also hardcoded), Tolerations the same

The CR's `spec.nodeSelector` and `spec.tolerations` are silently ignored in all cases.

When a user sets `spec.nodeSelector` and/or `spec.tolerations` on the Submariner CR, those values should be merged into the pod template spec of all managed DaemonSets, on top of the existing defaults.

1. Create a Submariner CR with custom nodeSelector:
```yaml
apiVersion: submariner.io/v1alpha1
kind: Submariner
spec:
  nodeSelector:
    node.kubernetes.io/instance-type: m5.large
```
2. After reconciliation, inspect any DaemonSet:
```bash
kubectl get ds submariner-gateway -n submariner-operator -o json | jq .spec.template.spec.nodeSelector
```
3. The output only shows `{"submariner.io/gateway": "true"}` — the user's `node.kubernetes.io/instance-type` is missing.

**Environment**:
- N/A (code bug, not environment-specific)

Eivs:
RouteAgent DaemonSet now respects `cr.Spec.NodeSelector`, matching ServiceDiscovery behavior. `DisableIntraClusterConnectivity` merges into NodeSelector instead of overwriting.

Fixes #4138

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route-agent DaemonSet node selector handling by initializing when absent and merging configured selector entries without overwriting existing ones.
  * When intra-cluster connectivity is disabled, the gateway node label is added alongside any configured node selectors.
* **Tests**
  * Added additional reconciliation coverage to assert the DaemonSet template includes user-specified node selector labels and the gateway label when applicable.
* **Chores**
  * Updated an indirect tooling dependency.


skitt:
Thanks for taking care of this @Eivs! This looks fine to me, but linting is catching a few issues to address. When you do that, please amend the existing commit rather than add a new one.

skitt:
The nitpick comment in https://github.com/submariner-io/submariner-operator/pull/4139#pullrequestreview-4661122303 is also worth addressing IMO.

Eivs:
> Thanks for taking care of this @Eivs! This looks fine to me, but linting is catching a few issues to address. When you do that, please amend the existing commit rather than add a new one.

Feeling that your reply, where https://github.com/submariner-io/submariner-operator/pull/4139/changes#diff-11f03b0f97705b85807a720c4608013062002af3604f046a09a2571b0ae02554L261 is to fix the issue of Vulnerability Scanning job failure.

https://github.com/submariner-io/submariner-operator/actions/runs/29003235662/job/86069833348?pr=4139

Anchore grype scan found that oras.land/oras-go/v2 v2.6.0 in /tools/go.mod has 1 high severity CVE (GHSA-jxpm-75mh-9fp7), and --fail-on high caused the build to fail. Upgrading to v2.6.1 fixes it.

skitt (approved):
Thanks!

tpantelis (approved):

PR #4140, opened 2026-07-09
Fix CVEs in release-0.23

vthapar:
See commit msg for details

## Summary by CodeRabbit

* **Chores**
  * Updated Go module versions across the main app and tooling dependencies.
  * Bumped the Go toolchain version used for tooling.
  * Expanded security scan ignore entries to cover additional known advisories and transitive dependency cases.


skitt (changes requested):
Same issue here, the x/sys commit is a downgrade.

.grype.yaml
```diff
@@ -60,3 +60,43 @@ ignore:
   - vulnerability: GO-2026-5932
     package:
       name: golang.org/x/crypto
+
+  # Transitive dependency from github.com/submariner-io/submariner@v0.23.1 - cannot upgrade without updating submariner package
```
skitt: Why not? Each Go module controls its own dependencies — transitive dependencies are *minimum* versions, not strict dependencies.

skitt (changes requested):
There seems to be a fundamental misunderstanding wrt transitive dependencies…

skitt (approved):

Issue #1668, opened 2021-11-19
Invalid GlobalnetCidrRange during subctl deploy-broker


jinglina:
**What happened**:
errros when deploy broker using subctl (version: 0.10.1)

```
E1119 01:53:34.223223       1 controller.go:267] controller-runtime/manager/controller/broker "msg"="Reconciler error" "error"="invalid GlobalnetCidrRange: invalid CIDR address: " "name"="submariner-broker" "namespace"="submariner-operator" "reconciler group"="submariner.io" "reconciler kind"="Broker"
```

when globalnet is desabled, globalnetInfo.GlobalnetCidrRange is empty, 
when check net.parse(), it reports an error

```
if globalnetInfo != nil {
		if err = IsValidCIDR(globalnetInfo.GlobalnetCidrRange); err != nil {
			return fmt.Errorf("invalid GlobalnetCidrRange: %s", err)
		}
	}
```

**What you expected to happen**:
the error not occurs again.
**How to reproduce it (as minimally and precisely as possible)**:
subctl deploy-broker
**Anything else we need to know?**:

**Environment**:

- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`)
- Cloud provider or hardware configuration:
- Install tools:
- Others:

PR #4142, opened 2026-07-20
Add CableDriverOptions to Submariner CR

sanek9:

Adds cluster-wide `spec.cableDriverOptions` (`map[string]string`) on the Submariner CR
and injects it into the gateway as:

```text
SUBMARINER_CABLEDRIVEROPTIONS=<json>
```

The field is intentionally **driver-agnostic**: it is an opaque `key=value` bag that the
selected cable driver may consume. Drivers that do not recognize a key simply ignore it.
That keeps peer-facing `Endpoint.BackendConfig` unchanged and avoids baking
driver-specific knobs into the CR schema (one field instead of a growing set of
AmneziaWG- / Libreswan- / WireGuard-only API fields).

Today the AmneziaWG driver is the first consumer (obfuscation parameters). The same
mechanism is meant for other drivers later without another CRD change. For Libreswan,
a natural fit would be things discussed in #1668 — for example selecting IKE/ESP
proposals / ciphers so hardware crypto offload can be used — rather than growing the
CR with more Libreswan-specific fields. WireGuard or future drivers can opt in the
same way via `cable.GetDriverOptions()`.

Empty/absent map → env var is not set, so gateways keep built-in driver defaults
(compatible with older submariner images that do not read the variable yet).

## More details

- Companion: AmneziaWG cable driver (reads the env var) — already merged: https://github.com/submariner-io/submariner/pull/4097
- Companion: subctl `--cable-driver-option` — https://github.com/submariner-io/subctl/pull/1830
- CRD must be updated in both `config/crd` and `deploy/crds` (subctl installs from deploy).

## Related issues

- Related to https://github.com/submariner-io/submariner/issues/1668 (earlier discussion of configurable cable/encryption settings; this is a narrower opaque-options approach rather than a broad IPsec settings API).

## Related PRs

- Consumer (merged): https://github.com/submariner-io/submariner/pull/4097
- Companion subctl (Depends-on this PR): https://github.com/submariner-io/subctl/pull/1830

## Checklist

- [x] CRD / deepcopy updated (`config/crd` **and** `deploy/crds`)
- [x] OLM bundle CRD/CSV descriptors updated
- [x] Tests added


sanek9:
Follow-up subctl PR: >>> Creating PR #03: subctl feat/cable-driver-options -> devel
https://github.com/submariner-io/subctl/pull/1830

sanek9:
Part of the proposal overview: https://github.com/submariner-io/submariner/issues/4101

tpantelis (approved):

tpantelis:
@sanek9 "Bundle validation" job failed which means you need to run `make bundle` and check in the changes.

tpantelis (approved):

Issue #3005, opened 2024-03-24
Operator fail to update coredns configmap in rke2


adibrastegarnia:
**What happened**:
According to this document, https://submariner.io/operations/usage/, coredns configmap in kube-system should be updated automatically to forward requests sent for the clusterset.local domain to the to Lighthouse CoreDNS Server in the cluster. However, I noticed that the name of configmap is hardcoded to "coredns" based on the following code. 

https://github.com/submariner-io/submariner-operator/blob/devel/controllers/servicediscovery/servicediscovery_controller.go#L68

How does that work for rke2 enviroment where the name is prefixed with rke2? (i.e rke2-coredns) 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Diagnose information (use `subctl diagnose all`):
- Gather information (use `subctl gather`)
- Cloud provider or hardware configuration:
- Install tools:
- Others:

adibrastegarnia:
@skitt  since you are a major contributor to the project, would you please advise on this issue?

adibrastegarnia:
I managed to fix it by overriding the name of the rke2-coredns chart but would be nice to support it by default.

vthapar:
@adibrastegarnia This looks similar to an issue we fixed back for AKS https://github.com/submariner-io/submariner-operator/issues/1046

You can set the `customCoreDNSName` field in ServiceDiscovery CR to `rke2-coredns`. The hardcoded constant is used as default if none specified in `customCoreDNSName` .

If using subctl to join, you can set it using following flag:

```
 --coredns-custom-configmap string           Name of the custom CoreDNS configmap to configure forwarding to lighthouse. It should be in <namespace>/<name> format where <namespace> is optional and defaults to kube-system
```
Let us know if this worked for you.

adibrastegarnia:
@vthapar  for some reasons that flag is not listed in subctl document. I haven't tried it yet to see if it is available that is why I asked the question in the first place.

tpantelis:
> @vthapar for some reasons that flag is not listed in subctl document. I haven't tried it yet to see if it is available that is why I asked the question in the first place.

yeah it's not described on the [doc site](https://submariner.io/operations/deployment/subctl/) - unfortunately that has gotten a bit stale in some parts. `subctl join --help` is the source of truth.

PR #4144, opened 2026-07-20
Add CoreDNS custom ConfigMap key for Corefile merge

sanek9:

Extends `spec.coreDNSCustomConfig` with an optional **`key`** field:

- Default / unset → `lighthouse.server` (existing AKS-style behavior)
- `key: Corefile` → merge lighthouse forwards into a full Corefile snippet using `#lighthouse-start` / `#lighthouse-end` markers

This is needed whenever CoreDNS does not pick up AKS-style `*.server` snippet keys
and instead loads a full Corefile from the custom ConfigMap — for example:

```text
import /etc/coredns-user/Corefile
```

In that layout, writing only `lighthouse.server` appears to “succeed” but DNS never
loads the Lighthouse forwards. We hit this on **Yandex Managed Kubernetes**; the same
pattern can appear on other managed / customized CoreDNS setups (see also #3005).

When merging into `Corefile`, existing unmarked `clusterset.local` blocks are replaced so re-joins update a stale Lighthouse ClusterIP without wiping other user zones.

Embedded `deploy/crds` and the OLM bundle (CRDs + CSV `specDescriptors`) are updated so installs from CRDs or bundle both expose `key`.

## Related issues

- Extends the model from https://github.com/submariner-io/submariner-operator/issues/1046 (AKS `coredns-custom` / `lighthouse.server`)
- Related to https://github.com/submariner-io/submariner-operator/issues/3005 (non-default CoreDNS ConfigMap layouts; RKE2 discussion of custom CM naming)

## Related PRs

- Companion subctl: https://github.com/submariner-io/subctl/pull/1832 (`--coredns-custom-configmap-key`)

## Checklist

- [x] API + CRD (`config/crd` **and** `deploy/crds`)
- [x] OLM bundle CRDs + CSV `specDescriptors` include `key`
- [x] Tests

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the ConfigMap data key used for custom CoreDNS forwarding configuration.
  * CoreDNS configuration now safely merges Lighthouse forwarding rules into existing Corefiles while preserving user-managed content.
  * Updated configuration schemas and UI metadata to expose the new key setting.

* **Bug Fixes**
  * Prevented stale or duplicate forwarding rules during configuration updates and cleanup.

### DIFF
--- a/api/v1alpha1/servicediscovery_types.go
+++ b/api/v1alpha1/servicediscovery_types.go
@@ -105,6 +105,11 @@ type CoreDNSCustomConfig struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="CoreDNS Custom Config Namespace"
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Namespace string `json:"namespace,omitempty"`
+
+	// ConfigMap data key that holds the lighthouse forwarding config.
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="CoreDNS Custom Config Key"
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	Key string `json:"key,omitempty"`
 }
 
 func (sd *ServiceDiscovery) UnmarshalJSON(data []byte) error {
--- a/bundle/manifests/submariner.clusterserviceversion.yaml
+++ b/bundle/manifests/submariner.clusterserviceversion.yaml
@@ -160,6 +160,11 @@ spec:
         path: coreDNSCustomConfig.namespace
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
+      - description: ConfigMap data key that holds the lighthouse forwarding config.
+        displayName: CoreDNS Custom Config Key
+        path: coreDNSCustomConfig.key
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
       version: v1alpha1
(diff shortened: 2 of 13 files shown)


sanek9:
Companion subctl flag: >>> Creating PR #10: subctl feat/coredns-custom-key -> devel
https://github.com/submariner-io/subctl/pull/1832

sanek9:
Part of the proposal overview: https://github.com/submariner-io/submariner/issues/4101

sanek9:
@coderabbitai review

yboaron:
@sanek9 Can you also cover the CoreDNS customizations/requirements in the [enhancement proposal ](https://github.com/submariner-io/submariner/pull/4100#issuecomment-5156947174) ?

thc1006:
Hi @sanek9, I came here from #4101 while looking at how operators merge into an existing Corefile, and I hit something in `removeLighthouseMarkedSection` that seems worth settling in the proposal @yboaron asked for.

The scan uses `strings.Contains`, so any line that merely mentions the marker text switches skipping on, and only an end marker switches it back off. I lifted the function out as-is and ran it over:

```
.:53 {
    errors
    # our runbook explains the lighthouse-start marker
    health
}

example.org:53 {
    forward . 1.1.1.1
}
```

which comes back as:

```
.:53 {
    errors
```

The `example.org` block is gone and the first block has lost its closing brace, so what is left is not a valid Corefile either. A file that was hand-edited and lost its end marker would go the same way.

Matching markers as exact full lines, and leaving the ConfigMap untouched when they are not one properly ordered pair, would cover both cases.

Whether an existing unmarked `clusterset.local` block should be replaced or reported as a conflict feels like a separate question, and probably one for the proposal rather than this PR, since I can see the argument for replacing it on re-join.

Happy to send a patch for the marker part if that would be useful.

PR #4145, opened 2026-07-23
Fix gateway IPsec HostPath on immutable OS

sanek9:

The gateway pod mounts IPsec config from the node via HostPath.
Previously the host path was `/etc/ipsec.d`.

On Talos (and similar OS where `/etc` is read-only), kubelet cannot create
that directory, so the gateway pod fails to start.

This PR changes only the **host** path to `/var/lib/ipsec/ipsec.d`
(writable, next to the existing NSS mount under `/var/lib/ipsec/`).
Inside the container the path is still `/etc/ipsec.d` — gateway/Libreswan
code does not change.

## How to verify

- On Talos: gateway should start without workarounds
- On normal Linux: existing behavior preserved (DirectoryOrCreate under `/var/lib/ipsec`)
- Unit tests cover HostPath + mount for `ipsecd` / `nssdb` / `nssdir`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the gateway IPsec storage mount to use a writable host path, improving compatibility with immutable operating systems and preventing container startup failures.
* **Tests**
  * Enhanced the gateway DaemonSet test to validate the expected IPsec hostPath volumes (paths and creation mode) and verify the `ipsecd` container volume mount settings, including read-only behavior.

### DIFF
--- a/internal/controllers/submariner/gateway_resources.go
+++ b/internal/controllers/submariner/gateway_resources.go
@@ -97,7 +97,9 @@ func newGatewayPodTemplate(cr *v1alpha1.Submariner, name string, podSelectorLabe
 	}
 	volumes := []corev1.Volume{
 		{Name: "ipsecd", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
-			Path: "/etc/ipsec.d", Type: new(corev1.HostPathDirectoryOrCreate),
+			// Host path under /var (not /etc): Talos and other immutable OS have a read-only
+			// root; kubelet DirectoryOrCreate on /etc/ipsec.d fails with CreateContainerError.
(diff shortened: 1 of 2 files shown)


skitt (changes requested):
The point of using a host path mount here is to be able to share configuration with the host, which is necessary in some circumstances. This means that we can’t just switch to a different host directory…

I can see two solutions here:
1. either make the host path configurable
2. or make the type of volume configurable (host path using `/etc/ipsec.d`, or regular volume in environments where we don’t need to share with the host).

sanek9:
Hi @skitt!

Sorry for the delay... I haven't been able to give this much attention.

I've opened a new [pull request](https://github.com/submariner-io/submariner-operator/pull/4174) where I tried to address these points with a slightly different implementation. I haven't tested it on a live environment yet, only unit tests. If the approach looks good, I'll test it on the clusters.

In that case, we can close this PR.

skitt:
Closing in favour of #4174.

PR #4154, opened 2026-07-29
Add configurable metrics bind address

tpantelis:
Obtains "metrics-bind-address" setting (default: 0.0.0.0:8383) from the operator component-specific `ConfigMap` to allow users to configure it.
    
Rationale for opt-in approach vs. alternatives:
    
 1. TLS rejected: Would require OpenShift TLS Security Profile compliance (intermediate/modern/old/custom cipher suites, dynamic reconfiguration). Significant implementation and maintenance burden for a Medium-severity finding (CVSS 4.3).
    
 2. Localhost-by-default rejected: The operator creates Kubernetes Services pointing to port 8383 for Prometheus scraping (submariner-operator-metrics, submariner-gateway-metrics, etc.). Binding to 127.0.0.1 by default would break all existing metrics collection on upgrade, as Services cannot reach localhost-bound ports. Users would face immediate monitoring outages.
    
This approach maintains backward compatibility (no breaking changes) and allows security-conscious deployments to opt into localhost binding
    
When binding address set to loopback (127.0.0.1/::1):
  - Skips Service and ServiceMonitor creation (prevents scrape failures)
  - Metrics accessible only via localhost or kubectl port-forward

## Summary by CodeRabbit

* **New Features**
  * Metrics exposure is now driven by a cluster configuration bind address.
  * Metrics service settings are derived from the configured `host:port`.
  * The metrics configuration updates automatically when cluster configuration changes.

* **Bug Fixes**
  * Improved validation and parsing of the metrics bind address as `host:port`.
  * Metrics resources are not created when the bind address is unset/disabled (`""` or `0`) or targets loopback.
  * Metrics Service/ServiceMonitor creation is consolidated for consistent behavior.


cmd/main.go
```diff
@@ -80,9 +76,14 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var pprofAddr string
+	var metricsAddr string
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&pprofAddr, "pprof-bind-address", "",
 		"The address the profiling endpoint binds to. Disabled by default; set e.g. 127.0.0.1:8082 for local debugging.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0.0.0.0:8383",
+		"The address the metrics endpoint binds to. Set to \"0\" to disable metrics; "+
```
skitt: Could we use `disabled` instead of `0` to disable?


```suggestion
		"The address the metrics endpoint binds to, or \"disabled\" to disable metrics; "+
```

tpantelis: Perhaps but "0" (or "") is what controller-runtime uses to indicate disabled.

skitt: Two things bother me about this (it took me a while to figure that out) — one is that `0` is a valid IP address in this context (but that’s a controller-runtime problem), and the other is that the option values are inconsistent with the documented default for `--pprof-bind-address`. But `--metrics-bind-address 0` is easier than `--metrics-bind-address ""` so let’s leave it at that.

skitt (approved):

cmd/main.go
```diff
@@ -181,23 +182,62 @@ func main() {
...
+		log.Info("Metrics disabled - skipping Service and ServiceMonitor creation", "address", metricsAddr)
+	} else {
+		// Parse the bind address to extract host and port
+		host, portStr, err := net.SplitHostPort(metricsAddr)
+		if err != nil {
+			log.Error(err, "Invalid metrics-bind-address format", "address", metricsAddr)
+			os.Exit(1)
+		}
+
+		port, err := strconv.ParseInt(portStr, 10, 32)
```
skitt: This should use https://pkg.go.dev/net#LookupPort to support named ports.

cmd/main.go
```diff
@@ -181,23 +182,62 @@ func main() {
...
+		if err != nil {
+			log.Error(err, "Invalid port in metrics-bind-address", "port", portStr)
+			os.Exit(1)
+		}
+
+		// Validate port is in valid Kubernetes Service port range
```
skitt: This isn’t related to Kubernetes:


```suggestion
		// Validate port is in valid port range
```

cmd/main.go
```diff
@@ -181,23 +182,62 @@ func main() {
...
+		if port < 1 || port > 65535 {
+			log.Error(nil, "Port must be between 1 and 65535", "port", port)
+			os.Exit(1)
+		}
+
+		// Check if binding to loopback interface
+		bindIP := net.ParseIP(host)
```
skitt: This should use https://pkg.go.dev/net#LookupHost ideally.

skitt (commented):
Some minor comments if you update this to address Code Rabbit’s comments…

tpantelis:
> Some minor comments if you update this to address Code Rabbit’s comments…

I'll take a look at those.

Also, I think having this and the pprof one as CLI args may be problematic for users. They'd have to modify the operator deployment which I believe might be overwritten if they re-run `subctl join` or helm. We have support for component-specific `ConfigMaps` so I think we should use that Instead for these configuration options. WDYT?

In the future, if we ever have to enable TLS for metrics then we could also use the `ConfigMap` for TLS config options (ie TLS min version, cipher suites) for OCP TLS profile compliance.

skitt:
> > Some minor comments if you update this to address Code Rabbit’s comments…
> 
> I'll take a look at those.
> 
> Also, I think having this and the pprof one as CLI args may be problematic for users. They'd have to modify the operator deployment which I believe might be overwritten if they re-run `subctl join` or helm. We have support for component-specific `ConfigMaps` so I think we should use that Instead for these configuration options. WDYT?
> 
> In the future, if we ever have to enable TLS for metrics then we could also use the `ConfigMap` for TLS config options (ie TLS min version, cipher suites) for OCP TLS profile compliance.

Right, a ConfigMap would indeed be easier to manage for users!

tpantelis:
@coderabbitai approve

tpantelis:
@coderabbitai approve

cmd/main.go
```diff
@@ -258,6 +262,106 @@ func main() {
...
+			log.Info("DNS resolver timeout resolving port; assuming non-loopback and proceeding with Service creation",
+				"port", portStr)
+			// Port string must be numeric if DNS timed out (named ports need DNS)
+			// This will fail with clear error if portStr is invalid
+			port64, parseErr := strconv.ParseInt(portStr, 10, 32)
+			if parseErr != nil {
+				return fmt.Errorf("invalid port %q in metrics-bind-address: %w", portStr, parseErr)
+			}
+
+			port = int(port64)
```
skitt: Port lookups don’t involve DNS (see https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/net/lookup_unix.go;l=70 — port names are resolved by calling `getservbyname()`, which looks at `/etc/services`, or by finding them in a list stored in the `net` package).


```suggestion
		// Only treat timeout errors gracefully; fail on truly invalid ports
		if errors.Is(portCtx.Err(), context.DeadlineExceeded) {
			log.Info("timeout resolving port; assuming non-loopback and proceeding with Service creation",
				"port", portStr)
			// Port string must be numeric if the resolver timed out
			// This will fail with clear error if portStr is invalid
			port64, parseErr := strconv.ParseInt(portStr, 10, 32)
			if parseErr != nil {
				return fmt.Errorf("invalid port %q in metrics-bind-address: %w", portStr, parseErr)
			}

			port = int(port64)
```

skitt (changes requested):

tpantelis:
@skitt This is ready to go

skitt (approved):

PR #4174, opened 2026-08-12
Mount IPsec volumes only for libreswan

sanek9:

The gateway DaemonSet always mounted IPsec HostPath volumes
(`/etc/ipsec.d`, `/var/lib/ipsec/nss`, `/var/run/pluto`) for **every**
cable driver. On immutable OS roots (e.g. Talos), kubelet
`DirectoryOrCreate` under `/etc` fails and the gateway pod never starts.

Those host mounts exist to **share state with host OVN IPsec** (cert
mode). They are not needed for WireGuard / VXLAN / AmneziaWG, and for
default libreswan PSK the gateway runs its own pluto in the pod.

This PR:

1. **Cable-driver registry** (`internal/controllers/submariner/cabledriver/`):
   only **libreswan** contributes IPsec volumes; other drivers get none.
2. **Volume type by existing cert flag** (`ceIPSecUseOVNCertAuthMode`,
   set by `subctl join --use-ipsec-cert-auth-mode`):
   - PSK (default, flag unset): `EmptyDir` for `ipsecd` / `ipsecnss`
     (pre-OVN behavior)
   - OVN cert mode (flag set): `HostPath` to share with host pluto
3. **Optional HostPath overrides** via `spec.cableDriverOptions` (no new
   API fields), for immutable OS / non-standard layouts when cert mode
   is required (`--use-ipsec-cert-auth-mode`). Path options do **not**
   switch volume type - only the cert flag does:
   - `ipsecHostPath` (default `/etc/ipsec.d`)
   - `ipsecNSSHostPath` (default `/var/lib/ipsec/nss`)
   - `plutoHostPath` (default `/var/run/pluto`)

Container mount paths are unchanged (libreswan still expects
`/etc/ipsec.d`, `/var/lib/ipsec/nss`, `/var/run/pluto` in the pod).

`SUBMARINER_CABLEDRIVEROPTIONS` is still the full map from the CR
(including operator-consumed keys); libreswan ignores unknown options.

Empty `spec.cableDriver` is treated like libreswan (same as the gateway
binary default).

### Why this design

**Why EmptyDir vs HostPath follows `--use-ipsec-cert-auth-mode`?**  
That flag already means "use host OVN IPsec / do not start pluto in the
pod". HostPath is required only then. For default PSK, the gateway
starts its own pluto and only needs writable dirs **inside** the pod -
`EmptyDir` (same idea as before OVN cert support), with no host `/etc`
dependency. Reusing the existing join flag avoids a second
"volume type" knob that would have to stay in sync with auth mode.

**Why path overrides via `cableDriverOptions`, not new Spec fields?**  
Overrides are rare (immutable OS + cert mode, or non-default OVN
layout). `cableDriverOptions` + `subctl join --cable-driver-option`
already exist; no CRD churn. Keys apply only when building HostPath in
cert mode - omitting them keeps OCP defaults. They do not select
EmptyDir/HostPath.

### Extensibility

WireGuard, VXLAN, and AmneziaWG stay out of the registry (no host
volumes). A new entry belongs here only if a future cable driver must
**share persistent state with the host** (same class as OVN + libreswan
cert mode), for example:

- another host IPsec stack (e.g. strongSwan/charon coexisting with a CNI)
- hardware / SmartNIC offload that needs host device or vendor paths
- a host-managed userspace VPN with state under a node directory

Plain in-pod tunnels (kernel/userspace WireGuard variants, GENEVE/GRE-like
overlays, OpenVPN entirely inside the gateway pod) would not need
operator HostPath mounts. Add `<driver>.go` under `cabledriver/` and
register it in the map when such a case appears.

## How to verify

- Unit: `go test ./internal/controllers/submariner/...`
- libreswan PSK (default join, no cert flag): EmptyDir IPsec volumes, no HostPath
- libreswan + `subctl join --use-ipsec-cert-auth-mode`: HostPath defaults
- libreswan cert + `--cable-driver-option ipsecHostPath=...` (and optional
  `ipsecNSSHostPath` / `plutoHostPath`): HostPath uses overrides
  (still requires `--use-ipsec-cert-auth-mode`)
- `cableDriver=wireguard` (or vxlan / amneziawg): no IPsec volumes
- Talos + libreswan PSK: gateway starts without `/etc` HostPath

## Related

- Supersedes approach in #4145 (hardcoded `/var/lib/ipsec/ipsec.d` for all)
- Review feedback: configurable host path and/or volume type

## Summary by CodeRabbit

* **New Features**
  * Added cable-driver-specific gateway volume configuration.
  * Added support for certificate-based and pre-shared-key IPsec volume setups.
  * Added configurable host paths for certificate-based IPsec components.
  * WireGuard gateways now omit unnecessary IPsec volumes.

* **Bug Fixes**
  * Improved gateway pod volume handling across supported cable drivers.
  * Ensured gateway workloads use the appropriate storage type for each IPsec configuration.

### DIFF
--- a/internal/controllers/submariner/cabledriver/cabledriver.go
+++ b/internal/controllers/submariner/cabledriver/cabledriver.go
@@ -1,0 +1,51 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package cabledriver customizes gateway DaemonSet construction per cable driver.
+//
+// To add gateway volumes (or other DaemonSet bits) for a new cable driver:
+//  1. Add <driver>.go in this package implementing Driver
+//  2. Register it in the drivers map below
+//
+// Drivers with no operator-side volumes stay unregistered and contribute nothing.
+package cabledriver
+
+import (
+	"github.com/submariner-io/submariner-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Driver customizes gateway DaemonSet construction for a cable driver.
+type Driver interface {
+	Volumes(cr *v1alpha1.Submariner) ([]corev1.VolumeMount, []corev1.Volume)
+}
+
+// drivers is the static registry (same idea as cable.AddDriver / network discoverFunctions).
+var drivers = map[string]Driver{
+	"libreswan": libreswanDriver{},
+	"":          libreswanDriver{}, // omitempty → same as gateway default
+}
+
+// Volumes returns cable-driver-specific volume mounts and volumes for the gateway pod.
+func Volumes(cr *v1alpha1.Submariner) ([]corev1.VolumeMount, []corev1.Volume) {
+	if driver := drivers[cr.Spec.CableDriver]; driver != nil {
+		return driver.Volumes(cr)
+	}
+
+	return nil, nil
+}
--- a/internal/controllers/submariner/cabledriver/cabledriver.go
+++ b/internal/controllers/submariner/cabledriver/cabledriver.go
@@ -20,14 +20,16 @@ limitations under the License.
 //
 // To add gateway volumes (or other DaemonSet bits) for a new cable driver:
 //  1. Add <driver>.go in this package implementing Driver
-//  2. Register it in the drivers map below
+//  2. Register it in init() via AddDriver (see libreswan.go)
 //
 // Drivers with no operator-side volumes stay unregistered and contribute nothing.
 package cabledriver
 
 import (
+	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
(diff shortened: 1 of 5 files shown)


internal/controllers/submariner/cabledriver/cabledriver.go
```diff
@@ -0,0 +1,51 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Driver customizes gateway DaemonSet construction for a cable driver.
+type Driver interface {
+	Volumes(cr *v1alpha1.Submariner) ([]corev1.VolumeMount, []corev1.Volume)
+}
+
+// drivers is the static registry (same idea as cable.AddDriver / network discoverFunctions).
+var drivers = map[string]Driver{
```
skitt: Instead of maintaining a list of drivers here, I’d rather have the drivers register themselves, as is done in Submariner.

internal/controllers/submariner/cabledriver/libreswan.go
```diff
@@ -0,0 +1,87 @@
...
+				Path: nssHostPath, Type: &hostPathType,
+			}}},
+			{Name: "plutosocket", VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{
+				Path: plutoHostPath, Type: &hostPathType,
+			}}},
+		}
+}
+
+func hostPathOption(cr *v1alpha1.Submariner, key, defaultPath string) string {
+	if path := cr.Spec.CableDriverOptions[key]; path != "" {
```
skitt, line 89: I know I suggested allowing paths to be configured, but I need to think about the ramifications here (in particular, does it enable privilege escalation).

skitt (changes requested):
I like this a lot, thanks! Let’s flip the driver registration though.

yboaron:
Could be a good idea to also add cable-driver-change test 
- Start with cabledriver: "vxlan" → no IPsec volumes
- Update spec to cabledriver: "libreswan" → libreswan volumes
-  Reverse direction ( --> to vxlan )

PR #4181, opened 2026-08-19
RELATED_IMAGE_* should win over imageOverrides in GetImagePath

tpantelis:
`GetImagePath()` returned `imageOverrides[component]` verbatim ahead of the _RELATED_IMAGE_<component>_ env var. In production, `imageOverrides` comes from the `Submariner` CR which is namespaced and, as such, should be treated as untrusted. A namespaced CR-writer could therefore obtain root on every node by pointing `imageOverrides` at an attacker image.

Swap the precedence: treat _RELATED_IMAGE_<component>_, when set, as a trusted source (e.g. OLM) and use it ahead of `imageOverrides`[component].

## Summary by CodeRabbit

- **Bug Fixes**
  - Related-image environment variables now take precedence over configured image overrides.
  - Conflicting image overrides are logged and ignored.
  - Image overrides are applied only when no related-image environment variable is set.
  - Improved reliability when waiting for aggregated service imports during deployment validation.

- **Tests**
  - Expanded coverage for image path resolution, including environment variables, overrides, default paths, and local repositories.
  - Reorganized operator image parsing tests for clearer coverage.

### DIFF
--- a/pkg/images/image_parsing_test.go
+++ b/pkg/images/image_parsing_test.go
@@ -41,18 +41,48 @@ var imageTests = []struct {
 	{"submariner-operator", "", apis.DefaultSubmarinerOperatorVersion},
 }
 
-var _ = Describe("image parsing", func() {
-	When("Parsing image", func() {
-		It("Should parse version and repository", func() {
-			_, rep := images.ParseOperatorImage("localhost:5000/submariner-operator:local")
-			Expect(rep).To(Equal("localhost:5000"))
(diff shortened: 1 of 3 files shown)


vthapar (approved):
Only problem with this change is that image override is often used to troubleshoot or test pre-release images in user environments. This will break that feature, but security is more important. Will have to figure out a new way to test such images on OLM  deployments.

